PR #9, opened 2020-11-27
Experiments with database migrations under peewee

pylipp:
## General findings
- peewee-moves not tested yet with Python 3.8, PR submitted
- flask CLI integration (`flask db`) puts migrations into `flask/boxwise_flask/migrations/`, not configurable
- with the `peewee-db` CLI there are more options but it requires database access info: `peewee-db --directory boxwise_flask/migrations --database mysql://root:dropapp_root@mysql:3306/dropapp_dev status`
- any migration beyond creating a new table has to be written as Python script

### peewee
- the base library peewee itself has support for [schema migrations](http://docs.peewee-orm.com/en/latest/peewee/playhouse.html#schema-migrations). This does not come with any logic taking care of migration history.
- peewee-moves wraps around a [SchemaMigrator](http://docs.peewee-orm.com/en/latest/peewee/playhouse.html#SchemaMigrator). Hence the peewee migrate API is also accessible in peewee-moves migration files.

### peewee-migrate
- [peewee-migrate](https://github.com/klen/peewee_migrate) provides the same functionality as peewee-moves, uses a very similar concept, and also wraps around peewee's native `migrate` module
- latest version of peewee it is tested against: 3.3.3 (boxtribute uses 3.13.3)
- documentation of the API is basically non-existent as pointed out in several issues... we'd have to read the source, and tests if we have detailed scenarios
- last commit to the Github repo was April 2019; there's already an active [fork](https://github.com/spumer/peewee_migrate2)

### flask-pw
- [flask-pw](https://github.com/klen/flask-pw) brings an integration of peewee-migrate into flask projects, similar to what peewee-moves offers with the `flask db` CLI
- several [config options](https://github.com/klen/flask-pw#settings)
- not compatible with [peewee < 3.0](https://github.com/klen/flask-pw/issues/56)
- last commit in Nov 2018

### Preparation

Start services

    docker-compose up mysql
    docker-compose up flask

Open shell in flask container

    docker-compose exec flask sh

#### Query migrations

    flask db status

Output: No migrations found

#### Show database info

    flask db info

Output: some lines about the MySQL database, configuration, and access

#### Create empty migration file

    flask db revision empty_migration

results in `boxwise_flask/migrations/0001_empty_migration`

```
"""
empty_migration
date created: 2020-11-27 20:11:28.804903
"""


def upgrade(migrator):
    pass


def downgrade(migrator):
    pass
```

    flask db status

now displays: `[ ] 0001_empty_migration`

#### Apply migration

    flask db upgrade

    flask db status

now displays: `[x] 0001_empty_migration`

### Create a model

Limitation: Only for models that are not part of the db yet!

E.g. if `boxwise_flask/models/developer.py` existed:

```
from peewee import CharField

from boxwise_flask.db import db


class Developer(db.Model):
    name = CharField(null=True)
```

we could run

    flask create boxwise_flask.models.developer

which creates `boxwise_flask/migrations/0003_create_table_developer.py`

```
"""
create table developer
date created: 2020-11-27 21:03:19.282704
"""


def upgrade(migrator):
    with migrator.create_table('developer') as table:
        table.primary_key('id')
        table.char('name', max_length=255, null=True)


def downgrade(migrator):
    migrator.drop_table('developer')
```

which we apply via

    flask db upgrade

We can now apply queries from the mysql CLI to the developer table

    docker-compose exec mysql mysql -u root -p

    use dropapp_dev;
    insert into developer (name) values ("Alex");
    select * from developer;

### Advanced migrations

Not possible via CLI! See [this issue](https://github.com/timster/peewee-moves/issues/10). We have to code them in Python.

#### Add a column to a table

Create empty migration

    flask db revision table_box_add_color_column

and modify the created migration script to

```
"""
table_box_add_color_column
date created: 2020-11-27 21:19:34.821372
"""


def upgrade(migrator):
    migrator.add_column("stock", "color", "char", default="")


def downgrade(migrator):
    migrator.drop_column("stock", "color")
```

Note that we have to use the MySQL table names.

Modify a color to the box in the MySQL CLI:

    use dropapp_dev;
    update stock set color="red" where id=999
    select id,color from stock

When adding a new migration, we should test whether an upgrade can be immediately reverted to detect errors in the downgrade function:

    flask db downgrade


pylipp:
[Write an experimental migration Using Peewee Moves #small](https://trello.com/c/1hkFdod6/643-write-a-test-migration-using-peewee-moves-small)

pylipp:
[Set up Peewee Moves #small](https://trello.com/c/doH8RyhT/642-set-up-peewee-moves-small)

aerinsol:
Thanks for the very detailed research @pylipp. I definitely want to discuss this with you on our next sync up call.

HaGuesto:
Thanks @pylipp for the exploratory PR. Very interesting behavior.

### What I learned
- It is possible to generate a migration based on the peewee model definitions (Nice), but only for a model that did not exist before. 
- we cannot configure
- Any edits to the db tables we must write ourselves as python scripts. (I expected that and we are doing this in the legacy migration handler, too.)
- It is not supported(/tested) for python 3.8 and you created a PR for that one :)

### What I would like you to do before merging
- please test how to add and drop a foreign key
- how does the migration handler keep track of which migrations were applied and which not. Does it create an additional table in the db?
- write up a basic documentation in `flask/README.md`

aerinsol:
Don't put it in the readme yet. I want to discuss the possibility of using
some alternative options.

pylipp:
> Thanks @pylipp for the exploratory PR. Very interesting behavior.
> ### What I learned
> 
> * It is possible to generate a migration based on the peewee model definitions (Nice), but only for a model that did not exist before.
> 
> * we cannot configure
> 
peewee-moves does not come with configuration files or related environment variables or similar, that's true. We could write our own wrapper around the package (with a central [`DatabaseManager`](https://peewee-moves.readthedocs.io/en/latest/usage.html#direct-usage) but's there's barely an advantage over the `peewee-db` CLI.

> * Any edits to the db tables we must write ourselves as python scripts. (I expected that and we are doing this in the legacy migration handler, too.)
> 
> * It is not supported(/tested) for python 3.8 and you created a PR for that one :)
> 
> 
Your learnings are correct :) 
> ### What I would like you to do before merging
> 
> * please test how to add and drop a foreign key

Will do!
> 
> * how does the migration handler keep track of which migrations were applied and which not. Does it create an additional table in the db?

I answered this in the Trello card: a migration history is generated by peewee-moves using a dedicated table named 'migration_history' in the DB. Migration history can be queried using the CLI 'flask db status'
> 
> * write up a basic documentation in `flask/README.md`

Will wait a bit as @aerinsol suggested.

pylipp:
@aerinsol @HaGuesto Added a section about related packages to the first comment above.

pylipp:
### Advanced migrations (cont.)

#### Managing foreign keys

This was a bit of a journey yet helped me to gain a lot of understanding for the workings of peewee-moves and peewee.
The scenario I put up was to add a foreign key to the camps (term in boxtribute app: Base) table into the genders (ProductGender) table (regardless whether that makes sense or not...).

I first tried to make it work using the peewee-moves library. Had some trouble because the docs don't exactly match the implementation, so I jumped into the source code rather. This brought me closer to the native peewee source code (the migrate module) since peewee-moves wraps around it. So I tried to use that API directly but somehow the generated SQL was not accepted by the database. Final option, writing successful queries in the mysql CLI, and run them from Python (still possible using the peewee-moves framework!). That worked so I tried to go upwards level by level. It helped me to enable peewee's logging of the SQL it generates out of the ORM code one writes.

Long story short, below is a migration script that can run the same up/downgrade functionality in three different ways:
- using raw SQL commands
- using native peewee interface
- using peewee-moves interface
Regardless of the method, the execution happens in the context of peewee-moves, meaning that the up/downgrade is recorded in the migration history.

Some findings (beyond the ones commented into the script below):
- did not manage to run to SQL commands in one call to `execute_sql` (boils down to peewee)
- our table primary keys are _unsigned_ integer. peewee-moves/peewee assume signed by default. Having mismatching types leads to the Mysql error 'cannot add foreign key constraint' because it is attempted to refer a signed foreign key field (`camp_id`) to an unsigned primary key (`camps.id`).
- `ON DELETE/UPDATE` handling is possible but I did not experiment with it. By default, peewee does a `CASCADE` when dropping a column
- the operations of the native peewee migrate module are only effective when they are being `.run()` or passed into the convenience runner function `migrate()` (peewee-moves makes this unnecessary by wrapping native peewee)
- also, peewee-moves slightly abstracts the native peewee implementation for e.g. adding columns of various types (compare `upgrade_native_peewee/upgrade_peewee_moves` below; the former requires a Field variable to be defined but the latter internally creates it)

### Conclusion

I think peewee-moves meets our requirements. It nicely integrates into the flask-based project. It offers some level of abstraction of native peewee (plus the migration history management), yet it is flexible (native peewee migrate API can be used as well as raw SQL).

Let me know if there's more questions :) 

"""
table_genders_add_foreign_key_camps
date created: 2020-11-27 21:19:34.821372
"""
from peewee import ForeignKeyField, SQL
from playhouse import migrate
from boxwise_flask.models import base

import logging
logger = logging.getLogger('peewee')
logger.addHandler(logging.StreamHandler())
logger.setLevel(logging.DEBUG)

# Select either of "sql", "peewee_moves", or "native_peewee"
METHOD = "sql"


def upgrade(migrator):
    """Execute globally specified upgrade function."""
    globals()[f"upgrade_{METHOD}"](migrator)


def downgrade(migrator):
    """Execute globally specified downgrade function."""
    globals()[f"downgrade_{METHOD}"](migrator)


def upgrade_sql(migrator):
    cursor = migrator.execute_sql("ALTER TABLE genders ADD camp_id INT(11) UNSIGNED;")
    cursor.close()
    cursor = migrator.execute_sql("ALTER TABLE genders ADD CONSTRAINT fk_camp_id FOREIGN KEY (camp_id) REFERENCES camps(id);")
    cursor.close()


def downgrade_sql(migrator):
    cursor = migrator.execute_sql("ALTER TABLE genders DROP FOREIGN KEY fk_camp_id;")
    cursor.close()
    cursor = migrator.execute_sql("ALTER TABLE genders DROP COLUMN camp_id")
    cursor.close()


def upgrade_native_peewee(migrator):
    field = ForeignKeyField(
        model=base.Base,
        # if omitted: 'Foreign keys must specify a field'
        # use Field type, not name ("id")
        # otherwise: 'str object has no attribute column_name'
        field=base.Base.id,
        # if omitted: 'camp_id is not null but has no default'
        null=True,
        # Crucial! Otherwise: 'Cannot add foreign key constraint'
        constraints=[SQL("UNSIGNED")],
        # on_delete="CASCADE",
        # on_update="SET NULL",
    )
    # alternative for constraints:
    # https://github.com/coleifer/peewee/issues/542#issuecomment-393831940

    # not relevant
    # migrator.migrator.explicit_create_foreign_key = True

    # effectively runs
    # 'ALTER TABLE `genders` ADD COLUMN `camp_id` INTEGER UNSIGNED'
    # 'ALTER TABLE `genders` ADD CONSTRAINT `fk_genders_camp_id_refs_camps` FOREIGN KEY (`camp_id`) REFERENCES `camps` (`id`)'
    # 'CREATE INDEX `genders_camp_id` ON `genders` (`camp_id`)'
    operation = migrator.migrator.add_column("genders", "camp_id", field)
    migrate.migrate(operation)  # or operation.run()


def downgrade_native_peewee(migrator):
    # effectively runs
    # 'ALTER TABLE `genders` DROP FOREIGN KEY `fk_genders_camp_id_refs_camps`'
    # 'ALTER TABLE `genders` DROP COLUMN `camp_id` CASCADE'
    migrator.migrator.drop_column("genders", "camp_id").run()


def upgrade_peewee_moves(migrator):
    # raw SQL commands run are identical to those of upgrade_native_peewee()
    migrator.add_column(
        "genders", "camp_id", "foreign_key",
        model=base.Base,
        field=base.Base.id,
        null=True,
        constraints=[SQL("UNSIGNED")],
    )


def downgrade_peewee_moves(migrator):
    # raw SQL commands run are identical to those of downgrade_native_peewee()
    migrator.drop_column("genders", "camp_id")


def upgrade_peewee_moves_ineffective(migrator):
    # with 'safe=False' error: Table exists
    with migrator.create_table("genders", safe=True) as table:
        table.foreign_key("integer", "camp_id", "camps.id")

aerinsol:
Due to the lack of testing on Python 3.8, I'm not sure I want to move forward with merging this into master yet, despite @pylipp extensive exploration on workarounds. There are some other options that do not work with Peewee per se, but change the database itself (e.g., https://www.liquibase.org/). 

Perhaps there is a way to easily cascade direct DB changes into Peewee models? It's at least worth exploring.

pylipp:
> Due to the lack of testing on Python 3.8, I'm not sure I want to move forward with merging this into master yet, despite @pylipp extensive exploration on workarounds. There are some other options that do not work with Peewee per se, but change the database itself (e.g., https://www.liquibase.org/).

I ran the CI tests for Python 3.8 on my [fork](https://travis-ci.org/github/pylipp/peewee-moves/builds/747360039), and it passes, so I don't think we should be concerned about (it's not that the library claims to break in 3.8, rather it's simply not automatically tested on 3.8).

aerinsol (approved):
Approving with reservations:

- @pylipp I very much like your experiments, however as I may have mentioned earlier, I'm yet not convinced that adopting the peewee moves library over some other data versioning solution is the right technological solution
- I would like to explore whether the possibility of managing the schema version directly through something like [liquibase](https://www.liquibase.org/) and then cascading those changes through peewee might be a better solution
- Nevertheless, I'm happy to merge this into master for now and potentially re-examine it at a later date.

HaGuesto (approved):
Thanks a lot @pylipp ! Great work! I was able to replicate all you said and your test migration script is perfectly written.

Some comments about your findings which we probably will discuss today:

> - did not manage to run to SQL commands in one call to execute_sql (boils down to peewee)

Annoying, but not a big problem.

> - our table primary keys are unsigned integer. peewee-moves/peewee assume signed by default. Having mismatching types leads to the Mysql error 'cannot add foreign key constraint' because it is attempted to refer a signed foreign key field (camp_id) to an unsigned primary key (camps.id).

Unfortunately, our database is a still a bit of a mess in certain parts since it is still mostly based on the first prototype we created in 2016/2017. Not all primary keys are unsigned. It is a mix.

> - ON DELETE/UPDATE handling is possible but I did not experiment with it. By default, peewee does a CASCADE when dropping a column

This is quite annoying and we need to be careful here. The mysql defaults are 'RESTRICT'. a cascade update can do much harm.

> - the operations of the native peewee migrate module are only effective when they are being .run() or passed into the convenience runner function migrate() (peewee-moves makes this unnecessary by wrapping native peewee)
> - also, peewee-moves slightly abstracts the native peewee implementation for e.g. adding columns of various types (compare upgrade_native_peewee/upgrade_peewee_moves below; the former requires a Field variable to be defined but the latter internally creates it)
> - the add_column operation results in two SQL commands: add a camp_id column to the genders table, then add reference between that column and the camps table. If e.g. upgrade_peewee_moves was faulty (missing constraints argument leading to the problem above), the first command is executed but the second fails. peewee-moves does not record the upgrade as applied. This means that running flask db downgrade will run the previous downgrade operation but not the one from the current script. The database however is altered, and has to be cleaned up manually (alter table genders drop column camp_id). I thought that wrapping the add_column part into a transaction remedies this but it doesn't.

annoying stuff. It feels like raw SQL commands seem the best choice for updating existing foreign keys and adding foreign keys to existing tables. I do not see any options to do this properly with native peewee or  peewee-moves.
If we create a new column with a foreign key, the peewee-moves command looks the best way to do, but we need to properly test it against the production database before we deploy to make sure that no error occurs which we cannot properly roll back.

> - the table-based operations of peewee-moves are ineffective when the table already exists

- adding and droping existing tables works in peewee-moves
- adding and droping existing columns works in peewee-moves
- adding and droping indexes works in peewee-moves
- adding and droping NULL constraints works in peewee-moves
- renaming existing columns and tables work

- adding droping foreign keys to existing columns can only be down with raw SQL
- altering table and column definitions (like defaults) besides above can only be done with raw raw SQL 

did I understand you correctly?

pylipp:
> > * our table primary keys are unsigned integer. peewee-moves/peewee assume signed by default. Having mismatching types leads to the Mysql error 'cannot add foreign key constraint' because it is attempted to refer a signed foreign key field (camp_id) to an unsigned primary key (camps.id).
> 
> Unfortunately, our database is a still a bit of a mess in certain parts since it is still mostly based on the first prototype we created in 2016/2017. Not all primary keys are unsigned. It is a mix.

Okay, we have to take care to pass the correct options when creating a ForeignKey then. 

> > * the add_column operation results in two SQL commands: add a camp_id column to the genders table, then add reference between that column and the camps table. If e.g. upgrade_peewee_moves was faulty (missing constraints argument leading to the problem above), the first command is executed but the second fails. peewee-moves does not record the upgrade as applied. This means that running flask db downgrade will run the previous downgrade operation but not the one from the current script. The database however is altered, and has to be cleaned up manually (alter table genders drop column camp_id). I thought that wrapping the add_column part into a transaction remedies this but it doesn't.
> 
> annoying stuff. It feels like raw SQL commands seem the best choice for updating existing foreign keys and adding foreign keys to existing tables. I do not see any options to do this properly with native peewee or peewee-moves.
> If we create a new column with a foreign key, the peewee-moves command looks the best way to do, but we need to properly test it against the production database before we deploy to make sure that no error occurs which we cannot properly roll back.

I still enjoy the API of peewee-moves over raw SQL commands. Using the ORM makes database interaction more secure overall.

> > * the table-based operations of peewee-moves are ineffective when the table already exists
> 
>     * adding and droping existing tables works in peewee-moves
> 
>     * adding and droping existing columns works in peewee-moves
> 
>     * adding and droping indexes works in peewee-moves
> 
>     * adding and droping NULL constraints works in peewee-moves
> 
>     * renaming existing columns and tables work
> 
>     * adding droping foreign keys to existing columns can only be down with raw SQL
> 
>     * altering table and column definitions (like defaults) besides above can only be done with raw raw SQL
> 
> 
> did I understand you correctly?
I'm not sure what you're trying to say. Do you want me to do the remaining experiments for all the cases you listed?

HaGuesto:
Sorry for the confusing comments. 

With the following list I just wanted to summarize what I learned from your tests and check if I am correct. I'm trying to rephrase now:

I learned that
- we can add and drop tables with peewee-moves
- we can add and drop columns with peewee-moves
- we can add Foreign Keys with peewee-moves only when the column is created (we cannot add a Foreign Key with peewee-moves to an existing column)
- we can drop Foreign Keys with peewee-moves only if we drop the column as well (we cannot drop a Foreign Key alone with peewee-moves)
- we have to use raw SQL to add and drop Foreign Keys to existing columns.
- we can rename columns and tables with peewee-moves
- we cannot alter table and column definitions (like defaults or data types) with peewee-moves and have to use raw SQL.

If I get a yes to the statements from above @pylipp then we can merge it. If my statements are wrong please tell me what I misunderstood.

aerinsol:
Just adding one last comment that most of my concerns around using Peewee moves have been alleviated, but would be better if we had some extra support in maintaining Peewee. Will be talking to @pylipp privately to see what options we have here

PR #12, opened 2020-12-27
fix base query

mc18g13:


pylipp (approved):

HaGuesto:
Thanks for flagging this @mc18g13! I added a bit of refactoring on the frontend to make typescript and myself happy.

PR #13, opened 2021-01-22
Mjr68/make box on scan

matt-rose6:
Added backend functionality for querying a Box by the QR code.
- created endpoint tests (and another fixture) to check 3 cases
  - code associated with box returns the box successfully
  - code not associated with any box returns a Box error with box = None
  - code not in database returns QR error with box = None
- implemented the endpoint by defining the query and adding a resolver/model method for Box

### DIFF
--- a/flask/boxwise_flask/models/box.py
+++ b/flask/boxwise_flask/models/box.py
@@ -92,5 +92,8 @@ class Box(db.Model):
 
     @staticmethod
     def get_box(box_id):
-        box = Box.select().where(Box.box_id == box_id).get()
-        return box
+        return Box.select().where(Box.box_id == box_id).get()
+    
+    @staticmethod
+    def get_box_from_qr(qr_id):
+        return Box.select().where(Box.qr_id == qr_id).get()
--- a/flask/boxwise_flask/models/box.py
+++ b/flask/boxwise_flask/models/box.py
@@ -92,8 +92,8 @@ class Box(db.Model):
 
     @staticmethod
     def get_box(box_id):
-        return Box.select().where(Box.box_id == box_id).get()
+        return Box.get(Box.box_id == box_id)
     
     @staticmethod
     def get_box_from_qr(qr_id):
-        return Box.select().where(Box.qr_id == qr_id).get()
+        return Box.get(Box.qr_id == qr_id)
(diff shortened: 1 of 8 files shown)


flask/test/endpoint_tests/test_box.py
```diff
@@ -0,0 +1,45 @@
...
+            }}"""
+    data = {"query": graph_ql_query_string}
+    response_data = client.post("/graphql", json=data)
+    queried_box = response_data.json["data"]["box"]
+    assert "<Model: Box> instance matching query does not exist" in response_data.json["errors"][0]["message"]
+    assert queried_box == None
+
+
+def test_code_does_not_exist(client):
+    code = '"%s"' % "-1"
```
pylipp, line 35: You could put `code = '"-1"'` or even write `qr_code: "-1"` in the query string directly

flask/test/data/qr_code.py
```diff
@@ -6,11 +6,20 @@ def default_qr_code_data():
     mock_qr_code = {"id": 1, "code": "999"}
     return mock_qr_code
 
+def qr_code_without_box_data():
```
pylipp, line 9: The structure of the QR code is the same as in `default_qr_code_data`. Why did you not use that one?

pylipp: I think I understood; it's because the default box data has the ID from the default QR code data...?

flask/boxwise_flask/models/box.py
```diff
@@ -92,5 +92,8 @@ def create_box(box_creation_input):
 
     @staticmethod
     def get_box(box_id):
-        box = Box.select().where(Box.box_id == box_id).get()
-        return box
+        return Box.select().where(Box.box_id == box_id).get()
```
pylipp, line 95: It should be possible to use the shorter `return Box.get(Box.box_id == box_id)`, see [here](http://docs.peewee-orm.com/en/latest/peewee/querying.html#selecting-a-single-record); also below

pylipp (changes requested):
Done with the code review, thanks for the small and clear PR :)
Will try a functional review next. I suppose for that I'd try to scan a test QR code via the frontend?

flask/test/data/qr_code.py
matt-rose6: Exactly I wanted to test that the default QR code returns the default QR box. And then I wanted a second QR code not associated with any boxes for the other test case.

flask/boxwise_flask/models/box.py
matt-rose6: Nice catch--thanks!

pylipp (approved):

PR #16, opened 2021-02-14
Format all files with pre commit

HaGuesto:
@pylipp something to be aware of between the config of isort and pre-commit.

https://pycqa.github.io/isort/docs/configuration/black_compatibility/
https://pycqa.github.io/isort/docs/configuration/pre-commit/

I changed the pre-commit config here.


pylipp:
:+1: thanks for cleaning up!
Is this the reason to remove the seed-isort-config hook?

HaGuesto:
yes @pylipp

PR #18, opened 2021-02-25
Code Coverage Badges

billyhao12:
https://trello.com/c/RJqXFDeI/707-add-code-coverage-with-react-small-react


billyhao12:
Using the [jest-coverage-badges package](https://www.npmjs.com/package/jest-coverage-badges), I have been able to generate badges by configuring the test:coverage and test:badges scripts in the react folder's package.json. Running `yarn run test:coverage` generates a folder with a coverage-summary.json file containing the coverage data. Then, running `yarn run test:badges` generates four svg files that I can copy and paste into the README as badges.

When running `yarn run test:coverage`, I get this error message twice:
```
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

This prevents me from writing a script that does the work of both `test:coverage` and `test:badges` in one command (like what's written in the [jest-coverage-badges documentation](https://www.npmjs.com/package/jest-coverage-badges) for the `test:badges` script). Below is a screenshot of the error:
coverage-error

Any advice on resolving this error (or if it is safe to ignore) would be appreciated. Also, should we automate this badge generation, or can we just re-run these commands every time tests are written/updated?

react/package.json
```diff
@@ -35,11 +35,18 @@
     "start": "npm run build && craco start",
     "build": "react-scripts build",
     "test": "react-scripts test",
+    "test:coverage": "cross-env CI=true yarn test --env=jsdom --coverage",
```
HaGuesto, line 38: cross-env is an outdated package which you can and probably should drop

HaGuesto:
I was able to reproduce your error. I believe it is coming from the failing tests. when I removed app.test.js or home.test.ja it worked.

In general, though I would recommend you to do almost re-start and rather follow this medium article. https://medium.com/front-end-weekly/ci-cd-with-react-f4af73618d57

I also updated the trello ticket

react/package.json
billyhao12: The medium article you linked also uses cross-env. Do you know about any other packages I can use to account for different environments? `CI=true yarn test --env=jsdom --coverage` works for a Mac but not a PC.

HaGuesto:
@billyhao12 I have edited the config file now and it runs now on CircleCi. You should delete the other tests like Home and App that are not from you.

.circleci/config.yml
```diff
@@ -78,6 +79,16 @@ jobs:
...
+      - run:
+          name: Test React App
+          command: |
+            yarn test
+      - run:
+          name: Test React App Codecover
+          command: |
+            yarn run test:coverage
+      - codecov/upload:
+          file: react/coverage
```
HaGuesto, line 91: this should be a hidden folder (react/.coverage).
At the moment it does not upload data.

react/yarn.lock
```diff
@@ -4320,6 +4327,15 @@ cross-spawn@^7.0.0:
     shebang-command "^2.0.0"
     which "^2.0.1"
 
+cross-spawn@^7.0.1:
```
HaGuesto, line 4330: what's cross-spawn?

billyhao12: It's part of the cross-env package: https://www.npmjs.com/package/cross-env

.circleci/config.yml
billyhao12: I noticed that the .gitignore file in the react folder contains `/coverage`, so I thought it would be fine to name it like that. However, I also understand that it would be nice for the formatting to be consistent with what's written in the back-end. Is `react/.coverage` is preferable to `react/coverage`?

Not sure if there is a difference between a hidden folder and an ignored folder.

aerinsol:
[add code coverage with react #small #react](https://trello.com/c/RJqXFDeI/707-add-code-coverage-with-react-small-react)

.circleci/config.yml
pylipp: A folder (more general: any file) can be hidden in the context of the operating system directory view I reckon. E.g. on Linux the standard directory view `ls` does not show hidden files (such starting with a period) but `ls -a` does.
Any file can be ignored in the context of git (probably version control in general) meaning that the file is not taken into account for version control by git. It is irrelevant whether the file is hidden or not, git only cares about the filename and whether it matches a pattern in `.gitignore`.

So far about the terms in general :laughing: For the main question I'd propose that the coverage data file should be ignored since in a codebase it's mostly the code files and configuration files that are being cared about. It's consistent to follow the filename convention of the back-end. 
tldr rename the coverage data file into `.coverage` and add `.coverage` to ´.gitignore` :)

billyhao12: @pylipp Thanks for the clarification! As of now, the `codecov/upload` step looks like this:
```
- codecov/upload:
    file: coverage/clover.xml
```
I am using `coverage` instead of `.coverage` because that is what `yarn run test:coverage` generates. In order to use `.coverage`, we will have to find a way to change what the command generates.

billyhao12: Currently, Codecov indicates that it is able to read `coverage/clover.xml`, but we are getting an HTTP 400 error. It says that the "Provided token is not a UUID." You can see the error that we are getting [here](https://app.circleci.com/pipelines/github/boxwise/boxtribute/240/workflows/cb158d04-2048-4ef0-9a57-a0e5fe5f54a5/jobs/1042) under "Upload Coverage Results".

PR #19, opened 2021-03-14
changed workflow for ScanBox component

matt-rose6:
### Context
This PR is to address [this issue](https://trello.com/c/YTC4DnDx/604-create-react-tests-for-qr-code-submission-react) and partially address [this issue](https://trello.com/c/np5Z0eLP/700-make-a-box-on-qr-code-scan-follow-up-createboxtsx-react-flask-graphql-medium-xlarge). Overall, it is to update the workflow of scanning a box in the frontend.

### Changes
1. Added a new query in the utils class to fetch a box with GraphQL endpoint
2. Updated the ScanBox.tsx component to display the fetched box information on the screen when found
3. Updated the ScanBox.tsx component to redirect to the CreateBox.tsx component when code exists but is not associated with a box

### Note
- Error handling is not very robust, dependent on string matching and makes some assumptions about causes of errors
- May have overfetched with the query, not sure we need that much box info per request


aerinsol:
Nice work @matt-rose6 ! You fetched the correct amount of information for the box, don't worry 👍

react/src/views/ScanBox.tsx
```diff
@@ -1,59 +1,110 @@
...
+    if (code) {
+      const myQR = code.split("barcode=")[1];
+      setQR(myQR);
+      try {
+        getQrExistsQuery({
+          variables: {
+            qrCode: String(myQR),
+          },
+        });
+        getQrBoxExistsQuery({
```
pylipp, line 37: If the QR code does not exist this part wouldn't even be called?

flask/boxwise_flask/graph_ql/resolvers.py
```diff
@@ -69,6 +69,25 @@ def resolve_user(_, info, email):
     return get_user_from_email_with_base_ids(email)
 
 
+@query.field("qrExists")
+def resolve_qr_exists(_, info, qr_code):
+    try:
+        QRCode.get_id_from_code(qr_code)
+    except QRCode.DoesNotExist:
```
spaudanjo, line 76: Maybe it would be semantically cleaner to use or introduce a specific exist or count method on the model which returns false/true?

react/src/views/BoxInfo.tsx
```diff
@@ -0,0 +1,46 @@
...
+import React, { useEffect, useState } from "react";
+import { emptyBox } from "../utils/emptyBox";
+import { BOX_BY_QR } from "../utils/queries";
+
+function BoxInfo(qrCode) {
+  const [box, setBox] = useState(emptyBox);
+
+  const [getBoxQuery, { loading: queryLoading, error: queryError }] = useLazyQuery(BOX_BY_QR, {
+    onCompleted: (data) => {
+      var newBox = data.box;
```
spaudanjo, line 11: could we use `const` here?

react/src/views/ScanBox.tsx
matt-rose6: Good call, I just changed the workflow to better take advantage of GraphQL and fetch both pieces of info in one query

yarn-error.log
```diff
@@ -0,0 +1,37 @@
+Arguments:
```
spaudanjo, line 1: I think yarn-error.log ideally goes to .gitignore

react/src/views/ScanBox.tsx
```diff
@@ -80,24 +68,25 @@ function ScanBox() {
...
-      );
+      if (qrInfo.qrBoxExists)
+        return (
+          <Redirect
+            to={{
+              pathname: "/box-info",
+              state: { qr: qr },
+            }}
+          />
+        );
```
spaudanjo, line 80: not necessary, shorter version could look like this though: 
```
const pathname = qrInfo.qrBoxExists ? "/box-info" : "/create-box"
return <Redirect to={{to: pathname, state: { qr }, }} />;
```

react/src/views/ScanBox.tsx
```diff
@@ -1,26 +1,19 @@
 import React, { useState } from "react";
-import { useLazyQuery } from "@apollo/client";
+import { useLazyQuery, useQuery } from "@apollo/client";
```
spaudanjo, line 2: my linter notified me: unused import

react/src/views/BoxInfo.tsx
```diff
@@ -24,7 +24,9 @@ function BoxInfo(qrCode) {
   });
 
   useEffect(() => {
-    getBoxQuery(qrCode);
+    getBoxQuery({
+      variables: { qrCode: props.location.state.qr },
+    });
   }, []);
```
spaudanjo, line 30: React-Linter info: 
```React Hook useEffect has missing dependencies: 'getBoxQuery' and 'props.location.state.qr'. Either include them or remove the dependency array  react-hooks/exhaustive-deps```

File: react/src/views/BoxInfo.tsx [hunk 8]
```diff
@@ -0,0 +1,48 @@
...
+        product_id: newBox.product_id,
+        size_id: newBox.size_id,
+        items: newBox.items,
+        location_id: newBox.location_id,
+        comments: newBox.comments,
+        qr_id: newBox.qr_id,
+        box_state_id: newBox.box_state_id,
+      });
+    },
+    onError: (err) => {},
```
spaudanjo, line 23: Apollo error gets swallowed here (maybe ok for now - not sure whether the plan is to merge this PR already to master :) )

spaudanjo:
I did some manual testing and from that point it general looks good to me :) (Work In Progress though I guess?) 
I'm not sure about the exact scope of this PR. The first linked ticket in the PR description above is mainly about tests I think, this PR doesn't have tests yet though, I guess I miss some context.

spaudanjo:
As discussed with @HaGuesto we will merge this PR now and open a new one for the changes I did on top today. 
@HaGuesto could you approve and merge this one within the next hour or so?

pylipp:
@spaudanjo Can you remove the `yarn-error.log` file and git-ignore it?

pylipp (approved):
I added some tests to keep coverage in the backend code :)
Be careful if the other branch you're working on is derived from this one (you might need `rebase --onto`, [cf.](https://www.viget.com/articles/how-to-fix-your-git-branches-after-a-rebase/))

PR #36, opened 2021-06-19
Update README for peewee-moves

pylipp:
Not going too much into detail to avoid duplicating the peewee-moves documentation

### DIFF
--- a/flask/README.md
+++ b/flask/README.md
@@ -80,6 +80,11 @@ Most of our developers use [MySQL workbench](https://dev.mysql.com/doc/workbench
 
 The development database is called `dropapp_dev` and the password is `dropapp_root`.
 
+#### ORM
+
+From the Python side of the application we use an Object Relational Mapper (ORM) to interact with the database. An ORM provides a convenient abstraction interface since it leverages Python's language features and is more secure compared to using raw SQL queries.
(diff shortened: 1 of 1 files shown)


flask/README.md
```diff
@@ -200,4 +200,13 @@ A sample query you can try if it works is:
 
 ## Authentication and Authorization
 
-## Database Migrations
+## Database Schema Migrations
+
+Occasionally it might be required to update the database schema. To this end we use the [peewee-moves](https://github.com/timster/peewee-moves) tool.
```
aerinsol, line 205: Why not use the official repo? (Coleifer)

aerinsol (changes requested):
Maybe make a quick comment about rollbacks and rollforwards in versioning.

I'd also like to see aside from your comment about needing to date the schema, if you're altering the schema it's good practice to date /version it.

aerinsol:
No issue with not replicating the peewee documentation, just being clear about how the structure ties together. 

Also would be nice to tie in with a general commentary on peewee as well and how it's used here as an ORM that sits on top of the DB.

flask/README.md
pylipp: coleifer is the maintainer of peewee itself. I didnt mention it since the section is about the peewee-moves tool.

aerinsol (approved):

PR #38, opened 2021-08-22
More tables as peewee models

pylipp:
Just wanting to get some early feedback here - I added four new peewee models using pwiz (two are covered by unit tests).
Following the 'Python Models Import for Peewee document', the following transfers are still missing
* people_tags
* tags
* x_people_languages
* history (mentioned twice in the document with ChangeHistory / DBChangeHistory as Python model)

1. Should I add those here?
2. Regarding the field naming: do we go for underscored names rather?

### DIFF
--- a/flask/README.md
+++ b/flask/README.md
@@ -65,10 +65,11 @@ Since we are working with docker you do not have to install a local MySQL server
 #### General notes on Docker network
 
 In the docker-compose file we define a separate docker network called `backend` to which the back-end containers are joined. Each container can now look up the hostname `flask` or `mysql` and get back the appropriate container’s IP address.
-To access the mysql database, there are now two possibilities:
+To access the mysql database, there are now three possibilities:
 
 1. You reach the mysql db at `MYSQL_HOST=mysql` and `MYSQL_PORT=3306` or
-2. by specifying the IP-address of the gateway for `MYSQL_HOST` and `MYSQL_PORT=32000`.
+1. You execute the mysql command line client in the running container by `docker-compose exec mysql mysql -u root -p` or
+1. by specifying the IP-address of the gateway for `MYSQL_HOST` and `MYSQL_PORT=32000`.
 
 To figure out the gateway of the docker network `backend` run
 
--- a/flask/README.md
+++ b/flask/README.md
@@ -181,8 +181,8 @@ Test data is setup in the `test/data` folder and each piece of data is split up 
 
 #### Please be aware that
 
-- for new data the fixtures need to be imported in the required `conftest.py` and
-- the call to create needs to be added to `setup_tables.py` in the `test/data` directory.
+- for new data the fixtures need to be imported in `test/data/__init__.py` and
+- the call to create needs to be added to `test/data/setup_tables.py`
 
 ### Coverage analysis
 
--- a/flask/README.md
+++ b/flask/README.md
@@ -27,7 +27,7 @@ For almost all features of our development set-up you should [install Python >=3
 These recommendations are mainly meant for people developing on the back-end. If you are just a front-end person, but would like use pre-commit and the linters and formatters defined there, you can skip these steps.
 
 - [Use a version control for python like pyenv.](https://github.com/pyenv/pyenv) It provides you with much more clarity which version you are running and makes it easy to switch versions of python.
-- [Have a look at direnv >= v2.21.0](https://github.com/direnv/direnv). Virtual environments must be activated and deactivated. If you are moving through folders in the terminal it can easily happen that you either miss activating or deactivating the venv resulting in errors and time wasted for development. With direnv you can automate the activation and deactivation of venv depending on which folder you are in. There is already a `.envrc` file in the root of this repo. If you install `direnv` and allow to run it for your local repo, it will activate the python virtual environment `venv` everytime you enter the folder via a command line.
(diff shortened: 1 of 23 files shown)


aerinsol:
> I'm a bit confused here. Looking at most models it seems we follow snake_case for attributes (e.g. the [base](https://github.com/boxwise/boxtribute/blob/master/flask/boxwise_flask/models/base.py) model). I remember camelCase being discouraged for a Pythonic style.

Apologies @pylipp, you're correct. I got this mixed up with the conventions for the GraphQL schema. 

[Peewee documentation](https://docs.peewee-orm.com/en/latest/peewee/quickstart.html) also encourages snake_case for attributes and PascalCase for models, so let's go with that. 

> 
> > I would ignore these two fields as we're refactoring the design of tags so that it includes inventory as well as people.
> 
> I assume you mean models instead of fields, however, all clear! I'll add x_people_languages though.
> 

Yes you're right I meant models! Thanks.

pylipp:
[Import Beneficiary/Shop and Log Tables into Peewee models](https://trello.com/c/Ew2XkaDt/695-import-beneficiary-shop-and-log-tables-into-peewee-models)

pylipp:
@aerinsol In total it's now six new model definitions. Please review them and let me know about any inconsistencies.

flask/boxwise_flask/models/log.py
```diff
@@ -0,0 +1,12 @@
+from boxwise_flask.db import db
+from peewee import CharField, DateTimeField, TextField
+
+
+class Log(db.Model):
+    content = TextField()
+    ip = CharField(null=True)
+    date = DateTimeField(column_name="logdate", null=True)
+    user = CharField(null=True)
```
pylipp, line 9: @aerinsol @HaGuesto what user is meant here since it's not the User table that is referenced?

aerinsol (approved):
Created models: 

1. Beneficiary, 
2. Transaction, 
3. DBChangeHistory, 
4. Log, 
5. XBeneficiaryLanguage, 
6. Settings.

Approved. Not sure Settings is needed but will let @HaGuesto review this. 

We will probably need your help wiring up the resolvers for GraphQL as well soon @pylipp. Also note that this PR is failing the flask CircleCI tests.

pylipp:
@aerinsol @vahidbazzaz Do you want to have a final look?

flask/README.md
```diff
@@ -64,19 +64,20 @@ Since we are working with docker you do not have to install a local MySQL server
...
 #### General notes on Docker network
 
-In the docker-compose file we define a separate docker network called `backend` to which the back-end containers are joined. Each container can now look up the hostname `flask` or `mysql` and get back the appropriate container’s IP address.
-To access the mysql database, there are now two possibilities:
+In the docker-compose file we define a separate docker network called `backend` to which the back-end containers are joined. Each container can now look up the host name `flask` or `mysql` and get back the appropriate container’s IP address.
+To access the mysql database, there are now three possibilities:
 
 1. You reach the mysql db at `MYSQL_HOST=mysql` and `MYSQL_PORT=3306` or
-2. by specifying the IP-address of the gateway for `MYSQL_HOST` and `MYSQL_PORT=32000`.
+1. You execute the mysql command line client in the running container by `docker-compose exec mysql mysql -u root -p` or
```
aerinsol, line 71: Use the correct numbers please @pylipp  (1,2,3)

flask/boxwise_flask/models/x_beneficiary_language.py
```diff
@@ -0,0 +1,27 @@
...
+from boxwise_flask.models.language import Language
+from peewee import ForeignKeyField
+
+
+class XBeneficiaryLanguage(db.Model):
+    language = ForeignKeyField(
+        column_name="language_id",
+        field="id",
+        model=Language,
+        on_update="CASCADE",
```
aerinsol, line 12: Do we want to cascade cross reference tables? I'm a bit rusty but I think there may be some risk there

aerinsol (commented):
Happy to approve, but please double check the cascade permission for cross reference tables @pylipp

flask/README.md
pylipp: It's a markdown feature that this an enumeration can be indicated by `1., 1., 1.` and will be rendered as `1., 2., 3.`. I find it helpful because I don't have to change surrounding numbers when I insert an enumeration point inbetween. Should I still update the numbers?

flask/boxwise_flask/models/x_beneficiary_language.py
pylipp: In `init.sql` we have
```
  KEY `language_id` (`language_id`),
  CONSTRAINT `x_people_languages_ibfk_2` FOREIGN KEY (`language_id`) REFERENCES `languages` (`id`) ON DELETE CASCADE ON UPDATE CASCADE,
```
The model is derived from those instructions. If we want to have it defined differently it should happen in the 'upstream' dropapp database, and then be ported here, I think.

pylipp: Same for the `cms_usergroups_functions` table and the `UsergroupModuleAccess` model, resp.

flask/README.md
aerinsol: Ah, I didn't know that trick. Good for me to know too ;). Let's keep as is

flask/boxwise_flask/models/x_beneficiary_language.py
aerinsol: Noted, logged here: https://trello.com/c/OQPbLsvo

pylipp:
[EPIC: Mirror Database Modeling in Peewee](https://trello.com/c/zmAYuUrZ/641-epic-mirror-database-modeling-in-peewee)

flask/boxwise_flask/models/log.py
vahidbazzaz: @pylipp I think this field used as reference note of the user name in the log, since the user may be deleted or changed later on the system, but @HaGuesto can give more insight on that.

vahidbazzaz (approved):
@pylipp, thanks for a job well done. For language and x_beneficiary_language, we need more tests, but you already mentioned it in your PR. I will check this as well and get back to you.

PR #40, opened 2021-09-07
Configure dependabot to update pip and npm packages

pylipp:
Basically followed the instructions [here](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/about-dependabot-version-updates) and in the sibling pages.
Are you fine with the values for the check interval?
In the beginning we'll probably see some PRs to get all dependencies up-to-date.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/flask/"
+    schedule:
+      interval: "weekly"
+    assignees:
+      - "pylipp"
+  - package-ecosystem: "npm"
+    directory: "/react/"
+    schedule:
+      interval: "weekly"


aerinsol:
@pylipp does this include security updates or no? I think security updates should be daily. Version updates I'm happy with weekly.

pylipp:
> @pylipp does this include security updates or no? I think security updates should be daily. Version updates I'm happy with weekly.

Security updates are included by default since our repo meets the criteria listed [here](https://docs.github.com/en/code-security/supply-chain-security/managing-vulnerabilities-in-your-projects-dependencies/configuring-dependabot-security-updates#supported-repositories). This should also be visible from Settings -> Security (but I don't have access to that menu).
Security updates are pushed out in an _alert_ fashion rather than after checks scheduled for certain times.

aerinsol (approved):

PR #59, opened 2021-09-13
Proof of concept: box resolvers

pylipp:
This PR shows a PoC for box resolvers. I implemented the four resolvers as requested, and learnt a lot about the resolver 'magic' that ariadne provides. It ties together nicely with peewee since the rows returned by DB queries can be directly accessed for their attributes (we should watch out for a 1-to-1 mapping between the GraphQL field names and the Peewee field names. Conversion of camelCase and snake_case happens semi-automatically. For anything else we can have custom resolvers, e.g. below for the ID field).

@aerinsol 
1. you asked me about the query design, I'm not quite decisive yet about the approach. As a BE engineer I'd probably opt for the second approach because it does not require additional handling (number of) of the arguments.
1.1 parametrized query (`GetBoxDetails`): requires caution on client-side to pass only either argument, and/or handling of the exceptional case when passing zero/two arguments. The handling becomes more difficult the more (mutually exclusive) parameters we have. Yet it might be convenient for the client to only have a single query which accepts either argument
1.2 specialized query (`GetBoxesByLocation`, `GetBoxesByGender`): we can use `!` to require the argument being passed in the query. Very simple resolver implementation without additional conditional branches (basically a DB query and overhead from the Python decorators)
2. for the select-boxes-given-a-gender query, is the double join the correct way to go (it's literally the first join query I've written)
3. I applied more camelCase style to the arguments in the query definition, hope that's fine!

If you or @vahidbazzaz have any other questions or suggestions, let me know =)

Still TBD: limiting the query depth using [cost validators](https://ariadnegraphql.org/docs/query-validators#query-cost-validator)

### DIFF
--- a/flask/boxwise_flask/graph_ql/type_defs.py
+++ b/flask/boxwise_flask/graph_ql/type_defs.py
@@ -150,7 +150,7 @@ type_defs = gql(
         size_id: Int #this is a foreign key
         items: Int
         location_id: Int! #this is a foreign key
-        comments: String!
+        comment: String!
         #this will get looked up to turn into qr_id, which is a foreign key
         qr_barcode: String!
         created: Datetime #this is an output, but not an input
--- a/flask/boxwise_flask/graph_ql/type_defs.py
+++ b/flask/boxwise_flask/graph_ql/type_defs.py
@@ -5,7 +5,7 @@ type_defs = gql(
     """
     type Box{
         ID: ID!
-        boxLabelNumber: Int!
+        boxLabelIdentifier: String!
         location: Location!
         items: Int!
         product: Product!
--- a/flask/boxwise_flask/graph_ql/type_defs.py
+++ b/flask/boxwise_flask/graph_ql/type_defs.py
@@ -3,11 +3,132 @@ from ariadne import gql
 
 type_defs = gql(
     """
-    type Base {
-        id: Int
+    type Box{
+        ID: ID!
+        boxLabelNumber: Int!
+        location: Location!
+        items: Int!
+        product: Product!
+        gender: ProductGender
+        # A size from a size range, consider making this enum
+        size: String!
+        state: BoxState!
+        qrCode: QRCode!
+        createdBy: String!
+        createdOn: Datetime!
+        # The user who last changed something about the box
+        lastModifiedBy: String!
+        lastModifiedOn: Datetime!
+        comment: String
+    }
+
+    type QRCode{
+        ID: ID!
+        code: String!
+        box: Box
+        createdOn: Datetime
+        # The user who generated QR code
+        createdBy: String!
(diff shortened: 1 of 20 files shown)


pylipp:
[Proof of Concept: Peewee resolvers for "Boxes" model in GraphQL](https://trello.com/c/IwOvwIm0/744-proof-of-concept-peewee-resolvers-for-boxes-model-in-graphql)

aerinsol:
@spaudanjo you might want to check this out.

flask/boxwise_flask/graph_ql/type_defs.py
```diff
@@ -3,11 +3,132 @@
 
 type_defs = gql(
     """
-    type Base {
-        id: Int
+    type Box{
+        ID: ID!
+        boxLabelNumber: Int!
```
pylipp, line 8: @aerinsol what field of the Peewee model do `ID` and `boxLabelNumber` refer to? Currently the `stock` table has the columns `id` and `box_id`.

pylipp:
FYI  @aerinsol the coverage-check is failing because I corrected the coverage target. Previously also test code was included in the analysis which resulted in an erroneously high coverage rate.

flask/boxwise_flask/graph_ql/type_defs.py
vahidbazzaz: @pylipp I think the boxLabelNumber refer to box_id and ID is refer to id is stock table. Hope this helps!

pylipp: Thanks! Then here `boxLabelNumber` needs to be defined as String.

vahidbazzaz: @pylipp yes, I think so; this should be change to String to be sync with current data model.

aerinsol: @vahidbazzaz @pylipp if that's so then that's super confusing and definitely needs to be changed :D

I've updated the schema so `boxLabelNumber` is now `boxLabelIdentifier ` and it's a string. In real terms, this is the human-readable "ID" that gets written on the box. @pylipp can we change the Peewee model to something similar like `box_label_identifier` to match?

vahidbazzaz: @aerinsol I agree with you; also keeping the GraphQL schema in sync between UI and backend is of utmost importance as we use schema-first GraphQL engine.

pylipp: Got it! As a consequence, the related query should then be `getBoxDetails(boxLabelIdentifier: String)`, right?

aerinsol (approved):

PR #67, opened 2021-09-21
GraphQL schema update

pylipp:
This PR
- adds a couple of queries and mutations for completeness of the PoC incl. resolver implementation
- moves GraphQL type definitions into a separate file such that it can be easier synced with GraphQL editor export


pylipp:
[Proof of Concept: Peewee resolvers for "Boxes" model in GraphQL](https://trello.com/c/IwOvwIm0/744-proof-of-concept-peewee-resolvers-for-boxes-model-in-graphql)

flask/boxwise_flask/graph_ql/query_defs.py
```diff
@@ -5,16 +5,21 @@
...
     type Query {
         hello: String!
-        allBases: [Base]
-        orgBases(orgId: Int): [Base]
-        base(id: Int!): Base
+        bases: [Base!]!
+        base(id: ID!): Base
+        organisation(id: ID!): Organisation
+        organisations: [Organisation!]!
         allUsers: [User]
```
aerinsol, line 12: why the inconsistency between `allUsers `vs. `user `and `bases `instead of `allBases`?

flask/test/endpoint_tests/test_box.py
```diff
@@ -1,73 +1,32 @@
 import pytest
 
 
-@pytest.mark.usefixtures("default_qr_code")
 @pytest.mark.usefixtures("default_box")
-def test_get_box_from_code(client, default_box, default_qr_code):
-    code = default_qr_code["code"]
-    graph_ql_query_string = f"""query Box {{
-                getBoxDetails(qrCode: "{code}") {{
+def test_get_box_from_id(client, default_box):
```
aerinsol, line 5: test_get_box_from_id seems like a confusing name. Usually id refers to the internal id in the DB, but in this case it looks like you mean the boxLabelIdentifier, which essentially means the number written on the outside of the box. It would better if the test clarified that, or had a comment explaining what it did, or both.

flask/test/model_tests/test_box.py
```diff
@@ -30,5 +30,5 @@ def test_box_model(
     assert queried_box_dict["items"] == default_box["items"]
     assert queried_box_dict["location"]["id"] == default_box["location"]
     assert queried_box_dict["product"]["id"] == default_product["id"]
-    assert queried_box_dict["product"]["product_gender"] == default_product_gender
-    assert queried_box_dict["product"]["product_category"] == default_product_category
+    assert queried_box_dict["product"]["gender"] == default_product_gender
```
aerinsol, line 33: here we do want to keep the differentiation between `product_gender `and `gender`. Gender will be used to refer to beneficiaries down the line, whereas product_gender just refers to products.

aerinsol (changes requested):
Reviewed, some clarifications needed.

flask/boxwise_flask/graph_ql/query_defs.py
pylipp: The user-related queries are still to be updated. I suggest we name them `user` and `users` then

flask/test/model_tests/test_box.py
pylipp: This means a product can have both a gender and a product-gender?

aerinsol:
I just want to hear you explain the naming convention a little bit

aerinsol:
No - products have product genders. Humans have genders. That's a different
list. For example, you can distribute unisex adult products to men and
women, but men and women are not unisex adults 😅

flask/boxwise_flask/graph_ql/query_defs.py
pylipp: As a naming convention for GraphQL queries I try to follow the style of most examples, and the official docs: singular noun for single resource, plural noun for list of resources, no additional verbs (like get, fetch, ...), no additional adverbs (as I said, the 'all' in `allUsers` will soon be dropped ;-)

flask/test/model_tests/test_box.py
pylipp: Ok I see. For the model fields I try to follow the convention of having little duplication in the names. Given a `Product` instance, the `gender` field is used as `product.gender` hence IMHO from the context it can be distinguished from e.g. `beneficiary.gender`, and the duplication `product.product_gender` is not necessary.

aerinsol:
I think that's fine on the back-end, but there will likely be a time where
we're matching beneficiaries and products on the front-end so that's likely
to get confusing

flask/boxwise_flask/graph_ql/types.gql
```diff
@@ -0,0 +1,154 @@
...
+  name: String
+  email: String!
+  validFirstDay: Date
+  validLastDay: Date
+  bases: [Base]
+  lastLogin: Datetime
+  lastAction: Datetime
+}
+
+input CreateBoxInput {
```
pylipp, line 124: @aerinsol I need your input about what fields could possibly be included in a `UpdateBoxInput`. All of `CreateBoxInput` but qrCode and createdBy?

aerinsol: Not Updated
- id
- boxLabelIdentifier
- qrcode
- createdBy

Always Updated
- lastModifiedBy
- lastModifiedOn

Not yet dealt with in this implementation
- BoxState

flask/boxwise_flask/models/box.py
```diff
@@ -110,38 +110,20 @@ def __unicode__(self):
         return self.box_label_identifier
 
     @staticmethod
-    def create_box(box_creation_input):
+    def get_box(box_id):
+        return Box.get(Box.box_label_identifier == box_id)
```
aerinsol, line 114: looking at the old deleted stuff, box_label_idenfier is not the primary key. I believe @spaudanjo in his implementation put it as box_short_uuid which is a clever solution. I'll leave it to the two of you to determine which makes more sense.

aerinsol: I will also say @spaudanjo comments on the business logic are quite good, you might want to keep them @pylipp

pylipp: I'm aware that `box_label_identifier` is not the PK. However it is used here because of the way the query `box(boxId: String)` is defined. If it makes more sense to have `box(id: ID)` I'm happy to change the query (but I think on the FE the box label identifier is requested as an input to look a box), or to simply rename the field to e.g. `box_uuid`

The comments about the business logic were part of the code since the days it was still in a separate `boxwise-flask` repo :D I don't think there have to be comments in the BE/FE referencing any possibilities or options what to do on the other end, that should be part of the backlog rather than staying in the code base (and potentially becoming stale). However I did check the comments again, and noticed an important aspect about the length of the box_label_id field, thank you!

flask/boxwise_flask/models/box.py
```diff
@@ -110,38 +110,20 @@ def __unicode__(self):
...
-            created=today,
-            # this is consistently NULL in the table, do we want to change that?
-            created_by=box_creation_input.get("created_by", None),
-            box_state_id=1,  # always 1 for create?
-        )
-        return new_box
+def create_box(box_creation_input):
+    today = datetime.now()
+    qr_code = box_creation_input.pop("qr_code", None)
+    qr_id = QRCode.get_id_from_code(qr_code)
```
pylipp, line 120: @aerinsol is it true that a QR code is always generated _before_ a box is created? Can a box be created _without_ a QR code at all?

vahidbazzaz: @pylipp As far as I know, the QR code generated separately and printed as label will later be scanned and assigned to the box, so I think this should be considered as required input for the create_box method. Hope this helps.

flask/boxwise_flask/models/box.py
vahidbazzaz: @aerinsol It looks like box_id is indexed with a unique constraint, so it should be fine but I concur with @pylipp comments about fixed length and @spaudanjo nice suggestion for using uuid, but it will likely affect DropApp. As a result, if we decide to change to uuid, we will have to verify DropApp will work well.

flask/boxwise_flask/models/box.py
aerinsol: Yes that's correct, currently it's possible to generate a box without having a qrcode attached to it. We are thinking of discontinuing this feature, but since this behavior is currently allowed we should not force this validation on the back-end.

flask/boxwise_flask/models/box.py
```diff
@@ -110,38 +114,38 @@ def __unicode__(self):
...
 
-    @staticmethod
-    def get_box(box_id):
-        return Box.get(Box.box_label_identifier == box_id)
+    new_box = Box.create(
+        box_label_identifier=str(uuid.uuid4())[: Box.box_label_identifier.max_length],
+        qr_id=qr_id,
+        created_on=now,
```
pylipp, line 132: @aerinsol Do we consider creation a modification? For consistency we should then set the `last_modified_*` at creation, too, even if the values are identical to the `created_*` fields.

flask/boxwise_flask/graph_ql/resolvers.py
```diff
@@ -38,41 +44,33 @@ def serialize_date(value):
...
-# TODO get currrent user based on email in token
 @query.field("user")
 def resolve_user(_, info, email):
-    return get_user_from_email_with_base_ids(email)
+    data = get_user_from_email_with_base_ids(email)
+    data["organisation"] = (
+        Organisation.select()
+        .join(Usergroup)
+        .where(Usergroup.id == data["usergroup"]["id"])
+        .get()
```
pylipp, line 71: @aerinsol @vahidbazzaz I just noticed that this lookup does not make much sense anymore. We decided to have the organisation ID sent with the JWT, so here the ID would be used to select the corresponding entry from the organisation table. I have to look into how to test that though.

flask/boxwise_flask/models/box.py
aerinsol: Yes, set it identical to the create field. @pylipp

flask/boxwise_flask/models/box.py
aerinsol: @pylipp @vahidbazzaz `box_label_identifier` is required as an input to retrieve boxes, that's true. However, it's likely we'll have to request the internal PK for the `box` type as well. Therefore I'd prefer it if there was some indication that the id being used is for the label vs. the internal PK.

aerinsol: > I'm aware that `box_label_identifier` is not the PK. However it is used here because of the way the query `box(boxId: String)` is defined. If it makes more sense to have `box(id: ID)` I'm happy to change the query (but I think on the FE the box label identifier is requested as an input to look a box), or to simply rename the field to e.g. `box_uuid`
> 

Sorry, I think you are right, it's fine as is. Realized this is in the models section of the codebase where you are connecting the model definitions with the true MySQL back-end, I misread it and thought it was having to do with the GraphQL definitions.

Please keep as is and carry on! @pylipp

flask/boxwise_flask/graph_ql/types.gql
aerinsol: @pylipp again, we may want to change some of the behavior down the line for the seamless transfer functionality, but that's out of scope of this PR :)

vahidbazzaz (approved):
Thanks @pylipp for addressing all the comments.

flask/boxwise_flask/graph_ql/resolvers.py
pylipp: Agreed to work on this in a follow-up PR

pylipp:
@aerinsol Merging is possible once you will have approved :) (if you merge right away, please do by merge-commit strategy)

aerinsol (approved):

PR #71, opened 2021-09-27
Update JWT handling in flask app

pylipp:
Highlights of this PR:
- cleanup of ineffective patching for unit tests
- store JWT custom claim information in Flask `g` object (aka application context)
- require authorization for some GraphQL queries (base, product, location, organisation)

@aerinsol @vahidbazzaz Can you check whether the way the resolvers are implemented is the right direction for authorizing access (using `base_ids` and `organisation_id` from JWT)? I can then update the 'plural' queries, too. For some queries I was not certain how to authorize, e.g.
- box: check box -> location -> base ?
- qrCode: check qrCode <- box -> location -> base ?
- user(s): simply check email ? (will probably become more complex with `permissions`)
- productCategory: ?

Also I noticed that whenever authorization fails, a 200 response is returned (despite our custom AuthError and its handler) instead of a 401. So I'll try to look into that as well.

And I have to find out why the CircleCI runs fail (seems like the `AUTH0_*` variables are not suitable such that a JWT for testing can be retrieved).

FYI I tried to incorporate the `flask-jwt-extended` package but
- installation is a bit sophisticated (its dependency `cryptography` requires a rust and a C compiler to be built from source; it's fine in CircleCI but locally I didn't manage to compile everything)
- half of the package is designed for creating JWT which we don't need since it happens on Auth0 side
- the `current_user` convenience object can be replaced by `g.user` (see above and code)
Bottom line: package is not integrated


pylipp:
[Integrate flask-jwt-extended package](https://trello.com/c/kr8cAgVo/754-integrate-flask-jwt-extended-package)

vahidbazzaz:
@pylipp Regarding the authorization, I will go through the queries and get back to you, as for flask-jwt, could you please check if there are wheel files for that, as that would save you the compilation step.

pylipp:
> @pylipp Regarding the authorization, I will go through the queries and get back to you, 

Thanks!

> as for flask-jwt, could you please check if there are wheel files for that, as that would save you the compilation step.

I looked for wheels for the critical dependency `cryptography` [here](https://pypi.org/project/cryptography/#files) but unfortunately none are available for my system (CPython 3.7 on Linux 64-bit). It's not a blocker though to not being able to use `flask-jwt-extended` so that's fine!

vahidbazzaz:
> > @pylipp Regarding the authorization, I will go through the queries and get back to you,
> 
> Thanks!
> 
> > as for flask-jwt, could you please check if there are wheel files for that, as that would save you the compilation step.
> 
> I looked for wheels for the critical dependency `cryptography` [here](https://pypi.org/project/cryptography/#files) but unfortunately none are available for my system (CPython 3.7 on Linux 64-bit). It's not a blocker though to not being able to use `flask-jwt-extended` so that's fine!

@pylipp Thanks for letting me know about this. I'm sorry to hear that there is no appropriate wheel for your environment, and let me know if I can be of any help in compiling.

aerinsol:
@pylipp not sure I understand your question in terms of check the direction of the authorization, sorry. Can you elaborate?

pylipp:
> @pylipp not sure I understand your question in terms of check the direction of the authorization, sorry. Can you elaborate?

It's the first time I implement authorization in an application. Is the approach of intercepting an incoming request at resolver level appropriate? If so I'd add similar checks for the remaining resolvers.

vahidbazzaz:
> > @pylipp not sure I understand your question in terms of check the direction of the authorization, sorry. Can you elaborate?
> 
> It's the first time I implement authorization in an application. Is the approach of intercepting an incoming request at resolver level appropriate? If so I'd add similar checks for the remaining resolvers.

@pylipp I think you are doing the right things, but we may consider using some authorization helpers such as has_permission, is_authenticated, etc., as well as base_ids to filter out the data for the organizations the user belong to. 

We also need to filter locations based on user base_ids, and boxes for authorized locations should also be created. As for users, I think we should consider to use user_id and not email, since email can be modified, but user_id (also part of JWT token [sub]) are unique to each user. We can discuss these together and figure out the best course of action.

aerinsol:
@vahidbazzaz @pylipp a little research goes a long way. Looks like it's strongly not recommended to implement authorization this way for GraphQL resolvers and to handle it in one location in the business logic layer: https://graphql.org/learn/authorization/

This makes sense to me. An Apollo architect (Ariadne is based off of Apollo) talks about the use cases for when you might want to do resolver - level authorization: https://www.apollographql.com/blog/backend/auth/authorization-in-graphql/

Essentially when we start wanting to expose the API to the public. However it's possible to defer this decision and implementation decision down the line, so we should do that.

As for the correct pattern to implement on the business logic layer, I ask you both to take a look at the [common architectural patterns for Flask](https://flask.palletsprojects.com/en/2.0.x/patterns/packages/) and see what makes sense.

pylipp:
> @vahidbazzaz @pylipp a little research goes a long way. Looks like it's strongly not recommended to implement authorization this way for GraphQL resolvers and to handle it in one location in the business logic layer: https://graphql.org/learn/authorization/
> 
> This makes sense to me. An Apollo architect (Ariadne is based off of Apollo) talks about the use cases for when you might want to do resolver - level authorization: https://www.apollographql.com/blog/backend/auth/authorization-in-graphql/

Thanks for researching! I also found [this](https://graphql.org/learn/thinking-in-graphs/#business-logic-layer) in the official GraphQL docs, stating that authorization should happen in a dedicated layer right before accessing data. **I conclude that in the long run we should handle authorization outside the resolvers.** To some extend we already have `auth_helper.authorization_test` but we don't have it combined yet with accessing data.

@vahidbazzaz you suggested an `is_authenticated` function, I think it comes very close to the `require_auth` decorator that we already have. It protects the `/graphql` endpoint and beyond that we can be sure that the user is authenticated.

Good point about user email and ID. However I'd discuss it in a separate task for the sake of the scope :)

vahidbazzaz:
@pylipp With regards to is_authenticated, I thought that this was particularly fitting for situations where we might have queries that do not require authentication, but since we are discussing the authorization, we can figure out what is most appropriate and finalize it soon.

vahidbazzaz (approved):
Thanks @pylipp, just wanted to ask if it is possible to include one test for invalid_token and if possible for expired_token to improve the test coverage. I think for invalid_token you should create some mock data or environment variables for testing.

PR #72, opened 2021-09-28
Fix frontend queries

pylipp:
- Git-ignore react core file
- Update query implementations for frontend
- Run prettier


spaudanjo:
Hola @pylipp thanks! :) I aim to look into it today!

spaudanjo:
I couldn't resolve the issue yesterday ad-hoc. 
I propose the following: 

1) I would like to have a review/catchup at first of the latest schema changes at the beginning of the next hackathon (ideally with your help if you will join @pylipp - something like 20-30 minutes). 
2) Then I will look into fixing this test and the UI (doesn't work locally for me atm to create a new box) so we can merge the PR. 
3) I would also like to understand maybe already existing best practices (otherwise coming up of / aligning on some) for local manual testing/development (resetting mock data in the Mysql etc - currently it behaves a bit funny sometimes). 
4) And then also a review of the current app codebase, user flow and next frontend tickets as suggested last week. 

btw: looks like there is now a user-id (`createdBy`) needed when creating a new box, right?

aerinsol:
> btw: looks like there is now a user-id (`createdBy`) needed when creating a new box, right? 

Yes!

pylipp:
@spaudanjo I like your proposals!

The `createdBy` field has been removed again in #71 . I think it's improper design to let the client specify user information. Instead it is derived from the JWT authentication information (and hence handled on the backend).

PR #77, opened 2021-10-05
Add Beneficiary GraphQL schema

pylipp:
- Remove ineffective patching
- Remove complicated auth unit tests
- Consistently use builtin urllib module
- Store JWT custom claim information in Flask g object
- Use organisation ID from JWT for authorized lookup
- Remove redundant methods to retrieve DB data
- Select only authorized instances in bases query
- Authorize for product/location/organisation queries
- Derive user information from JWT data
- Add smarter resolvers
- Use pre-set JWT for running tests
- Make resolvers more selective about current user authorization
- Convert base ID in authorization_test function
- Clarify function to retrieve public key from JWKS
- Identify user by ID instead of email
- Align Beneficiary model definition with GraphQL fields
- Add createBeneficiary mutation and Beneficiary type
- Move CRUD operations into dedicated module
- Handle language(s) field of Beneficiary
- Store Beneficiary gender as single character in database


pylipp:
[Design and Implement Beneficiary and Shop in GraphQL Schema](https://trello.com/c/2dVKdoUv/743-design-and-implement-beneficiary-and-shop-in-graphql-schema)

aerinsol:
1. Yes, I think implicit treatment is best.
2. It's handled in the underlying MySql through a unique index. Therefore because it's a FK reference you just delete the FK language that matches the beneficiary id for that language.

vahidbazzaz:
Thanks @pylipp for the changes you made. 

@aerinsol @pylipp I found out that there are some business logics behind creating a beneficiary in DropApp, for instance, some fields are only available if the corresponding features are available on the base (such as `family-identifiers`, `maxfooddrops`, etc.). The same is true for box modification, where if the location changes, it should be indicated in the `itemsout` table.  We need to check if this should be handled in the authorization or in the business rule.

aerinsol:
Ugh the inconsistency in this db design 🤦🏻.

Put it in another card to limit scope: https://trello.com/c/ouJBZppq

Ban is not used in large majority of cases (deprecated feature for bicycles). Can default false or null.

PR #83, opened 2021-10-12
Add GraphQL mutation to create QR code

pylipp:
As specified in the trello ticket linked below, this PR implements the `createQrCode` mutation.
If the `boxLabelIdentifier` argument is not null, the QR code is associated with an existing box.

Question: can there be a scenario when the `boxLabelIdentifier` is passed in but no corresponding box exists? In that case the QR code creation would fail because peewee can't find the box for the specified identifier.

### DIFF
--- a/flask/boxwise_flask/exceptions.py
+++ b/flask/boxwise_flask/exceptions.py
@@ -28,3 +28,10 @@ class Forbidden(Exception):
             "user": user,
         }
         super().__init__(*args, **kwargs)
+
+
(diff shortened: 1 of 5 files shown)


vahidbazzaz:
@pylipp I believe the data input should be validated in the backend, so, even though this can be prevented on the front-end by only passing valid box identifier to this mutation, it would be necessary to validate the input or fulfill the request with a partial response that contains the errors.

PR #107, opened 2021-11-05
Add base locations resolver

pylipp:
This PR adds a resolver that I had previously missed to implement.

On a side-note @vahidbazzaz, I also checked the boxes resolver of the location query. It's not possible to retrieve unauthorized data since the location resolver checks whether the user has access to the location's base, and only if the base can be accessed, the `boxes` will be resolved by filtering w.r.t. the location.


vahidbazzaz (approved):
@pylipp, thank you for checking this

PR #109, opened 2021-11-08
App load testing and profiling

pylipp:
This PR contains some modifications to enable load-testing and profiling of the app.

The goal is to find out whether the app currently already shows severe bottlenecks for data IO, network, memory, and/or CPU usage.

### Experiments

On my local machine I ran two different experiments using the flask/mysql Docker services. For the results I don't put absolute numbers since they'll be different/non-comparable when running the experiments with the app deployed to GAE or similar.

- Load-testing using [k6](https://k6.io/docs/)
- Profiling the app

1. Load-testing
    1. The scenario is a running server running the webapp, and several clients sending **as many requests as possible**. We want to find out about the throughput (requests / sec), and the latency (i.e. request waiting time for client == processing time for server)
    1. I extended our Docker services to use the WSGI server `gunicorn` to resemble a production-level setup. For `gunicorn`, it's most important to [tune its number of worker processes](https://docs.gunicorn.org/en/stable/design.html#how-many-workers) acc. the available CPU resources
    1. I ran the `load-testing.js` script by `k6` for several combinations of virtual users, requests (single beneficiary vs. page of 100 beneficiaries) and gunicorn settings. The duration of the load-test is 15s.
    1. Result:
        - scaling workers in sensible manner improves both latency and throughput by a factor of 4
        - below some threshold (~nr of workers) increasing the number of users does not have an impact on app performance
        - latency for the beneficiaries query (50 DB rows read and returned) already noticeable: data IO plays a role
    1. Follow-up:
        - various smart and performant worker classes (`gevent`, `meinheld`) can be used in combination with gunicorn
        - Flask as a framework already has some overhead when processing requests. More minimalistic frameworks like [falcon](https://falcon.readthedocs.io/en/stable/user/intro.html#how-is-falcon-different) show [better performance](https://gist.github.com/nhymxu/814cf9b3294276629d2231248b709e26) but have a smaller community
1. Profiling
    1. I wanted to find out more about the internal workings in the app during a request. I could hook in the built-in `ProfilerMiddleware` for Flask, and visualize the resulting profile files using `tuna` and `snakeviz` (showing the hierarchy of Python calls and their expense)
    1. Findings:
        - for decoding the user JWT, we retrieve the public key of the boxtribute domain from Auth0. [Like others](https://community.auth0.com/t/where-are-jwks-key-ids-obtained-from-anywhere-in-the-auth0-gui/38890) we should consider **hard-coding the public key in our codebase or providing it via an env variable at least**
        - after having cut the public-key-retrieval overhead about 1/3 of the processing time is due to opening a MySQL connection at the start of processing the request. peewee provides support for [connection pooling](https://docs.peewee-orm.com/en/latest/peewee/playhouse.html#pool) but it has to be evaluated how efficient this is in combination with Flask
        - the remaining processing time is mostly spent processing the actual database query
        - peewee [caches](http://docs.peewee-orm.com/en/latest/peewee/querying.html#iterating-over-large-result-sets) rows returned by select queries. For large queries it is recommended to work-around this behavior
        - a query like `bene = Beneficiary.get_by_id(id)`, followed by `bene.base.id` causes a full select of the Base model. Better be explicit and do `Beneficiary.select(Beneficiary, Base.id).join(Base).where(Beneficiary.id == id).get()`. This can avoid overhead, e.g. due to formatting datetime in peewee.

### Conclusion

The requests used in the experiments are still fairly simply, and don't put the app under concerning pressure. Profiling revealed some overhead that we can easily reduce. I'm not sure how much benefit further optimizations (e.g. using an async framework) would bring (but they would take some effort for sure). On that topic, I found [articles](https://calpaterson.com/async-python-is-not-faster.html) that question the performance of (and benchmarks issued by) async frameworks.

### Follow-up questions:
1. How many (simultaneous) users do we currently have in dropapp?
1. What are the resources (CPU, memory) of the stack we deploy to?

### Tasks

- [x] optimize peewee usage (avoid additional select call when resolving ID for FK field)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,6 +3,8 @@ services:
     flask:
         build:
             context: ./flask
+            args:
+                flask_env: ${FLASK_ENV:-development}
         ports:
             - 5000:5000
             # request localhost:5001 to run debugger in vscode (cf. README)
@@ -12,7 +14,6 @@ services:
         volumes:
             - ./flask:/app/flask
         environment:
-            FLASK_ENV: development
             PYTHONUNBUFFERED: 1
             AUTH0_DOMAIN: ${AUTH0_DOMAIN:?auth0 config not set in .env file}
             AUTH0_AUDIENCE: ${AUTH0_AUDIENCE:?auth0 config not set in .env file}
--- a/flask/Dockerfile
+++ b/flask/Dockerfile
@@ -1,11 +1,28 @@
+ARG flask_env
+
 # using alpine for smallest docker image
-FROM python:3.8-alpine
+FROM python:3.8-alpine AS base
+
 WORKDIR /app/flask
 # install packages seperately to use docker build-caching
 # so we don't need to re-fetch these if code has changed
 # but our package.json hasn't
 COPY requirements.txt /app/flask
 RUN pip install -r requirements.txt && rm requirements.txt
+
+# Build two stages for the conditions production/development
+# cf. https://stackoverflow.com/a/60820156/3865876
+FROM base AS production
+# Launch WSGI server, configured by gunicorn.conf.py
+ENV FLASK_ENV production
+CMD gunicorn
+
+FROM base AS development
 # Path of module containing flask app, relative to WORKDIR
 ENV FLASK_APP boxwise_flask/main
+ENV FLASK_ENV development
 CMD flask run -h 0.0.0 -p 5000
+
+# Select final stage depending on flask_env argument
+FROM ${flask_env} AS final
+RUN echo Using $FLASK_ENV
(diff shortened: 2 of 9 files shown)


pylipp:
[Research on async support for back-end](https://trello.com/c/cz7slHyP/784-research-on-async-support-for-back-end)

pylipp:
[Production-level optimizations for new app](https://trello.com/c/CaipG15Q/786-production-level-optimizations-for-new-app)

pylipp:
I added follow-up trello cards (see above) and optimized some peewee queries for FK IDs. I think this is completed for now.
Mind to approve @vahidbazzaz @aerinsol or do you have more remarks?

vahidbazzaz (approved):
Nice job 👍

PR #111, opened 2021-11-10
Fix GraphQL resolvers and enums

pylipp:
As a preparation for future FE implementation, I had a look at the [previous](https://github.com/boxwise/boxtribute/blob/36529720613db62176bf3221613369bb0630ce8c/react/src/utils/queries.ts) query definitions, and run them in the playground. Indeed I could find two inconsistencies :)

Further comments on the queries:
- USER: parameter `id` instead of `email`
- ALL_BASES, BASE, ORG_BASES: nested field `organisation { id }` instead of `organisationId`
- BOX_BY_QR: misleading name since it fetches box by `boxLabelIdentifier`. `size` and `state` instead of `size_id` and `box_state_id`
- PRODUCTS: now a paginated query. Accepts PaginationInput argument and returns ProductPage type

@aerinsol @haguesto How does the dropapp handle the ProductGender with ID 10? It does not have any name associated with it. Should it be represented by null?

@spaudanjo I don't expect a complete review, this is rather an FYI =)
Can you think of any other queries/mutations that might be relevant for the FE MVP?

### DIFF
--- a/back/boxtribute_server/graph_ql/enums.py
+++ b/back/boxtribute_server/graph_ql/enums.py
@@ -5,7 +5,15 @@ product_gender_enum = EnumType(
     "ProductGender",
     {
         "Women": 1,
+        "Men": 2,
         "UnisexAdult": 3,
+        "Girl": 4,
+        "Boy": 5,
+        "UnisexKid": 6,
+        "UnisexBaby": 9,
+        "None": 10,
+        "TeenGirl": 12,
+        "TeenBoy": 13,
     },
 )
 
@@ -13,6 +21,11 @@ box_state_enum = EnumType(
     "BoxState",
     {
         "InStock": 1,
+        "Lost": 2,
+        "Ordered": 3,
+        "Picked": 4,
+        "Donated": 5,
+        "Scrap": 6,
     },
 )
 
--- a/back/boxtribute_server/graph_ql/pagination.py
+++ b/back/boxtribute_server/graph_ql/pagination.py
@@ -185,3 +185,27 @@ def generate_page(*conditions, elements, cursor, selection, **page_info_kwargs):
         page["elements"] = elements[1:] if page_info.has_previous_page else elements
 
     return page
+
+
+def load_into_page(model, *conditions, selection=None, pagination_input):
+    """High-level convenience function to load result query of given model into a
+    GraphQL page type.
+    The query is constructed from the given selection (default: `model.select()`), and
+    optional conditions. The query results are ordered by model ID.
+    """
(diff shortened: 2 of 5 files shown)


vahidbazzaz:
@pylipp It seems that product gender id 10 represents unspecified / unclassified products. I will double check with @HaGuesto.

pylipp:
@vahidbazzaz I added the 'unspecified' product gender. Do you mind approving if you have no more questions?

aerinsol:
I thought we were just going to enable it as null?

vahidbazzaz:
@aerinsol @pylipp If we want the value to be nullable, we can set it to null. Otherwise, an unclassified label would make more sense. What do you think?

aerinsol:
I think from a business logic sense it makes more sense to be null if something doesn't have a gender (e.g. toothpaste, food). But is this field currently designed as nullable?

vahidbazzaz:
@aerinsol the gender_id is not nullable in DB and there is currently corresponding gender id 10 with an empty label.

pylipp:
The DB field is not designed to be nullable (I agree that it would make sense from a business POV but that would require a small schema migration in dropapp first I think). However I can change the resolver for the `Product.gender` GraphQL field to return null if the underlying gender ID == 10. FE then must decide how to handle the null-gender.

aerinsol:
Wait, let me have a think. There are cases where clothes are unisex, but
water has no gender at all. Let me consider what the impressions are and if
they're different.

pylipp:
@aerinsol Any update here? We can also discuss after tmrw's standup.

pylipp:
We decided to have a long dash (—) as representation for product gender ID 10 but it is not allowed to use it in GraphQL enum definition. I suggest to keep the field name `Unspecified` and then the FE has to do conversion for display.

aerinsol:
Does GraphQL only allow alphanumeric? In that case I would like us to use
"None". Unspecified creates too many opportunities for spelling errors 😅

pylipp:
> Does GraphQL only allow alphanumeric? In that case I would like us to use "None". Unspecified creates too many opportunities for spelling errors sweat_smile

Acc. to the [spec](http://spec.graphql.org/October2021/#sec-Names) a GraphQL name can contain only alphanumeric chars and underscore.

pylipp:
Glad we're coming to an end here :slightly_smiling_face: @vahidbazzaz do you mind approving? Note that the reported decrease in coverage is due to the reduction of (covered) duplicated code. Diff coverage is 100%, meaning that all changes introduced by the PR are covered.

vahidbazzaz (approved):

PR #123, opened 2021-11-29
Implement DB models for box transfer

pylipp:
Here they come, the new data models for box-transfer!

Following the [ERD](https://drive.google.com/file/d/1Yhjk-KOJERpwfvkisWGa-hyDxv_BPkTE/view).

The PR can be divided into 
- preparation of test environment (simpler setting up of test data)
- actual testing and implementation of models

As usual, I have a couple of questions.
1. I added the state types as Python enums. I think it gives good flexibility and readability when in the application layer. But I also thought it has benefits to have the states defined as models in the data layer (just as we do for `BoxState`). This makes the data more complete on its own (no application layer needed to understand the data). What do you think?
1. Whenever possible, I implemented the model fields with sensible defaults *on database layer* (e.g. `DEFAULT CURRENT_TIMESTAMP` for mandatory datetime fields). Any objections?
1. In the TransferAgreement model, will the format for `valid_from/until` datetime fields be a date (e.g. 2021-11-30 without time)? I would add `00:00:00` for start and `23:59:59` for end time to indicate that the agreement is still valid on the entire last day.
1. When working with unsigned integer primary keys, I ran into some issues due to the model tests using Sqlite as database. As I next task I'll change the tests to always work with MySQL. I think it makes a lot of sense to run all tests with same the DB implementation as used in production.
1. As a note, coverage is reported as decreased. This is due to a test simplification resulting in the `BaseModule` model (aka. the `cms_functions` table) to be not covered anymore. I think this is not relevant since we want to get rid of it anyways, don't we?


vahidbazzaz:
> Here they come, the new data models for box-transfer!
> 
> Following the [ERD](https://drive.google.com/file/d/1Yhjk-KOJERpwfvkisWGa-hyDxv_BPkTE/view).
> 
> The PR can be divided into
> 
> * preparation of test environment (simpler setting up of test data)
> * actual testing and implementation of models
> 
> As usual, I have a couple of questions.
> 
> 1. I added the state types as Python enums. I think it gives good flexibility and readability when in the application layer. But I also thought it has benefits to have the states defined as models in the data layer (just as we do for `BoxState`). This makes the data more complete on its own (no application layer needed to understand the data). What do you think?
> 2. Whenever possible, I implemented the model fields with sensible defaults _on database layer_ (e.g. `DEFAULT CURRENT_TIMESTAMP` for mandatory datetime fields). Any objections?
> 3. In the TransferAgreement model, will the format for `valid_from/until` datetime fields be a date (e.g. 2021-11-30 without time)? I would add `00:00:00` for start and `23:59:59` for end time to indicate that the agreement is still valid on the entire last day.
> 4. When working with unsigned integer primary keys, I ran into some issues due to the model tests using Sqlite as database. As I next task I'll change the tests to always work with MySQL. I think it makes a lot of sense to run all tests with same the DB implementation as used in production.
> 5. As a note, coverage is reported as decreased. This is due to a test simplification resulting in the `BaseModule` model (aka. the `cms_functions` table) to be not covered anymore. I think this is not relevant since we want to get rid of it anyways, don't we?

@pylipp Thank you for implementing the data model for box transfer. Please find the following answers to your questions:

1. The enum makes more sense in terms of readability, and we don't need to add status on the fly here. This is much neater and more suitable.
2. Having a default time would be nice, but also consider these inputs as optional, so if they are not provided, then the default value is used. 
3. From what I understand, we just need a date, and I don't think we need the time-based restriction at the moment, but we can talk to @aerinsol about it. 
4. The decision to use MySQL makes sense, but we'll have to figure out how to keep track of migrations between Drop and the new stack. 
5. You can make these models invisible for the testers for the time being so they won't affect the test coverage and then remove them later on. What do you think?

pylipp:
Conclusions:
1. Use MySQL ENUM type.
3. Keep track of start time as date+00:00:00 and end time as date+23:59:59 (from POV of requesting org). Displayed as local time for user.

pylipp:
@vahidbazzaz, I have created a custom peewee `EnumField` that represents MySQL ENUM. However, because Sqlite does not support ENUM, I can't use it yet. So I'll integrate it in the process of 4. (updating the tests to entirely use MySQL instead of Sqlite).

pylipp:
I also did more research about the storing of timestamps. Quoting from [this recommendation](https://www.vertabelo.com/blog/the-proper-way-to-handle-multiple-time-zones-in-mysql/#recommended%20approach) (emphasis mine):

> I would recommend storing MySQL dates and times as a DATETIME column (read this to understand why) and storing the **time zone** (like “America/Los_Angeles”) in **another VARCHAR column**. Note that I am NOT recommending that you store the UTC offset; this is because of the DST changes explained earlier.

> Above all, **store the UTC time in the database**. This is the most accurate representation of a point in time you can have. Since you also have the time zone column, you will always have the ability to “reconstruct” the original local date/time, either via the MySQL CONVERT_TZ() function or from your application end

For the `TransferAgreement` model and its fields `valid_from` and `valid_until` the use cases, and my implementation proposals are:
1. User does not specify `valid_from`: let server create current UTC time and store it. When retrieving the FE will  display local time using user locale information.
2. User specifies `valid_from` (in local time). FE *must* derive user timezone information (e.g. [this way](https://stackoverflow.com/a/22625076/3865876)), and pass it along in the request. **Since the user gives a date like 2022-01-01, we assume the time as 00:00:00 (local).** The server combines the datetime and the timezone info to store a UTC timestamp. Retrieval as above.
3. User does not specify `valid_until`: store field as NULL (i.e. no expiration date)
4. User specifies `valid_until` (in local time). FE *must* pass user timezone information (cf. 2.). **Since the user gives a date like 2022-01-01, we assume the time as 23:59:59 (local).** The server combines the datetime and the timezone info to store a UTC timestamp. Retrieval as above.

Do you agree with this?

Further reading: 
- [best practices for timezone handling in web apps](https://stackoverflow.com/a/33362399/3865876)
- [this goes more into sensible display for the user](https://stackoverflow.com/a/10924961/3865876)

vahidbazzaz:
> I also did more research about the storing of timestamps. Quoting from [this recommendation](https://www.vertabelo.com/blog/the-proper-way-to-handle-multiple-time-zones-in-mysql/#recommended%20approach) (emphasis mine):
> 
> > I would recommend storing MySQL dates and times as a DATETIME column (read this to understand why) and storing the **time zone** (like “America/Los_Angeles”) in **another VARCHAR column**. Note that I am NOT recommending that you store the UTC offset; this is because of the DST changes explained earlier.
> 
> > Above all, **store the UTC time in the database**. This is the most accurate representation of a point in time you can have. Since you also have the time zone column, you will always have the ability to “reconstruct” the original local date/time, either via the MySQL CONVERT_TZ() function or from your application end
> 
> For the `TransferAgreement` model and its fields `valid_from` and `valid_until` the use cases, and my implementation proposals are:
> 
> 1. User does not specify `valid_from`: let server create current UTC time and store it. When retrieving the FE will  display local time using user locale information.
> 2. User specifies `valid_from` (in local time). FE _must_ derive user timezone information (e.g. [this way](https://stackoverflow.com/a/22625076/3865876)), and pass it along in the request. **Since the user gives a date like 2022-01-01, we assume the time as 00:00:00 (local).** The server combines the datetime and the timezone info to store a UTC timestamp. Retrieval as above.
> 3. User does not specify `valid_until`: store field as NULL (i.e. no expiration date)
> 4. User specifies `valid_until` (in local time). FE _must_ pass user timezone information (cf. 2.). **Since the user gives a date like 2022-01-01, we assume the time as 23:59:59 (local).** The server combines the datetime and the timezone info to store a UTC timestamp. Retrieval as above.
> 
> We don't need to store any timezone information in the TransferAgreement model though.
> 
> Do you agree with this?
> 
> Further reading:
> 
> * [best practices for timezone handling in web apps](https://stackoverflow.com/a/33362399/3865876)
> * [this goes more into sensible display for the user](https://stackoverflow.com/a/10924961/3865876)

@pylipp Thank you for your very nice research. I think choosing UTC timezone is an excellent choice, and I believe this will work well with the use cases we may encounter. Nice work!

pylipp:
[Implement DB models for box-transfer](https://trello.com/c/Ujzps8Mn/791-implement-db-models-for-box-transfer)

pylipp:
Another conclusion:
1. We DO NOT use MySQL ENUM since it's not in the SQL standard
2. There are two options for implementing enums:
    i. have a state table listing enum values and labels; then have a FK referencing that table. On the application layer, the enum structure is replicated (:heavy_minus_sign:), also for conversion from/to GraphQL
    ii. have the application layer define the enum, and store values directly in state IntegerField. On the data layer, any semantics of the state integer are lost (:heavy_minus_sign:)

@vahidbazzaz raises a point about the user-extensibilty of the state field. I now think it's possible with both approaches, probably even simpler with the 2nd approach because it does not require a database table update. However I'll do more research about best practices.

pylipp:
After [quick research](https://stackoverflow.com/a/229919/3865876) it looks like 
- approach 2.i. is not favored because any changes of the enum definition require updates in both data layer and app layer
- storing enum *names* as string field in the DB preserves both the readability of the data, and integrates nicely with the app layer (i.e. Python enums). In terms of disk space it might be cheaper to store a little integer instead of a string but I think we can neglect that

Hence my proposal: store state as string field in table :tada:

aerinsol:
@pylipp @vahidbazzaz I approve

vahidbazzaz:
> After [quick research](https://stackoverflow.com/a/229919/3865876) it looks like
> 
> * approach 2.i. is not favored because any changes of the enum definition require updates in both data layer and app layer
> * storing enum _names_ as string field in the DB preserves both the readability of the data, and integrates nicely with the app layer (i.e. Python enums). In terms of disk space it might be cheaper to store a little integer instead of a string but I think we can neglect that
> 
> Hence my proposal: store state as string field in table 🎉 @vahidbazzaz @aerinsol
> 
> PS: a [different opinion](https://stackoverflow.com/a/6127838/3865876) objects that the string-field approach is troublesome when it comes to renaming of enum names (but it would require a simple migration only!?)

@pylipp Thanks for the lookup and the helpful summary of options you provided. Since ENUM is not an SQL's standard datatype and not many database systems provide support to it, Peewee does not support it, plus there are concerns around it's usage (modify value in production, [sorting based enum field](https://dev.mysql.com/doc/refman/8.0/en/enum.html#enum-sorting)), it is more sensible to use status string (which is not nullable), and enforce validation at the application layer.

back/boxtribute_server/models/enums.py
```diff
@@ -0,0 +1,46 @@
+"""Enumeration types used in the database."""
+import enum
+
+from peewee import CharField
+
+
+class EnumCharField(CharField):
```
pylipp, line 7: FYI @aerinsol This is the custom peewee field type to represent Python enums as varchar on data level.

vahidbazzaz (approved):

PR #127, opened 2021-12-06
Replace Sqlite by MySQL for testing

pylipp:
With this PR, the test suite uses MySQL entirely.
The advantage is not be surprised anymore by unexpected behavior because Sqlite treats types or FKs differently.
The drawback is longer test execution time (on my machine, 150s vs. 55s for the entire test suite; in CI 38s vs. 16s). I suspect it's due to not using an in-memory database but instead communicate with an actual MySQL server (network and IO overhead).

Other highlights:
- database is configured to use unsigned integer PKs (introduce custom types for peewee ForeignKeyField)
- Python enum classes for BoxState, ProductGender, et. al
- datetimes returned via GraphQL API are timezone-aware (UTC)
- strip fractional-second parts from datetimes before writing to database (seems to be configured for second precision)
- handle invalid pagination parameters
- restructured application code files for consistency
- simplify data model unit tests into single test

Now the questions =) (also @aerinsol )
1. Since the new app does not need them anymore, and we've moved to Auth0 for managing authz, I'd remove the data models `Usergroup`, `UsergroupModuleAccess`, `UsergroupBaseAccess`, `UsergroupAccessLevel`, and `BaseModule` (the corresponding tables are `cms_usergroups`, ..., `cms_usergroup_functions`). Any objections?
1. ~Our data model currently has some circular dependencies, e.g. `Organisation` has FK fields `created_by`, `modified_by` to `User`, and in turn `User` has the FK field `organisation`. peewee brings the [`DeferredForeignKey`](http://docs.peewee-orm.com/en/latest/peewee/api.html#DeferredForeignKey) as a remedy, still it's considered a bit of a code smell in the first place. Has it ever been considered non-ideal to have this circular dependency?~ I was thinking that in the future, the `User` model might not be needed anyways since user information is pulled from Auth0 (in the JWT for the current user, or using their [API](https://auth0.com/docs/users/user-search/retrieve-users-with-get-users-endpoint).


pylipp:
[Use MySQL entirely for testing new app](https://trello.com/c/wrw6svC3/801-use-mysql-entirely-for-testing-new-app)

pylipp:
@vahidbazzaz This is now finalized :sweat_smile: 
Please let me know about question 2., also @aerinsol (or we talk about it next week. It's not a blocker for the PR).

pylipp:
> @vahidbazzaz This is now finalized sweat_smile Please let me know about question 2., also @aerinsol (or we talk about it next week. It's not a blocker for the PR).

Actually nevermind, question 2. was caused by confusion on my end. We do _not_ have circular dependencies in the data models. I had issues with FKs when creating and populating all models during test setup but that was due to not taking the inter-dependencies of the data models into account. This is fixed now in feabeb4.

pylipp:
> Nice job, @pylipp! I have run all tests locally and they all work fine, just two have failed on my end, which seems to be due to my end.

Are those integration tests? Since they interact with Auth0 they need some env variables in `.env` to be properly set up. I think especially `AUTH0_USERNAME=dev_coordinator@boxaid.org ` is required (the test expects a user with dev coordinator permissions)

PR #131, opened 2021-12-15
Use smarter fixtures for testing database

pylipp:
Ever since #127, I was a bit annoyed with the increase in test execution time due to the overhead of creating fresh MySQL tables for *every* test function.

With this PR I introduce smarter fixtures (using two dedicated databases for read-only and arbitrary operations) which brings down test execution time from 150s to 30s on my machine. Yay! I think this is even faster than the Sqlite-based test suite before.

No functional changes.


vahidbazzaz (approved):
Nice job @pylipp on the test speed gain.

PR #132, opened 2021-12-16
Implement GraphQL queries for box-transfer

pylipp:
That worked well...setting up the queries to fetch transfer agreements and shipments via GraphQL

For filtering, the following applies:
- filtering by organisation happens implicitly by only returning agreements/shipments that the current client's organisation is involved in (as either source or target)
- agreements can be filtered by state
- shipments cannot be further filtered

Do you agree with these decisions?
Anyways we can always extend the filtering logic in the process of developing the MVP and beyond =)

https://trello.com/c/e7vZrQAP/804-implement-graphql-queries-for-box-transfer

### DIFF
--- a/back/boxtribute_server/enums.py
+++ b/back/boxtribute_server/enums.py
@@ -18,7 +18,7 @@ class TransferAgreementType(enum.IntEnum):
 class ShipmentState(enum.IntEnum):
     Preparing = 1
     Sent = enum.auto()
-    Received = enum.auto()
+    Completed = enum.auto()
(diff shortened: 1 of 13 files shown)


aerinsol:
Yep, that's fine for now.

vahidbazzaz (approved):

pylipp:
[Implement GraphQL queries for box-transfer](https://trello.com/c/e7vZrQAP/804-implement-graphql-queries-for-box-transfer)

aerinsol (approved):

PR #133, opened 2021-12-18
810 fix app setup with docker on mac and upgrade mysql to v8.0

spaudanjo:
closes https://trello.com/c/yZXAchvk/810-fix-docker-app-setup-for-macs


pylipp (approved):
I think you found all occurrences of the port number ;)
thanks for the mysql version update, too! tests are not breaking, yay :tada:

PR #137, opened 2021-12-20
GraphQL mutations for transfer-agreement flow

pylipp:
This PR implements four mutations for managing transfer agreements.

Again I did not validate operations on the specified agreement, e.g.
- only `UnderReview` agreements can be accepted
- only `UnderReview` agreements can be rejected
- only `UnderReview`/`Accepted` agreements can be canceled
- only a member of the target organisation can accept/reject

Should I add these validations to avoid inconsistencies or possible bugs? What kind of error would a failed validation be, a bad user input, or ...?

@aerinsol Note that I extended the test plan (integration test sheet), and linked the test case IDs in a new column in the box-transfer user action sheet.


pylipp:
[Implement GraphQL mutations for transfer-agreement flow](https://trello.com/c/6URigGX1/807-implement-graphql-mutations-for-transfer-agreement-flow)

aerinsol:
I honestly think in terms of errors you can do exactly that. The users would not always be the end user here, but would be front end devs too.

aerinsol (changes requested):
Any way we can pass in what the requested action is?

pylipp:
> Any way we can pass in what the requested action is?

Sorry, I force-pushed, and overwrote the commit that you might have referred to. Can you have another look?

back/boxtribute_server/exceptions.py
```diff
@@ -51,6 +51,22 @@ class InvalidTransferAgreement(Exception):
     }
 
 
+class InvalidTransferAgreementState(Exception):
+    extensions = {
+        "code": "BAD_USER_INPUT",
+        "description": "The state of the transfer agreement does not allow the "
+        "requested action.",
```
aerinsol, line 58: @pylipp can we pass in what "requested action" is? This is not particularly descriptive.

aerinsol (changes requested):

back/boxtribute_server/exceptions.py
pylipp: I assume this error message would be presented to the FE devs rather than to the end users. The full response from GraphQL looks like
```
{
  "data": {
    "acceptTransferAgreement": null
  },
  "errors": [
    {
      "extensions": {
        "code": "BAD_USER_INPUT",
        "description": "The state of the transfer agreement does not allow the requested action."
      },
      "locations": [
        {
          "column": 3,
          "line": 2
        }
      ],
      "message": "An unknown error occurred.",
      "path": [
        "acceptTransferAgreement"
      ]
    }
  ]
}
```
and here it's visible that `acceptTransferAgreement` was executed.
If we consider using the `description` alone (by displaying it to the end user but really the cause of the error should be fixed during FE development already I think) I can add the action name.

pylipp:
@aerinsol The error description now holds the actual state of the involved agreement, and a list of expected states: e.g. `The state of the transfer agreement (Expired) does not allow the requested action. Expecting state: Accepted`

vahidbazzaz (approved):

aerinsol (approved):

PR #138, opened 2021-12-23
GraphQL mutations for shipment flow

pylipp:
This PR adds the implementation for the `createShipment`, `cancelShipment`, `sendShipment` mutations, incl. appropriate validations (e.g. for shipment state).

The corresponding test cases are 3.2.1 - 3.2.14 in https://docs.google.com/spreadsheets/d/1sDhSsaVwNxAhGn1VYACDPVepZpvlwTkgy9nmeTaRy2Q/edit#gid=1709672082 (last sheet) with the exception of the cases dealing with non-existent resources (will be handled in generic way).

I have two questions regarding these test cases:
- *3.2.10* Is it correct that only a member of the organisation that originally created the shipment can cancel that shipment?
- *3.2.14* For validating that the confirming user is member of the organisation that created the shipment I first need to retrieve information about the creating organisation. We don't explicitly store that in the Shipment table. Instead I'd have to use the ID of the user who created the shipment, and query Auth0 for the organisation that the user belongs to. This creates some overhead but I don't see another way (without duplicating information about the user's organisation in the User table). Or do you think it's an edge case we don't have to take care of?

### DIFF
--- a/back/boxtribute_server/exceptions.py
+++ b/back/boxtribute_server/exceptions.py
@@ -44,15 +44,26 @@ class RequestedResourceNotFound(Exception):
     }
 
 
-class InvalidTransferAgreementState(Exception):
-    def __init__(self, expected_states, actual_state, *args, **kwargs):
+class _InvalidResourceState(Exception):
+    def __init__(self, resource_name, *args, expected_states, actual_state, **kwargs):
         self.extensions = {
             "code": "BAD_USER_INPUT",
-            "description": f"The state of the transfer agreement ({actual_state.name}) "
+            "description": f"The state of the {resource_name} ({actual_state.name}) "
             "does not allow the requested action. Expecting state: "
             f"{' or '.join(s.name for s in expected_states)}",
         }
         super().__init__(*args, **kwargs)
+
+
+class InvalidTransferAgreementState(_InvalidResourceState):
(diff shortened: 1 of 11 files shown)


pylipp:
[Implement GraphQL mutations for shipment flow](https://trello.com/c/p2fMrD8M/812-implement-graphql-mutations-for-shipment-flow)

pylipp:
Conclusion after standup with @aerinsol  and @vahidbazzaz :

> _3.2.10_ Is it correct that only a member of the organisation that originally created the shipment can cancel that shipment?

No. Members of both organisations can cancel (source might say "we can't send anymore", target might say "we don't want to receive that anymore").

> _3.2.14_ For validating that the confirming user is member of the organisation that created the shipment I first need to retrieve information about the creating organisation. We don't explicitly store that in the Shipment table. Instead I'd have to use the ID of the user who created the shipment, and query Auth0 for the organisation that the user belongs to. This creates some overhead but I don't see another way (without duplicating information about the user's organisation in the User table). Or do you think it's an edge case we don't have to take care of?

Simple solution: use cross-reference to Organisation table using given shipment's source base.

vahidbazzaz (approved):

pylipp:
FYI test-case 3.2.14 is implemented in the follow-up PR (already started it, and would have conflicts in test file otherwise)

PR #140, opened 2021-12-29
GraphQL mutations for update-shipment flow

pylipp:
This PR implements part II of the GraphQL mutations for the box-transfer feature.

The corresponding test cases are 3.2.20 - 3.2.32.

One question (cf. test case 3.2.25): can members of both organisations update the bases of an existing shipment?

Note that in the implementation for test cases 3.2.27, .28, .31, no errors are indicated in the GraphQL response (I didn't see any way to return a partial response because only one resolver (for `updateShipment` mutation is involved. It would be possible to return a custom `ShipmentUpdateResponse` type holding successful and erroneous updates. I don't how much benefit it brings since that part of the API won't be public).

### DIFF
--- a/back/boxtribute_server/models/crud.py
+++ b/back/boxtribute_server/models/crud.py
@@ -431,7 +431,8 @@ def update_shipment(
     Raise an InvalidTransferAgreementBase exception if specified source or target base
     are not included in given agreement.
     If boxes are requested to be updated that are not located in the shipment's source
-    base, they are silently discarded (i.e. not added to the ShipmentDetail model).
+    base, or have a state different from InStock, they are silently discarded (i.e. not
+    added to the ShipmentDetail model).
     """
     shipment = Shipment.get_by_id(id)
     if shipment.state != ShipmentState.Preparing:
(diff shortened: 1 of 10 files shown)


pylipp:
[Implement GraphQL mutations for update-shipment flow](https://trello.com/c/Bj3rX2kg/813-implement-graphql-mutations-for-update-shipment-flow)

back/boxtribute_server/models/crud.py
```diff
@@ -317,6 +318,29 @@ def retrieve_transfer_agreement_bases(*, transfer_agreement, kind):
     )
 
 
+def _validate_bases_as_part_of_transfer_agreement(
+    *, transfer_agreement, source_base_id=None, target_base_id=None
+):
+    """Validate that given bases are part of the given transfer agreement. Raise
+    InvalidTransferAgreementBase expection otherwise.
```
pylipp, line 325: typo: exception

vahidbazzaz (approved):

pylipp:
> One question (cf. test case 3.2.25): can members of both organisations update the bases of an existing shipment?

Clarified with @aerinsol and @vahidbazzaz : only the organisation that created the shipment can update the target base. The source base cannot be updated. I updated the [user actions](https://docs.google.com/spreadsheets/d/1WfdnhRIV2OXbCnUYtqCe0DigdSJbReZPunZotlHvW9c/edit#gid=1145101562) accordingly. Implementation of that change will be done in a follow-up branch which is already WIP.

PR #141, opened 2022-01-05
GraphQL mutation for complete shipment flow

pylipp:
This PR brings the implementation for the remaining GraphQL API parts for handling shipments.

On the target side
- boxes can be marked as received and have target product and location assigned
- boxes can be marked as lost

This is possible with the `updateShipment` mutation using the `receivedShipmentDetailUpdateInputs` and `lostBoxLabelIdentifiers` fields, resp.
If all boxes are marked as received or lost, they are moved into the target stock, and the shipment is marked as completed.

The implemented test cases are 3.2.34 - 3.2.42. I consider the implementation of the box-transfer feature finished, despite some clean-up in a follow-up PR (modularize business logic, reduce repetitive patterns in tests). Let me know about any questions!


pylipp:
[Implement GraphQL mutations for update-shipment flow, part II](https://trello.com/c/b2N3wpA5/817-implement-graphql-mutations-for-update-shipment-flow-part-ii)

back/boxtribute_server/models/crud.py
```diff
@@ -453,69 +475,204 @@ def _update_shipment_with_prepared_boxes(*, shipment, box_label_identifiers, use
...
+        received_boxes = []
+        for detail in details:
+            detail.deleted_by = user_id
+            detail.deleted_on = now
+            detail.box.product = detail.target_product
+            detail.box.location = detail.target_location
+            if detail.box.state_id == BoxState.Received:
+                detail.box.state = BoxState.InStock
+                received_boxes.append(detail.box)
+
```
vahidbazzaz, line 585: @pylipp Would it make sense to wrap these updates in transactions, since it seems to me that they should be atomic? What do you think?

pylipp: The function `_complete_shipment_if_applicable` is called from within an `atomic` context manager (lines 641-664). I think that should be fine then! Very glad though that you take some scrutiny for the review :D !

vahidbazzaz (approved):

PR #144, opened 2022-01-10
Integrate new permission format

pylipp:
- Make test data and its usage more consistent
- Integrate new Auth0 permissions format

Some ideas to decrease the JWT payload size for @vahidbazzaz
- only send `:write` or `:edit` permission (I can derive the corresponding `:read` permission)
- do we need `box_state:read`, `category:read`, `gender:read`, `language:read`, `sizegroup:read`? The underlying data is not user-specific. Even if the permissions were part of the payload, I would not enforce them in the BE for simplicity.

What do you think?

### DIFF
--- a/back/boxtribute_server/auth.py
+++ b/back/boxtribute_server/auth.py
@@ -104,6 +104,13 @@ def requires_auth(f):
     If authentication succeeds, user information is extracted from the JWT payload into
     the `user` attribute of the Flask g object. It is then available for the duration of
     the request.
+    The `permissions` field of `g.user` is a mapping of a permission name to a list of
+    base IDs that the permission is granted for, or to None if the permission is granted
+    for all bases. It is parsed from the `permissions` custom claim which contains
+    entries of form '[base_X[-Y...]/]resource:method', e.g.
+    - base_1/product:read    -> {"product:read": [1]}
+    - base_2-3/stock:write   -> {"stock:write": [2, 3]}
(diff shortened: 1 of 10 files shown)


pylipp:
[Integrate new permission format in back-end](https://trello.com/c/Vmlo9DVt/818-integrate-new-permission-format-in-back-end)

pylipp:
@spaudanjo once this PR is merged, can you try running the example again that you had an authz error with lately? Should be fixed :innocent: (or you try it on this branch)

vahidbazzaz:
> * Make test data and its usage more consistent
> * Integrate new Auth0 permissions format
> 
> Some ideas to decrease the JWT payload size for @vahidbazzaz
> 
> * only send `:write` or `:edit` permission (I can derive the corresponding `:read` permission)
> * do we need `box_state:read`, `category:read`, `gender:read`, `language:read`, `sizegroup:read`? The underlying data is not user-specific. Even if the permissions were part of the payload, I would not enforce them in the BE for simplicity.
> 
> What do you think?

@pylipp Thank you for your suggestions and for integrating the new permissions pattern in the BE. It definitely makes sense to not include read permission in JWT claim when edit or write permission has been granted. Regarding your second point, I agree, but we do not need to remove these permissions, maybe there is an edge case in the future.

vahidbazzaz (approved):
@pylipp Thanks!

PR #150, opened 2022-01-21
Use correct permissions

pylipp:
In #148, the support for the new format of DB-based permissions was implemented but I missed to adjust actual new permissions (`qr:create`, `beneficiary:create` and `edit`). This happened because the integration test that uses an actual Auth0 JWT was quite small before. I've extended it to cover most mutations and queries.

Also small fixes in schema and resolvers.

Merging before the hackathon would be beneficial because otherwise the `create/updateBeneficiary` and `createQrCode` queries are broken on master.


vahidbazzaz (approved):
Thanks @pylipp for correcting the permissions.

pylipp:
[Integrate new permission format in back-end](https://trello.com/c/Vmlo9DVt/818-integrate-new-permission-format-in-back-end)

PR #154, opened 2022-01-23
825 - add some first basic tests to locations view

spaudanjo:
contributes to https://trello.com/c/CYiylASq/825-add-basic-test-for-skeleton-for-fe-app


vahidbazzaz (approved):
@spaudanjo Thanks for adding the basic tests.

PR #157, opened 2022-01-25
GraphQL filters and fixes

pylipp:
This PR brings an extension of the GraphQL API:
- queries for beneficiaries and boxes can now be filtered (see the new filter types in `types.graphql` @aerinsol)
- the `Beneficiary` type is extended by fields `age`, `active` (indicate whether soft-deleted or not), and `transactions`

Let me know about any remarks.

### DIFF
--- a/back/boxtribute_server/graph_ql/filtering.py
+++ b/back/boxtribute_server/graph_ql/filtering.py
@@ -33,10 +33,10 @@ def derive_beneficiary_filter(filter_input):
             Beneficiary.is_volunteer if is_volunteer else ~Beneficiary.is_volunteer
         )
 
-    is_registered = filter_input.get("is_registered")
-    if is_registered is not None:
+    registered = filter_input.get("registered")
+    if registered is not None:
         condition &= (
-            ~Beneficiary.not_registered if is_registered else Beneficiary.not_registered
+            ~Beneficiary.not_registered if registered else Beneficiary.not_registered
         )
 
     pattern = filter_input.get("pattern")
--- a/back/boxtribute_server/graph_ql/filtering.py
+++ b/back/boxtribute_server/graph_ql/filtering.py
@@ -1,4 +1,5 @@
 from ..models.definitions.beneficiary import Beneficiary
+from ..models.definitions.box import Box
 
 
 def derive_beneficiary_filter(filter_input):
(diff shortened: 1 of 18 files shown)


pylipp:
[Export Age instead of DOB for Beneficiaries (requested by Imi O'Neill IHA)](https://trello.com/c/CMPAyymP/595-export-age-instead-of-dob-for-beneficiaries-requested-by-imi-oneill-iha)

pylipp:
[EPIC: Expose API on New App to Our Partners](https://trello.com/c/mnuzzEcb/814-epic-expose-api-on-new-app-to-our-partners)

vahidbazzaz (approved):
@pylipp thanks for adding the input filter and scalar field for calculating age. Although at first I thought it might be problematic to compute age for those whose birth dates are missing, it doesn't appear to be a problem.

pylipp:
> @pylipp thanks for adding the input filter and scalar field for calculating age. Although at first I thought it might be problematic to compute age for those whose birth dates are missing, it doesn't appear to be a problem.

So you're saying that technically, beneficiary DoB can be null (as per table definition) but practically it is always set? I checked the production database for `SELECT id,deleted FROM people WHERE date_of_birth IS NULL` which returns 58 rows but all are marked as deleted.

vahidbazzaz:
> > @pylipp thanks for adding the input filter and scalar field for calculating age. Although at first I thought it might be problematic to compute age for those whose birth dates are missing, it doesn't appear to be a problem.
> 
> So you're saying that technically, beneficiary DoB can be null (as per table definition) but practically it is always set? I checked the production database for `SELECT id,deleted FROM people WHERE date_of_birth IS NULL` which returns 58 rows but all are marked as deleted.

@pylipp It's true, but I remember the same issue for child and adult token distributions, as there were beneficiaries without birth data. Maybe @aerinsol could provide further insight into this.

aerinsol:
@pylipp @vahidbazzaz Previous version of MySQL didn't allow for null values on dates. Try the 0000 value - exact format linked in the Information section of the Trello board.

aerinsol:
@pylipp  @vahidbazzaz current thinking on null or empty dates: https://stackoverflow.com/questions/1691117/how-to-store-null-values-in-datetime-fields-in-mysql. Maybe we should treat this properly in the new stack.

pylipp:
I pushed an update for properly converting MySQL zero datetime values into `None` on application level. Let me know whether the business logic needs to support beneficiaries with unset DoB (database would support it) @aerinsol

aerinsol:
Yes, it's needed

vahidbazzaz (approved):

pylipp (commented):
Dropping Python 3.6 support, need to investigate

pylipp:
[New app: GraphQL query filters](https://trello.com/c/6HoMt6yY/831-new-app-graphql-query-filters)

PR #165, opened 2022-02-09
765 Box state management

leo-labs:
Automatic box state management, for spec details see https://trello.com/c/P09XKhJE

I changed `create_box` and `update_box` to take an explicit argument list decouple frontend/graphql from the service layer. Also removed passing of dictionaries to the data layer/peewee decoupling the data and service layer too.

@pylipp Let me know if the tests should be moved somewhere else and/or split into multiple.


back/boxtribute_server/graph_ql/resolvers.py
```diff
@@ -332,18 +331,32 @@ def resolve_create_qr_code(_, info, box_label_identifier=None):
...
     authorize(permission="stock:write")
-    box_creation_input["created_by"] = g.user["id"]
-    return create_box(box_creation_input)
+    return create_box(
+        product_id=box_creation_input["product_id"],
+        location_id=box_creation_input["location_id"],
+        comment=box_creation_input["comment"],
+        items=box_creation_input["items"],
+        size_id=box_creation_input["size_id"],
+        qr_code_code=box_creation_input.get("qr_code"),
```
pylipp, line 342: I wouldn't mind to have the detailed parameters hidden away by extracting `**box_creation_input` on the resolver level here. The explicit passing of parameters has a bit of redundancy and verbosity. If one wants to know the exact signature of `create_box` they can inspect its definition. I've used this approach in the resolvers for all `create`/`update` operations.

back/boxtribute_server/graph_ql/resolvers.py
```diff
@@ -332,18 +331,32 @@ def resolve_create_qr_code(_, info, box_label_identifier=None):
...
 @mutation.field("updateBox")
 @convert_kwargs_to_snake_case
-def resolve_update_box(_, info, box_update_input):
+def resolve_update_box(*_, box_update_input):
     authorize(permission="stock:write")
-    box_update_input["last_modified_by"] = g.user["id"]
-    return update_box(box_update_input)
+    return update_box(
+        label_identifier=box_update_input["label_identifier"],
+        updated_by_id=g.user["id"],
```
pylipp, line 353: In other resolvers, I used `user_id` as generic argument name, can you adjust it here?

back/boxtribute_server/graph_ql/resolvers.py
```diff
@@ -332,18 +331,32 @@ def resolve_create_qr_code(_, info, box_label_identifier=None):
 
 @mutation.field("createBox")
 @convert_kwargs_to_snake_case
-def resolve_create_box(_, info, box_creation_input):
+def resolve_create_box(*_, box_creation_input):
```
pylipp, line 334: I like the `*_` hiding of unused arguments. Already on the list for refactoring :smile:

back/boxtribute_server/models/crud.py
```diff
@@ -9,33 +9,49 @@
...
+    comment="",
+    items=0,
+    qr_code_code=None,
+):
     """Insert information for a new Box in the database. Use current datetime
     and box state "InStock" by default. Generate an 8-digit sequence to identify the
     box. If the sequence is not unique, repeat the generation several times. If
     generation still fails, raise a BoxCreationFailed exception.
     """
+
```
pylipp, line 34: Do you mind adding a sentence to the docstring about the role of location box-state?

back/boxtribute_server/models/crud.py
```diff
@@ -9,33 +9,49 @@
...
     now = utcnow()
-    qr_code = data.pop("qr_code", None)
-    qr_id = QrCode.get_id_from_code(qr_code) if qr_code is not None else None
+    qr_id = QrCode.get_id_from_code(qr_code_code) if qr_code_code is not None else None
 
+    location_box_state = Location.get(Location.id == location_id).box_state
```
pylipp, line 38: You can use the convenience method `Location.get_by_id(location_id)`. By using `box_state_id` you can avoid that peewee performs a full `SELECT` of the BoxState FK model, instead only the integer ID is returned (the docs hide it in [this section](http://docs.peewee-orm.com/en/latest/peewee/models.html#foreignkeyfield); noted to myself that I add this to the Readme about peewee quirks :slightly_smiling_face: )

back/test/data/location.py
```diff
@@ -35,11 +36,37 @@ def another_location_data():
     return data
 
 
+def null_box_state_location_data():
+    data = default_location_data().copy()
```
pylipp, line 40: Copying is redundant because the `data()` function returns a new dictionary at every call.

back/test/model_tests/test_crud.py
```diff
@@ -50,6 +45,54 @@ def test_box_label_identifier_generation(
...
+    # InStock
+    box = create_box(
+        product_id=default_product["id"],
+        created_by_id=default_user["id"],
+        location_id=null_box_state_location["id"],
+        size_id=default_size["id"],
+    )
+    assert box.state.id == BoxState.InStock
+
+    # updating to a location with box_state!+NULL should set the box state on the box
```
pylipp, line 70: probably typo, `box_state!=NULL`

pylipp (changes requested):
Thanks for this really helpful improvement :+1: I find the explicit signature of the `crud` functions increases readability quite a bit.
The introduction of the conditions in `update_box` made coverage decrease, can you add another test (I'd prefer extending `test_box_mutations` in `endpoint_tests/test_box.py`)?

pylipp:
[Update box state when changing box location](https://trello.com/c/P09XKhJE/765-update-box-state-when-changing-box-location)

back/boxtribute_server/graph_ql/resolvers.py
leo-labs: I am fine with this, but it basically limits us for now to use the same parameter names everywhere (which might not be too bad). E.g. see `qr_code` vs `qr_code_code`.

pylipp: Mhh ya I saw that you did a remapping. Maybe a comment in the signature of `create_box` helps? e.g. `qr_code=None  # code string of a QrCode object` or similar

back/test/model_tests/test_crud.py
```diff
@@ -19,29 +20,23 @@ def test_create_qr_code_for_nonexisting_box():
     assert nr_qr_codes == len(QrCode.select())
 
 
-def test_create_box_with_insufficient_data():
```
pylipp, line 22: I suppose if you add this test back, codecov stops complaining :smile: @leo-labs

leo-labs: I was already on it 🚀

back/boxtribute_server/graph_ql/resolvers.py
leo-labs: I just renamed it back, it is pretty clear that this is not a peewee object from the context here.

pylipp (approved):

PR #166, opened 2022-02-10
Implement GraphQL metrics query

pylipp:
This PR brings a `metrics` GraphQL query and type. The supported fields correspond to the 2nd-5th SQL queries listed [here](https://docs.google.com/document/d/10GFOZaWMKJCD4PAt6WSvIiBCWI09319CdLDU1PjZjWk/edit).
- computation of served families incl. filtering
- computation of sales incl. filtering
- computation of stock overview
- computation of moved stock overview

The product list can be fetched via the GraphQL API in a straightforward way.

@aerinsol
1. I settled for the name `metrics` instead of `statistics`. Imho a statistical value is e.g. an average or a variance but not a counted number. Any objections?
1. What is the difference between the first and the second query in the document?
1. When computing the number of sales, should we add `Transaction.tokens > 0` to avoid counting the transactions that actually represent granted tokens?
1. All queries return info for the current client's organisation. Should the god user be capable of fetching data for any organisation?
1. I compared the SQL queries generated by peewee with the ones from the document. They match in functionality I reckon. I still have to perform a sanity comparison by running the SQL queries directly on the production database.

As usual let me know about any remarks :)

### DIFF
--- a/back/README.md
+++ b/back/README.md
@@ -249,8 +249,8 @@ and inspect the reported output. Open the HTML report via `back/htmlcov/index.ht
 
 ## GraphQL API
 
-The back-end exposes the GraphQL API at the `/graphql` endpoint.
-It is consumed by our front-end, and by partners (for data retrieval).
+The back-end exposes the GraphQL API in two variants.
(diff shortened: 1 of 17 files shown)


pylipp:
Closing because GitHub messed up included files with previous PR

PR #167, opened 2022-02-10
Implement GraphQL metrics query

pylipp:
This PR brings a `metrics` GraphQL query and type. The supported fields correspond to the 2nd-5th SQL queries listed [here](https://docs.google.com/document/d/10GFOZaWMKJCD4PAt6WSvIiBCWI09319CdLDU1PjZjWk/edit).
- computation of served families incl. filtering
- computation of sales incl. filtering
- computation of stock overview
- computation of moved stock overview

The product list can be fetched via the GraphQL API in a straightforward way.

@aerinsol
1. I settled for the name `metrics` instead of `statistics`. Imho a statistical value is e.g. an average or a variance but not a counted number. Any objections?
1. What is the difference between the first and the second query in the document?
1. When computing the number of sales, should we add `Transaction.tokens > 0` to avoid counting the transactions that actually represent granted tokens?
1. When computing moved stock, we use `WHERE NOT location.visible` but when computing stock overview, we use `WHERE location.visible = 1`. Is that on purpose?
1. All queries return info for the current client's organisation. Should the god user be capable of fetching data for any organisation?
1. I compared the SQL queries generated by peewee with the ones from the document. They match in functionality I reckon. ~~I still have to perform a sanity comparison by running the SQL queries directly on the production database.~~ Sanity comparison was successful!

As usual let me know about any remarks :)


pylipp:
> Hey @pylipp , sorry for the late reply. Here are my answers:

Thanks!

> > 2. What is the difference between the first and the second query in the document?
> 
> We have a functionality where people are grouped into families, so each family has a family head (you should see a self-referential parent_id in the field there). So, the difference is counting all individual people vs. family groups.

Is it correct that transactions are always booked via the family head?

The new `metrics` query only provides the number of served families. I'll add the number of served people, too.
 
> > 3. When computing the number of sales, should we add `Transaction.tokens > 0` to avoid counting the transactions that actually represent granted tokens?
> 
> No, because there are a lot of transactions that don't use any granted tokens (i.e. product prices are zero or null). We should, however, exclude negative tokens, so I will update that query :)

Does it make sense to also update the queries for computing the number of served families/people too? For my understanding, negative tokens correspond to 'granted tokens'?

> > 4. When computing moved stock, we use `WHERE NOT location.visible` but when computing stock overview, we use `WHERE location.visible = 1`. Is that on purpose?
> 
> Yes. "invisible' locations are locations that are not counted as stock, i.e., the LOST "location" (should be solved when we correctly handle status via @leo-labs PR), and outgoing locations, where they're being sent on to another location, or being distributed en masse often in a non-choice way.

Ah, I suppose the computation of that query then changes anyway with the update of the box-location-state logic. Did you see my response to [this comment related to the topic](https://docs.google.com/document/d/10GFOZaWMKJCD4PAt6WSvIiBCWI09319CdLDU1PjZjWk/edit?disco=AAAASmfz4Xw) ?

> > 5. All queries return info for the current client's organisation. Should the god user be capable of fetching data for any organisation?
> 
> Yes. Also I should probably discuss with you and @vahidbazzaz we will probably want to have a way of granting a user API read-only access to all organizations without also granting godadmin access to the application.

Okay, I'll try to extend the PR to make the god user capable.

aerinsol:
> Is it correct that transactions are always booked via the family head?

Yes (for now).

> > > 3. When computing the number of sales, should we add `Transaction.tokens > 0` to avoid counting the transactions that actually represent granted tokens?
> > 
> > 
> > No, because there are a lot of transactions that don't use any granted tokens (i.e. product prices are zero or null). We should, however, exclude negative tokens, so I will update that query :)
> 
> Does it make sense to also update the queries for computing the number of served families/people too? For my understanding, negative tokens correspond to 'granted tokens'?

No, this is not correct. It most likely refers to reversed purchases. Yes, we can update those queries of served families to exclude negative tokens.

> > > 4. When computing moved stock, we use `WHERE NOT location.visible` but when computing stock overview, we use `WHERE location.visible = 1`. Is that on purpose?
> > 
> > 
> > Yes. "invisible' locations are locations that are not counted as stock, i.e., the LOST "location" (should be solved when we correctly handle status via @leo-labs PR), and outgoing locations, where they're being sent on to another location, or being distributed en masse often in a non-choice way.
> 
> Ah, I suppose the computation of that query then changes anyway with the update of the box-location-state logic. Did you see my response to [this comment related to the topic](https://docs.google.com/document/d/10GFOZaWMKJCD4PAt6WSvIiBCWI09319CdLDU1PjZjWk/edit?disco=AAAASmfz4Xw) ?

Yes, let's continue the conversation there.
 
> > > 5. All queries return info for the current client's organisation. Should the god user be capable of fetching data for any organisation?
> > 
> > 
> > Yes. Also I should probably discuss with you and @vahidbazzaz we will probably want to have a way of granting a user API read-only access to all organizations without also granting godadmin access to the application.
> 
> Okay, I'll try to extend the PR to make the god user capable.

To be clear - there are two roles that are needed, but we can put this in a separate PR. 1) Godadmin can do everything across all organizations. 2) Read-only Boxtribute internal user can read across all organizations.

We can do #1 first.

aerinsol (approved):

pylipp:
> > > > 5. All queries return info for the current client's organisation. Should the god user be capable of fetching data for any organisation?
> > > 
> > > 
> > > Yes. Also I should probably discuss with you and @vahidbazzaz we will probably want to have a way of granting a user API read-only access to all organizations without also granting godadmin access to the application.
> > 
> > 
> > Okay, I'll try to extend the PR to make the god user capable.
> 
> To be clear - there are two roles that are needed, but we can put this in a separate PR. 1) Godadmin can do everything across all organizations. 2) Read-only Boxtribute internal user can read across all organizations.
> 
> We can do #1 first.

This is implemented, I'll merge the PR then.

Thanks for the helpful clarifications :)

PR #170, opened 2022-02-14
Configure CircleCI to deploy api-staging service

pylipp:
This builds on boxwise/dropapp#461.
- only handle / endpoint for api* subdomain
- skip building and testing front-end
- test on current branch

EDIT: the `DB_SOCKET` env variable was incorrect in CircleCI but the 502 persists after another deploy.
EDIT2: The `MYSQL_HOST` and `MYSQL_PORT` env variables where not set for the flask app. Now it's live :sunglasses: 
EDIT3: Live on https://api-staging.boxtribute.org/. I added a dedicated `app-api.yaml`. It's reused in the CircleCI config when deploying the `api-demo` and `api` (production) subdomains


pylipp:
[EPIC: Expose API on New App to Our Partners](https://trello.com/c/mnuzzEcb/814-epic-expose-api-on-new-app-to-our-partners)

vahidbazzaz (approved):

back/app-api-staging.yaml
```diff
@@ -0,0 +1,7 @@
+runtime: python38
+service: api-staging
+entrypoint: gunicorn -b :$PORT
+handlers:
+- url: /
```
pylipp, line 5: @vahidbazzaz Do you have an idea how to arrange it here that `v2/graphql` is not visible? Otherwise I'd have to add a dedicated Flask app, holding the 'query' endpoints only.
Or we ditch the idea of exposing the API on a subdomain `api*.` but instead use an URL suffix `/api`. Honestly that would simplify the CircleCI config again.

vahidbazzaz: @pylipp I concur with you and I think that absolutely make sense to have a dedicate end point for api, but we can check it out and see if that fine with other use cases (FE).

pylipp: We agreed that we want to use subdomains for exposing the query API.

PR #174, opened 2022-02-26
Fix Auth0 configuration

pylipp:
- added dedicated API and application in Auth0
- introduce QUERY_API_* env variables in CircleCI (distinguish from
  those for the webapp backend)

This avoids returning an ID token on the `/token` endpoint. Instead only an access token is returned, and acc. to convention only such a token should be used to convey user authorization information (cf. https://auth0.com/blog/id-token-access-token-what-is-the-difference/)

Successfully tested with `POST /token` and api.boxtribute.org.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -200,8 +200,9 @@ jobs:
           command: |
             echo "
               AUTH0_DOMAIN=${AUTH0_DOMAIN}
-              AUTH0_CLIENT_ID=${AUTH0_CLIENT_ID}
-              AUTH0_CLIENT_SECRET=${AUTH0_CLIENT_SECRET}
+              AUTH0_AUDIENCE=${AUTH0_AUDIENCE}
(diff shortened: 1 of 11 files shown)


pylipp:
[Fix usage of access token for query API](https://trello.com/c/iD5f8lfW/846-fix-usage-of-access-token-for-query-api)

vahidbazzaz (approved):

pylipp:
@spaudanjo @leo-labs @vahidbazzaz with this PR, the naming of two variables in the `.env` file changed: TEST_AUTH0_CLIENT_ID and TEST_AUTH0_CLIENT_SECRET instead of AUTH0_CLIENT_TEST_ID and AUTH0_CLIENT_SECRET_TEST. Also AUTH0_USERNAME and AUTH0_PASSWORD will not be read from the `.env` file for testing anymore. Please update you local `.env` files accordingly, any WIP branches are probably worth rebasing :slightly_smiling_face: I've already updated the relevant env variables in CircleCI.

PR #178, opened 2022-03-05
staging-explorations: add tmp enviornment

spaudanjo:
related to https://github.com/boxwise/dropapp/pull/466

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -282,6 +282,18 @@ workflows:
           requires:
             - checkout-to-workspace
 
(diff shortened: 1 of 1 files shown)


vahidbazzaz (approved):

pylipp:
What's the benefit of using this over the staging environment?

vahidbazzaz:
> What's the benefit of using this over the staging environment?

@pylipp It seems to me that @spaudanjo would prefer to have a custom domain instead of one with the /v2 prefix. Anyhow, this is a temporary solution, and I agree with you that we can use the staging environment.

PR #187, opened 2022-03-15
853 recreate boxes view

flisowna:
This is part 1 of the implementation of the boxes view in the new frontend. Follow ups to do already identified (there might be more): 
1. adding tests
2. as soon as the feature "base selector for the current session" is implemented: change the currently hard coded baseId (1) to the selected baseId (@spaudanjo)
3. write an ADR proposal statement for using chakra-ui and react-table (@spaudanjo)
4. optimise for mobile screens


flisowna:
https://trello.com/c/BGjslF4u/853-recreate-boxes-list-view-in-new-fe

PR #192, opened 2022-03-17
853 recreate boxes view

flisowna:
This is part 1 of the implementation of the boxes view in the new frontend. Follow ups to do already identified (there might be more): 
1. adding tests
2. as soon as the feature "base selector for the current session" is implemented: change the currently hard coded baseId (1) to the selected baseId (@spaudanjo)
3. write an ADR proposal statement for using chakra-ui and react-table (@spaudanjo)
4. optimise for mobile screens

Contributes to https://trello.com/c/BGjslF4u/853-recreate-boxes-list-view-in-new-fe


vahidbazzaz (approved):
thanks @flisowna

spaudanjo:
Thanks @vahidbazzaz :) 
@flisowna will merge in a few hours as soon as she is back

react/src/types/react-table-config.d.ts
```diff
@@ -0,0 +1,120 @@
+import {
```
spaudanjo, line 1: Maybe a comment with a "Why is this needed" and a link (SO, github issue to the react-table project etc) could help here?

spaudanjo (approved):
I left one code comment - if you don't think it's needed, feel free to merge (but I think it would be helpful to add this info to the react-table types file) :)

react/src/types/react-table-config.d.ts
flisowna: I added a comment :)

PR #209, opened 2022-04-05
Base switcher

spaudanjo:
Closes https://trello.com/c/byffqWJL/865-enable-switching-between-bases-for-new-frontend

Also it adds
* a automatic base switcher: for root urls like `http://domain.xyz/`, the first base of the list of all bases where the user has access to will be selected (OR, in the future, we can implement a 'go to latest used base from previous session' - see next point)
* a global preference context (also things like the most recently used base can be stored there, if the same user comes back to the app in the future, it can be used to automatically go to this base; we can leverage the localStorage for that)

### DIFF
--- a/react/.prettierignore
+++ b/react/.prettierignore
@@ -1,0 +1,2 @@
+src/generated/
+*.d.ts
--- a/react/src/App.tsx
+++ b/react/src/App.tsx
@@ -1,36 +1,63 @@
-import React from "react";
+import React, { useContext, useEffect } from "react";
 import { Route, Routes } from "react-router-dom";
-import { useAuth0, withAuthenticationRequired } from "@auth0/auth0-react";
-import Home from "views/Home";
 import Boxes from "views/boxes/Boxes";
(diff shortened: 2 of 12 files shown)


vahidbazzaz (approved):
@spaudanjo Thanks for adding the base switcher, and for making use of context to keep track of the global bases. 😃 

I've run it locally and it worked just fine. 👍 

I have a few comments and questions as follow:

- Would you be able to add some tests to cover the functionality of the base switcher? 
- What is our current standard for naming conversion? I prefer not to have dashes "-" in the view folder, and I feel that you can use 'BaseSwitcher' instead of 'AutomaticBaseSwitcher'. 
- Could you please let me know when we place something under components instead of views? 
I don't have access to the current Figma design. Here are some comments related to UX: 
- Once I select one base, I cannot see the name of the selected base anymore, so it would be nice if we could show the selected base name.
- Changing bases changes your current page, but I think it should change only the base, not my current page?

PR #210, opened 2022-04-07
Adjust query API URL in docs depending on context

pylipp:
The URL for the respective query API is updated (both in the right and left column, and in the auth/authz section).

Please see https://api-staging.boxtribute.org/docs and https://api-demo.boxtribute.org/docs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -231,6 +231,9 @@ jobs:
                 name: Generate GraphQL documentation
                 command: |
                   cd ../docs/graphql-api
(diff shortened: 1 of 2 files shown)


pylipp:
[Update GraphQL documentation hosted on non-production API so it shows the other endpoints, preferably dynamically #small](https://trello.com/c/ZdYhBOCF/864-update-graphql-documentation-hosted-on-non-production-api-so-it-shows-the-other-endpoints-preferably-dynamically-small)

aerinsol (approved):

PR #211, opened 2022-04-10
Boxes view tests

flisowna:
We added tests to the boxes view.


vahidbazzaz (approved):
@flisowna, thanks for adding boxes and locations tests that cover current views. I have run them locally and they work fine.

PR #221, opened 2022-04-23
Refactor file structure and naming convention proposal

spaudanjo:
Some file and component refactoring which is also a suggestion for a naming/folder organisational convention: 

* following an "As local as possible, as global as needed" approach: components get only moved into a more global/outer folder if they are used on that level
* views of react-router paths going into the `views` folder
* each view can have it's own folder - which in return can have a local components folder 
* ideally only one component per file
* no `index.ts` files, besides the one root index file (makes file navigation more confusing and I don't see convincing reason to use it)
* files and folders which export a component/view are written UpperCamelCase, with the same name as actual exported component/view
* other files (like types.ts, helpers.ts etc) and folders (like `providers`, `utils` etc) are written in lowerCamelCase

@vahidbazzaz @flisowna please let me know what you think :)


spaudanjo:
For some reason CircleCI builds fail, even though locally the tests are green and the build succeeds. 
Will look into that later today after some other tasks.

vahidbazzaz:
@spaudanjo Thanks for the changes in the file structure and naming conversion. I tried to run it locally, but got the following error, which seems to be because of **forceConsistentCasingInFileNames**.:

`module "/Users/vahid/Projects/boxtribute/boxtribute/react/src/views/Boxes/components/BoxesTable"
Already included file name '/Users/vahid/Projects/boxtribute/boxtribute/react/src/views/Boxes/components/BoxesTable.tsx' differs from file name '/Users/vahid/Projects/boxtribute/boxtribute/react/src/views/boxes/components/BoxesTable.tsx' only in casing.
  The file is in the program because:
    Imported via "./components/BoxesTable" from file '/Users/vahid/Projects/boxtribute/boxtribute/react/src/views/Boxes/Boxes.tsx'
    Matched by include pattern 'src' in '/Users/vahid/Projects/boxtribute/boxtribute/react/tsconfig.json'ts(1261)
tsconfig.json(27, 5): File is matched by include pattern specified here.
`

spaudanjo:
Yes, this after I reconfigured git locally with `git config core.ignorecase false` but apparently now both upper and lower case folders are checked in :D 
I quickly try to fix this.

vahidbazzaz (approved):
Thank @spaudanjo its no working fine at my end. Also your proposal for structure and naming seems very reasonable 👍

PR #234, opened 2022-05-06
Boxes view pagination (on FE side so far)

flisowna:


spaudanjo:
Looks good and works for me :) 
Important to point out that it's atm indeed not using server sided pagination.

spaudanjo (approved):

PR #246, opened 2022-05-24
Handle authz for god users

pylipp:
Regardless of the value of the `organisation_id` claim in the JWT, set the internal `CurrentUser.organisation_id` to None for god users, and update involved filtering logic accordingly.


pylipp:
[Configure GraphQL background configurations to account for the fact that Boxtribute Gods don't belong to an organization](https://trello.com/c/bDJ1MfXS/858-configure-graphql-background-configurations-to-account-for-the-fact-that-boxtribute-gods-dont-belong-to-an-organization)

spaudanjo (approved):

vahidbazzaz (approved):

PR #292, opened 2022-07-06
Improve performance of boxes-for-base query

pylipp:
Loading the Boxes view in the FE was reportedly slow. I first looked at the GraphQL query issued by the FE, and then dug into the SQL queries run by peewee in the BE. I could find a redundant `select * from stock` query as a side-effect from peewee objects (still trying to understand the background).

@spaudanjo @flisowna Can you check the performance in the FE on your end?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,5 +12,6 @@ back/test-results
 yarn-error.log
 
 ./public/
+docs/graphql-api/public
 *.zip
 .VSCodeCounter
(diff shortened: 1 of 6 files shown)


spaudanjo (approved):

pylipp:
[Improve performance of boxes-for-base query in boxtribute 2.0](https://trello.com/c/1kJ9lewB/919-improve-performance-of-boxes-for-base-query-in-boxtribute-20)

aerinsol:
Happy to help with optimization down the line @pylipp if the raw sql gets printed out

PR #294, opened 2022-07-10
Mobile Distro Prototyping: Packing List (WIP)

spaudanjo:
* lot of POC work 
* work in FE and BE
* also tried out YUP for data validation
* Works: adding items to packing list and show packing list
* Removing packing list items doesn't work yet
* Not other stages then Planning supported yet at all
* partially broken UI


flisowna (approved):
ok

PR #298, opened 2022-07-11
Mobile distro events: packing view 

flisowna:
## Includes:
- Packing view for mobile distribution

## To include/adapt:
- integration of graphql
- bulk scanner - also views and functionality
- deskop view adjustment
- might need some refactoring once it gets integrated into the actual app flow


pylipp:
@flisowna I think the target branch should be mobile-distro-master :)

PR #299, opened 2022-07-13
Mobile distribution events: add "move to next stage" and "remove item from packing list"

spaudanjo:


back/boxtribute_server/graph_ql/resolvers.py
```diff
@@ -466,6 +474,48 @@ def resolve_location_default_box_state(location_obj, _):
...
+@convert_kwargs_to_snake_case
+def resolve_remove_packing_list_entry_from_distribution_event(
+    *_, packing_list_entry_id
+):
+    # TODO: add authorization here
+    # authorize(permission="distribution_event:write")
+
+    packing_list_entry = PackingListEntry.get(packing_list_entry_id)
+    if packing_list_entry is None:
+        raise GraphQLError("Packing list entry not found")
```
pylipp, line 487: Did you test that the error is raised when a non-existing ID is given?

I'd rather expect that for this case, in line 485 peewee already raises an exception. peewee exceptions for non-existent resources are handled automatically (in `exceptions.py` the `format_database_errors` function is used to intercept any exception bubbling up from the app code to the GraphQL layer). This means you don't have to do additional error handling in the resolvers. 

Error information will be contained in the `errors` field of the GraphQL response, holding fields `code` and `description`. For their content also see `exceptions.py`. The error codes are similar to what Apollo-Server proposes.

back/boxtribute_server/graph_ql/resolvers.py
```diff
@@ -466,6 +474,48 @@ def resolve_location_default_box_state(location_obj, _):
...
+    mobile_distribution_event = (
+        DistributionEvent.select()
+        .where(DistributionEvent.id == packing_list_entry.distribution_event_id)
+        .get()
```
pylipp, line 491: You can use the helper `get_by_id()`

PR #302, opened 2022-07-16
Integrate sentry into FE

spaudanjo:


pylipp (changes requested):
I feel like the `.DS_Store` file should not be commited but git-ignored. Otherwise LGTM :)

.gitignore
```diff
@@ -16,3 +16,4 @@ docs/graphql-api/public
 *.zip
 .VSCodeCounter
 .DS_Store
+.DS_Store
```
pylipp, line 19: now we have it twice :laughing:

spaudanjo: let's make it three then! 
weird so, let me check...

react/src/index.tsx
```diff
@@ -38,14 +41,35 @@ const theme = extendTheme({
 
 const AuthenticationProtectedApp = withAuthenticationRequired(App);
 
+const SentryProfiledApp = Sentry.withProfiler(AuthenticationProtectedApp);
+
+const sentryDsn = process.env.REACT_APP_SENTRY_DSN;
```
pylipp, line 46: Do we have to distinguish between two projects (FE and BE) in Sentry for the new app? Otherwise you can use `SENTRY_DSN` which is already set as env variable in CircleCI (the name is dictated by the Sentry SDK). See also #303 

Also for environment I introduced the env variable `SENTRY_ENVIRONMENT` with possible values staging, demo, production.

spaudanjo: I think it makes sense to separate the two systems also in different projects. 
Will integrate the `SENTRY_ENVIRONMENT` env variable! :)

pylipp: @vahidbazzaz @HaGuesto I don't have required permissions, can you rename the existing project `boxwise-production` into `boxtribute2-backend` and create a new project `boxtribute2-frontend` in Sentry? Or you grant me permissions to do it myself :wink:

pylipp (approved):

pylipp:
Now I'm wondering why we have `yarn.lock` and `react/yarn.lock`. Is it on purpose?

spaudanjo:
> Now I'm wondering why we have `yarn.lock` and `react/yarn.lock`. Is it on purpose?

Damnit... this PR is a black whole for silly issues...
Will remove it in the next PR (happens if you accidentally run `yarn add XYZ` in the root folder instead of the react folder).

PR #339, opened 2022-08-08
FE: deactivate failing Boxes List view tests temporary

spaudanjo:
Was introduced with 19d0f49dffabaf3b78a6e640abfd4909d5cbddf7

Focus is at the moment on prototyping and having a user-testable version ready. Automated test will to be updated/extended once features are confirmed as stable by user testing.


pylipp (approved):
Thanks for the investigation @spaudanjo :)

PR #341, opened 2022-08-10
Mobile Distro Prototyping: more BE extensions

spaudanjo:
Contributes to https://trello.com/c/iSwBNksN/932-extend-new-be-for-mobile-distribution-prototype


back/boxtribute_server/graph_ql/resolvers.py
```diff
@@ -475,6 +478,35 @@ def resolve_add_packing_list_entry_to_distribution_event(*_, creation_input):
...
+    authorize(permission="packing_list_entry:write")
+    return update_packing_list_entry(g.user.id, packing_list_entry_id, number_of_items)
+
+
+@mutation.field("removeAllPackingListEntriesFromDistributionEventForProduct")
+@convert_kwargs_to_snake_case
+def resolve_remove_all_packing_list_entries_from_distribution_event_for_product(
+    *_, distribution_event_id, product_id
+):
+    authorize(permission="packing_list_entry:write")
```
pylipp, line 493: Does this action require `distro_event:write` permission as well?

back/boxtribute_server/mobile_distribution/crud.py
```diff
@@ -1,5 +1,8 @@
 from boxtribute_server.models.definitions.distribution_event import DistributionEvent
```
pylipp, line 1: Just the usual nitpick..wouldn't mind relative imports like `from ..models.definitions...` here :wink:

back/boxtribute_server/mobile_distribution/crud.py
```diff
@@ -92,6 +95,88 @@ def change_distribution_event_state(distribution_event_id, distribution_event_st
     return distribution_event
 
 
+def set_products_for_packing_list(
+    user_id, distribution_event_id, product_ids_to_add, product_ids_to_remove
+):
+    """
+    Set the products for a packing list.
```
pylipp, line 102: The docstring is identical to the function name. I suggest we omit it, or explain the implementation below in a bit more detail (moving comments like `remove/add new products` into the docstring)

back/boxtribute_server/mobile_distribution/crud.py
```diff
@@ -92,6 +95,88 @@ def change_distribution_event_state(distribution_event_id, distribution_event_st
...
+            for size in sizes:
+
+                PackingListEntry.get_or_create(
+                    distribution_event=distribution_event_id,
+                    product=product_id,
+                    size=size.id,
+                    defaults={
+                        "number_of_items": 0,
+                        "created_by": user_id,
+                        "created_on": utcnow(),
```
pylipp, line 137: Can you move the `utcnow` call outside of the loops into line 119? That'd be more efficient and precise

back/boxtribute_server/mobile_distribution/crud.py
```diff
@@ -92,6 +95,88 @@ def change_distribution_event_state(distribution_event_id, distribution_event_st
...
+
+
+def remove_all_packing_list_entries_from_distribution_event_for_product(
+    user_id, distribution_event_id, product_id
+):
+    """
+    Remove all packing list entries from a distribution event for a product.
+    """
+    with db.database.atomic():
+        packing_list_entries = PackingListEntry.select().where(
```
pylipp, line 152: FYI I learnt more about the different ways of how `where` works here: http://docs.peewee-orm.com/en/latest/peewee/api.html#Query.where

back/boxtribute_server/mobile_distribution/crud.py
```diff
@@ -92,6 +95,88 @@ def change_distribution_event_state(distribution_event_id, distribution_event_st
...
+            PackingListEntry.distribution_event == distribution_event_id,
+            PackingListEntry.product == product_id,
+        )
+
+        for packing_list_entry in packing_list_entries:
+            packing_list_entry.delete_instance()
```
pylipp, line 158: You could instead do a `PackingListEntry.delete().where(...)` in line 152 (don't forget `.execute()` at the end). Probably more efficient to do an SQL bulk delete.

back/boxtribute_server/mobile_distribution/crud.py
```diff
@@ -92,6 +95,88 @@ def change_distribution_event_state(distribution_event_id, distribution_event_st
...
+
+        return True
+
+    return False
+
+
+def update_packing_list_entry(user_id, packing_list_entry_id, number_of_items):
```
pylipp, line 165: The GraphQL schema allows `numberOfItems` to be undefined, so you either need to add a default like None here (and handle it below), or require numberOfItems in the schema

back/test/data/size.py
```diff
@@ -9,7 +9,7 @@ def default_data():
 
 
 def another_data():
-    return {"id": 2, "label": "medium"}
+    return {"id": 2, "label": "medium", "size_range": size_range_data()["id"]}
```
pylipp, line 12: This addition broke the `test_product` test because the query there requests `sizeRange { sizes { id } }`. Hence there the expectation needs to be adjusted (add `another_size`)

back/boxtribute_server/mobile_distribution/crud.py
```diff
@@ -92,6 +95,88 @@ def change_distribution_event_state(distribution_event_id, distribution_event_st
     return distribution_event
 
 
+def set_products_for_packing_list(
+    user_id, distribution_event_id, product_ids_to_add, product_ids_to_remove
```
pylipp, line 99: Personally I like forcing the calling code to use keyword-argument form (`set_products(user_id=g.user.id, distribution_event_id=distribution_event_id`). Even though there is increased verbosity, the risk of misordering arguments is avoided.
Forcing keyword-arguments works by adding a `*` 'indicator' argument in the signature

back/boxtribute_server/mobile_distribution/crud.py
```diff
@@ -92,6 +95,88 @@ def change_distribution_event_state(distribution_event_id, distribution_event_st
...
+                    size=size.id,
+                    defaults={
+                        "number_of_items": 0,
+                        "created_by": user_id,
+                        "created_on": utcnow(),
+                        "state": PackingListEntryState.NotStarted,
+                    },
+                )
+
+        return True
```
pylipp, line 142: As we said, could be beneficial to return the distribution event here. Also since the function would never return false

pylipp (changes requested):

back/boxtribute_server/graph_ql/resolvers.py
spaudanjo: No, I think packing_list_entry:write is enough

back/boxtribute_server/mobile_distribution/crud.py
spaudanjo: Good point. I would even think, once we do a proper code cleanup of the new features, to make functions much shorter and extract things like "Remove products" and "Add products" into seperate functions with proper names (this would also remove the need for more descriptive comments).

pylipp (approved):

PR #347, opened 2022-08-13
Improve complex GraphQL queries

pylipp:
I had another look at the performance of the BoxesForBase query.
Using a profiler I found three bottlenecks:

- Avoid unnecessary SELECT for product gender ID (this is a peewee caveat, fixed)
- Remove redundant query for box place (was fixed in #292 but re-introduced when renaming happened, fixed)
- resolving `boxes { elements { product { } } }` results in a lot of separated SELECT queries for products (same for size, presumably). I hacked a basic loader (storing a single call of `Product.select()` in a dictionary) which brought down runtime of the product resolver from 1.4s to 0.14s

As a conclusion from the third bottleneck, and because the other slow queries have similar complexity, I think we defs have to look into data loaders.

FYI @aerinsol

### DIFF
--- a/back/boxtribute_server/graph_ql/resolvers.py
+++ b/back/boxtribute_server/graph_ql/resolvers.py
@@ -893,7 +893,7 @@ def resolve_resource_base(obj, _):
 @product.field("gender")
 def resolve_product_gender(product_obj, _):
     # Instead of a ProductGender instance return an integer for EnumType conversion
-    return product_obj.gender.id
+    return product_obj.gender_id
 
 
 @product_category.field("hasGender")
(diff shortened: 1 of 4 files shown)


pylipp:
[Improve performance of boxes-for-base query in boxtribute 2.0](https://trello.com/c/1kJ9lewB/919-improve-performance-of-boxes-for-base-query-in-boxtribute-20)

aerinsol:
@pylipp feel free to look into an example dataloader when the sponsorship comes online. What I've seen seems to show it conflicting with Flask and our current middleware, and solvable by writing resolvers as fragments. But we can do a deep dive on this together with @spaudanjo and later @vahidbazzaz

spaudanjo (dismissed):

spaudanjo:
Amazing @pylipp - performance improved a lot! :D

spaudanjo (approved):

vahidbazzaz (commented):
@pylipp nice job!

PR #350, opened 2022-08-16
Manage boxes

flisowna:
## Includes:
- state changes to lost and scrap on the box view level
- layout changes of the box view
- edit view updated 

## To do:
- other state changes on the box view by changing location more understandable for the user


spaudanjo (approved):

pylipp:
FYI updating a box location alone also changes the box state (to the location's default box state. This behavior is supposed to deal with the legacy of locations being (ab)used as states). If both location and state are given for update, the value indicated by the state field will be stored as new box state.

PR #356, opened 2022-08-16
GraphQL/PeeWee: rename 'items' to 'numberOfItems' and 'number_of_items'

spaudanjo:
Closes https://trello.com/c/sano156L/938-new-app-use-consistently-numberofitems-numberofitems-in-python-instead-of-items


spaudanjo:
@pylipp @aerinsol I'm actually wondering whether `quantity` is not even a better name (plus, one double-naming candidate less - we can call it both in the GraphQL and the Python internal worlds the same since it has just one word - instead of numberOfItems which would also have to be written - for Python - number_of_items). 
If you both agree, I can change it quickly. 

Also be aware that this doesn't change the database column names of course - so no changes for Dropapp. 
It would require query adaptions though (very easy/quick ones) for any partner using the API though.

aerinsol:
@spaudanjo I am ok with changing it, but I would like _item_ to be part of the naming, as _quantity_ is vague (boxes? Something else?). Not worried about leaving the db named the same, that's one of the benefits of using an ORM - no costly schema migrations.

I will leave @pylipp to comment on the other points.

back/boxtribute_server/graph_ql/resolvers.py
```diff
@@ -146,11 +146,6 @@ def resolve_packing_list_entry_matching_packed_items_collections(obj, *_):
     )
     unboxed_items_colletions = UnboxedItemsCollection.select(
         UnboxedItemsCollection,
```
pylipp, line 148: For `select *` you can omit the part in the parentheses.

back/boxtribute_server/graph_ql/types.graphql
```diff
@@ -32,7 +32,7 @@ type Box implements ItemsCollection {
   place: BoxPlace
   distributionEvent: DistributionEvent
   " The number of items the box contains. "
```
pylipp, line 34: you can remove this comment now, the field name is self-explanatory

react/src/views/Boxes/BoxesView.test.tsx
```diff
@@ -34,7 +34,7 @@ describe("Boxes view", () => {
                         gender: "Women",
                         name: "Long Dress",
                       },
-                      items: 64,
+                      itnumberOfItemsems: 64,
```
pylipp, line 37: This doesnt look right :laughing:

react/src/views/Boxes/BoxesView.test.tsx
```diff
@@ -126,7 +126,7 @@ describe("Boxes view", () => {
                         gender: "Boy",
                         name: "Top Boys (18-24 months)",
                       },
-                      items: 95,
+                      itnumberOfItemsems: 95,
```
pylipp, line 129: same

pylipp (changes requested):
Search&replace FTW, minor things to clean up :smiley:

pylipp:
[New App: use consistently numberOfItems (number_of_items in Python) instead of items](https://trello.com/c/sano156L/938-new-app-use-consistently-numberofitems-numberofitems-in-python-instead-of-items)

react/src/views/Boxes/BoxesView.test.tsx
spaudanjo: You passed the test! #weeklyPrReviewQualityAssurance

spaudanjo:
addressed you comments @pylipp 🌞

pylipp (approved):

PR #357, opened 2022-08-18
Handle zero DATE value in DoB field of beneficiary

pylipp:
Work-around legacy issue that DoB fields can have the value
"0000-00-00". Previously the beneficiary-age resolver would throw an
error (the zero date, when returned from the database, is not a valid
Python date, so peewee returns the raw string. The age resolver however
expects a Python date object).
This can be removed once the clean-up migrations have been applied.

FYI @aerinsol @HaGuesto


aerinsol (approved):
Nice work @pylipp

vahidbazzaz (approved):

PR #358, opened 2022-08-19
Update permission enforcement

pylipp:
This PR brings some updates regarding authz before the production-merge for Boxtribute 2.0

Please see the attached trello card and the commit messages for context.

FYI @aerinsol

### DIFF
--- a/back/boxtribute_server/graph_ql/resolvers.py
+++ b/back/boxtribute_server/graph_ql/resolvers.py
@@ -539,13 +539,6 @@ def resolve_create_distribution_spot(*_, creation_input):
     return create_distribution_spot(user_id=g.user.id, **creation_input)
 
 
-@mutation.field("createBox")
-@convert_kwargs_to_snake_case
-def resolve_create_box(*_, creation_input):
-    authorize(permission="stock:write")
-    return create_box(user_id=g.user.id, **creation_input)
-
-
 @mutation.field("moveBoxToDistributionEvent")
 @convert_kwargs_to_snake_case
 def resolve_move_box_to_distribution_event(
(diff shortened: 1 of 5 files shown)


pylipp:
[Update permission enforcement in Boxtribute 2.0](https://trello.com/c/OEfIxvmZ/939-update-permission-enforcement-in-boxtribute-20)

vahidbazzaz (approved):

aerinsol:
Sorry @pylipp  also tagged you on this in a different card in Trello but I read the commit messages and I'm still not sure what this does vs. https://github.com/boxwise/boxtribute/pull/388? Is it  base and org level permission enforcement for beneficiaries & users vs. boxes?

pylipp:
The PRs fix three different bugs, basically. @aerinsol 

This PR brings two things mostly:
- improve the `users` query such that non-god users will receive an empty list (in the the future a user authorized for base 1 should only be able to fetch data of users of base 1. Right now the new back-end is not able to infer the base that a user belongs to from our database alone, we need to reach out for Auth0 for that)
- bring a small permission enforcement to make it impossible to create/update a box with a location or product that belongs to a base the user is not authorized for (this can be seen as base-level permission enforcement)

#388 enforces base-level permission when querying for beneficiaries from an org level.

aerinsol:
Thanks, that's much clearer, appreciate it @pylipp

aerinsol:
[2.0 Fix for Base-level authorization: users should be granted access granted on a whitelist rather than blacklist basis](https://trello.com/c/67MMmeWv/948-20-fix-for-base-level-authorization-users-should-be-granted-access-granted-on-a-whitelist-rather-than-blacklist-basis)

aerinsol:
[2.0 Fix for Base-level authorization: by default users should be granted access on a whitelist basis only](https://trello.com/c/67MMmeWv/948-20-fix-for-base-level-authorization-by-default-users-should-be-granted-access-on-a-whitelist-basis-only)

PR #360, opened 2022-08-20
Introduce dataloaders

pylipp:
After some research I managed to integrate dataloaders provided by the [aiodataloader](https://github.com/syrusakbary/aiodataloader) package.

Under the hood this results in batching of DB queries, and hence a significant speed-up of request processing time.

Using a local testing setup I obtained a decrease of processing time from 4.5s down to 0.7s for the GraphQL request behind the [BoxesView](https://github.com/boxwise/boxtribute/blob/master/react/src/views/Boxes/BoxesView.tsx#L8).

I verified that the database queries are batched using peewee's logger, and also verified data integrity by comparing the result of the query between master and this branch.

As discussed, it is important to deploy this on the staging server to test how the BE behaves with concurrent requests.

### DIFF
--- a/back/boxtribute_server/graph_ql/resolvers.py
+++ b/back/boxtribute_server/graph_ql/resolvers.py
@@ -891,6 +891,16 @@ def resolve_resource_base(obj, _):
     return obj.base
 
 
+@product.field("category")
+def resolve_product_product_category(product_obj, info):
+    return info.context["product_category_loader"].load(product_obj.category_id)
+
+
+@product.field("sizeRange")
+def resolve_product_size_range(product_obj, info):
+    return info.context["size_range_loader"].load(product_obj.size_range_id)
+
+
 @product.field("gender")
 def resolve_product_gender(product_obj, _):
     # Instead of a ProductGender instance return an integer for EnumType conversion
--- a/back/boxtribute_server/graph_ql/resolvers.py
+++ b/back/boxtribute_server/graph_ql/resolvers.py
@@ -219,24 +219,32 @@ def resolve_qr_code(obj, _, qr_code=None):
 
 
 @box.field("tags")
-def resolve_box_tags(box_obj, _):
-    return (
-        Tag.select()
-        .join(TagsRelation)
-        .where(
-            (TagsRelation.object_id == box_obj.id)
-            & (TagsRelation.object_type == TaggableObjectType.Box)
-        )
-    )
+def resolve_box_tags(box_obj, info):
+    authorize(permission="tag:read")
+    return info.context["tags_for_box_loader"].load(box_obj.id)
 
 
 @query.field("product")
-@box.field("product")
-@unboxed_items_collection.field("product")
-def resolve_product(obj, _, id=None):
-    product = obj.product if id is None else Product.get_by_id(id)
+def resolve_product(*_, id):
+    product = Product.get_by_id(id)
     authorize(permission="product:read", base_id=product.base_id)
     return product
+
+
+@box.field("product")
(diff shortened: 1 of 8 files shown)


pylipp:
[Production-level optimizations for Python app](https://trello.com/c/CaipG15Q/786-production-level-optimizations-for-python-app)

back/boxtribute_server/graph_ql/resolvers.py
```diff
@@ -219,26 +219,33 @@ def resolve_qr_code(obj, _, qr_code=None):
...
-            (TagsRelation.object_id == box_obj.id)
-            & (TagsRelation.object_type == TaggableObjectType.Box)
-        )
-    )
+def resolve_box_tags(box_obj, info):
+    return info.context["tags_for_box_loader"].load(box_obj.id)
 
 
 @query.field("product")
+def resolve_product(obj, _, id):
```
spaudanjo, line 227: nit: shall we omit the obj param since it's not used anymore?

back/boxtribute_server/graph_ql/resolvers.py
```diff
@@ -219,26 +219,33 @@ def resolve_qr_code(obj, _, qr_code=None):
...
+    product = Product.get_by_id(id)
+    authorize(permission="product:read", base_id=product.base_id)
+    return product
+
+
 @box.field("product")
 @unboxed_items_collection.field("product")
-def resolve_product(obj, _, id=None):
-    product = obj.product if id is None else Product.get_by_id(id)
```
spaudanjo, line 237: why is this check not done / needed anymore?

back/boxtribute_server/graph_ql/resolvers.py
```diff
@@ -219,26 +219,33 @@ def resolve_qr_code(obj, _, qr_code=None):
...
-            & (TagsRelation.object_type == TaggableObjectType.Box)
-        )
-    )
+def resolve_box_tags(box_obj, info):
+    return info.context["tags_for_box_loader"].load(box_obj.id)
 
 
 @query.field("product")
+def resolve_product(obj, _, id):
+    product = Product.get_by_id(id)
```
spaudanjo, line 228: why aren't data loaders used here as well? 
Couldn't be there a query which at first queries the product e.g. as a subfield of a box related query, but later on it's also called on top-level (root query level)?

back/boxtribute_server/graph_ql/resolvers.py
pylipp: we shall :innocent:

back/boxtribute_server/graph_ql/resolvers.py
pylipp: I split the resolvers for product-by-id and product-for-box. Previously they were combined but only for the first case an id is passed as argument.

back/boxtribute_server/graph_ql/resolvers.py
pylipp: Your're right, dataloaders can be used in many more places (and should be). My focus rn was to improve the resolvers for the slow queries.

Honestly also here I was having trouble to load a product and then pass it to the subsequent authorize call, probably have to await it :grimacing:

back/scripts/load-test.js
```diff
@@ -45,11 +50,11 @@ export const options = {
       gracefulStop: '5s',
 
       // executor-specific configuration
-      vus: 10,
-      iterations: 100,
+      vus: 1,
+      iterations: 1,
```
spaudanjo, line 54: what is that? :)

spaudanjo (approved):

PR #374, opened 2022-08-28
Use JWT custom base_ids claim for admin users

pylipp:


spaudanjo (approved):
General qq: Is my understanding correct that for god users, there are no base_ids/permissions set all in the user object? So the default/initial state (no values) is the most powerful one? 
Would it maybe make more sense to explicitly set all bases/permissions for god users? Not totally sure whether this is really the case, but it feels more sound/safer to use an explicit allow-list approach (where during the auth logic flow permissions/bases are added and if something is missed, the default state is still the one with no privileges).

aerinsol:
I actually agree with this approach. The "no values" seemed vaguely uncomfortable for me but I hadn't thought about it in depth. I think this makes sense. Thoughts @HaGuesto @pylipp @vahidbazzaz?

pylipp:
The default user has `is_god = False` set, and empty permissions/base_ids. Hence they're not authorized for anything.

The issue above arose from explicitly having a default None value for base_ids result in accessing all bases. I'll check the code base again tmrw for critical authz enforcement.

pylipp:
[fix authorize() function](https://trello.com/c/67MMmeWv/948-fix-authorize-function)

pylipp:
[fix authorize() function](https://trello.com/c/67MMmeWv/948-fix-authorize-function)

PR #378, opened 2022-08-31
Make authorize() function even more robust

pylipp:
As a follow-up to #374.

- assign authorized=False and empty base IDs to User object by default
- handle edge case: resource-permission given but no bases assigned

### DIFF
--- a/back/boxtribute_server/auth.py
+++ b/back/boxtribute_server/auth.py
@@ -2,6 +2,7 @@
 import json
 import os
 import urllib
+from collections import defaultdict
 from functools import wraps
 
 from flask import g, request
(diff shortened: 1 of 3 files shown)


back/boxtribute_server/auth.py
```diff
@@ -176,9 +176,6 @@ def from_jwt(cls, payload):
             base_ids=base_ids,
```
HaGuesto, line 176: please update for loop in row 157

HaGuesto: @pylipp

pylipp: Done :heavy_check_mark:

HaGuesto (approved):
the dict stuff looks a bit hacky, but besides that well done. and nice to see someone who writes tests.

pylipp:
[BUG: API Authz Guards Don't Work, Create/Enforce Functional Tests #flask #python](https://trello.com/c/kn9fTa3R/944-bug-api-authz-guards-dont-work-create-enforce-functional-tests-flask-python)

pylipp:
[fix authorize() function](https://trello.com/c/67MMmeWv/948-fix-authorize-function)

HaGuesto:
[fix authorize() function](https://trello.com/c/67MMmeWv/948-fix-authorize-function)

PR #382, opened 2022-09-05
GraphQL mutations for (un)assigning tags

pylipp:
This PR brings the implementation for the `assignTag` and `unassignTag` mutations.

The necessary resource-based permissions are enforced (`tag_relation:assign`; asserting the tag's base is part of the user's bases).

For the business logic check the new functions in `crud.py`.

All implemented test cases are listed [here](https://docs.google.com/spreadsheets/d/1sDhSsaVwNxAhGn1VYACDPVepZpvlwTkgy9nmeTaRy2Q/edit?pli=1#gid=1709672082) (4th sheet, section 4).
I've implemented most, except

- the ones for input validation (e.g. 4.2.25. I think it's the FE responsibility to prevent that the same tag is applied twice to a resource),
- and the "complicated" ones for authorization (e.g. 4.2.21 making sure the tag is not assigned to a box in a base that the user does not have access to. Probably makes sense to implement it but I didn't want this PR to blow up too much =) )

For details, please have a look at the commits, I tried to keep them small and comprehensive.

FYI @spaudanjo @flisowna


pylipp:
[Update GraphQL API for new Tags structure](https://trello.com/c/SNpRIZWQ/888-update-graphql-api-for-new-tags-structure)

back/boxtribute_server/models/crud.py
```diff
@@ -191,6 +191,56 @@ def delete_tag(*, user_id, id):
...
 
+def assign_tag(*, user_id, id, resource_id, resource_type):
+    """Create TagsRelation entry as cross reference of the tag given by ID, and the
+    given resource (a box or a beneficiary). Insert timestamp for modification in
+    resource model.
+    Validate that tag type and resource type are compatible.
+    Return the resource.
+    """
+    tag = Tag.get_by_id(id)
+    if (
```
HaGuesto, line 202: @pylipp can you note down for our next meeting that we should talk about input validation in the backend. 
cause you are checking here if a beneficiary type is applied to a box or the other way around, but what if there was no type passed at all or ....

HaGuesto (approved):

back/boxtribute_server/models/crud.py
pylipp: Sure! The test suite is never complete :wink: 
Note that the GraphQL API requires a non-null resource type to be passed for `TagOperationInput`.

PR #388, opened 2022-09-10
Enforce base-specific permissions when resolving base beneficiaries

pylipp:
This PR fixes a leak that previously allowed fetching beneficiary data in an unrestricted way.

It's possible to fetch beneficiary data in different ways (comments explain authz implemented on master):

1. `query { beneficiary(id: 1) { } }`: base-specific permission enforced in `resolve_beneficiary`
1. `query { beneficiaries { } }`: base-specific filter applied in `resolve_beneficiaries`
1. `query { base(id: 1) { beneficiaries { } } }`: base-specific permission enforced in `resolve_base`
1. `query { bases { beneficiaries { } } }`: base-specific filter applied in `resolve_bases`
1. `query { organisation(id: 1) { bases { beneficiaries { } } } }`: Any user can access any first-level information of any other organisation (ID, name, and bases). Since `resolve_base_beneficiaries` does not enforce base-specific permissions (I thought any parent resolver would have done so already), beneficiaries of all bases can be listed.
1. `query { organisations { bases { beneficiaries { } } } }`: same as above

The issue described in the final two points is fixed.

I extended the tests for the desired behavior. For an organisation with ID 1 and two bases, and a user who's member of only one of those bases, the request `query { organisation(id: 1) { bases { id beneficiaries { } } } }` now returns a partial response (the data fields for base IDs are available. Base beneficiaries however only contain data for bases that the user is authorized for. Otherwise they're null, and the response errors field indicates forbidden errors)

Note that via the `organisation` root query it's still possible to access any organisation's location and product information (this is not sensitive information and also inevitable for the box-transfer feature).

### DIFF
--- a/back/boxtribute_server/graph_ql/types.graphql
+++ b/back/boxtribute_server/graph_ql/types.graphql
@@ -218,7 +218,7 @@ type Base {
   name: String!
   organisation: Organisation!
   " List of all [`Beneficiaries`]({{Types.Beneficiary}}) registered in this base "
-  beneficiaries(paginationInput: PaginationInput, filterInput: FilterBeneficiaryInput): BeneficiaryPage!
+  beneficiaries(paginationInput: PaginationInput, filterInput: FilterBeneficiaryInput): BeneficiaryPage
   currencyName: String
   " List of all [`ClassicLocations`]({{Types.ClassicLocation}}) present in this base "
   locations: [ClassicLocation!]!
--- a/back/boxtribute_server/graph_ql/resolvers.py
+++ b/back/boxtribute_server/graph_ql/resolvers.py
@@ -1032,7 +1032,7 @@ def resolve_distribution_event(*_, id):
 @base.field("beneficiaries")
 @convert_kwargs_to_snake_case
 def resolve_base_beneficiaries(base_obj, _, pagination_input=None, filter_input=None):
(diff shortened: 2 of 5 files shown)


pylipp:
[Enforce base-specific permission for base beneficiaries](https://trello.com/c/6BkBC56x/951-enforce-base-specific-permission-for-base-beneficiaries)

pylipp:
[2.0 BUG: API Authz Guards Don't Work, Create/Enforce Functional Tests #flask #python](https://trello.com/c/kn9fTa3R/944-20-bug-api-authz-guards-dont-work-create-enforce-functional-tests-flask-python)

back/boxtribute_server/graph_ql/types.graphql
```diff
@@ -218,7 +218,7 @@ type Base {
   name: String!
   organisation: Organisation!
   " List of all [`Beneficiaries`]({{Types.Beneficiary}}) registered in this base "
-  beneficiaries(paginationInput: PaginationInput, filterInput: FilterBeneficiaryInput): BeneficiaryPage!
+  beneficiaries(paginationInput: PaginationInput, filterInput: FilterBeneficiaryInput): BeneficiaryPage
```
HaGuesto, line 221: @pylipp I just have a quick question here.

HaGuesto (approved):

PR #409, opened 2022-09-26
Add entries in history table when running box mutation

pylipp:
This PR adds two utility decorators that shall be applied on the data-operational level to any function that creates or updated a resource.

Please see the function docstrings for information. The implemented functionality is identical to the dropapp functionality (in the formhandler module). See also [this section](https://github.com/boxwise/dropapp/blob/master/include/stock_edit.php#L30)

The decorators are applied to the `create_box` and `update_box` functions since they are the only relevant ones for the upcoming deploy. Two unit tests had to be converted into endpoint tests because the new decorators require access to an actual Flask app context (for fetching client information).

For local testing try the following:

1. Start docker-compose services `docker-compose up webapp db`
1. Open GraphQL playground on `localhost:5005`
1. Fetch JWT and insert it
1. Create a box

```
mutation { createBox (creationInput: {
  locationId: 1 
  sizeId: 1 
  productId: 1 
}) { labelIdentifier } }
```

1. Connect to the local `dropapp_dev` database and display the `history` table
1. You should see a `Record created` entry as most recent one.

1. Update the box

```
mutation { updateBox(updateInput: {labelIdentifier: "XXXX" comment: "hallo"}){comment}}
```

1. You should see a `comments changed from "" to "hallo".` entry as most recent one.
1. Play around with updating different box fields (e.g. `productId: 2`).

### DIFF
--- a/back/boxtribute_server/models/utils.py
+++ b/back/boxtribute_server/models/utils.py
@@ -42,24 +42,26 @@ def save_creation_to_history(f):
     return inner
 
 
-def save_update_to_history(*, model, id_field_name="id", field_names):
-    """Utility for writing information about modifying a resource to the history table,
-    intended to decorate a function that modifies a database resource (e.g. a box).
+def save_update_to_history(*, id_field_name="id", fields):
+    """Utility for writing information about updating a resource to the history table,
+    intended to decorate a function that updates a database resource (e.g. a box).
 
-    The type of the modified resource is indicated by `model`, the relevant field names
-    by `field_names`. `id_field_name` refers to the field name used to identify the
-    model instance that is being modified. In the signature of the decorated function an
-    argument with identical name must exist. The decorated function must return the
-    modified resource.
+    The relevant fields are indicated by the `fields` list. The type of the updated
+    resource (the model) is derived from the first field (hence all fields must stem
+    from the same model).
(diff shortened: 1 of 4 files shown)


back/boxtribute_server/models/utils.py
```diff
@@ -1,9 +1,103 @@
...
+                entry.ip = request.remote_addr
+                entry.change_date = now
+
+                if issubclass(field.__class__, (IntegerField, ForeignKeyField)):
+                    entry.from_int = old_value
+                    entry.to_int = new_value
+                    entry.changes = field.column_name
+                else:
+                    entry.changes = f"""{field.column_name} changed from "{old_value}" \
+to "{new_value}"."""
```
HaGuesto, line 93: ends with semi colon

HaGuesto (approved):
@aerinsol I found out that we did not add history entries for assigning tags in dropapp.
Hence, adding and removing tags does not appear in the Box History atm.
Should we do that?

If so, we also need to add this here.

Also @pylipp we need to add tags input in createBox and updateBox. Please write a ticket for this.

pylipp:
> Also @pylipp we need to add tags input in createBox and updateBox. Please write a ticket for this.

See https://trello.com/c/9WF3NSU0/964-20-extend-createbox-graphql-api-to-accept-tag-ids

PR #411, opened 2022-09-28
Enable running pytest in Docker

pylipp:
Running tests in Docker is helpful because it provides a standardized environment for all devs. It avoids having to debug local issues with Python environments.

Try it:

    docker-compose up --build -d webapp db
    docker-compose exec webapp pytest

None of the changes affects code that would be deployed.


HaGuesto:
@pylipp that's the result i get

pylipp:
> @pylipp that's the result i get

You need the `TEST_AUTH0_CLIENT_ID` and `_SECRET` variables in your .env file. Please see `example.env`, and also Slack, I sent you a PN.

pylipp:
Let me know when you've tried again @HaGuesto then I'll rebase on latest master to resolve the conflict, and also the warnings issued during test runs.

HaGuesto (approved):

PR #412, opened 2022-10-02
Export script for beneficiaries

pylipp:
- Initial version for export-beneficiary script
- Add logging controlled by --verbose option
- Format family head ID if given


HaGuesto:
@pylipp What is the status here? I would question if this is still necessary

pylipp:
Closing due to missing trello ticket and no activity in two years :smiley:

PR #420, opened 2022-10-04
Use client IP address for history table

pylipp:
Previously, creating a box in staging would store the address
127.0.0.1 which is the server's address.


pylipp:
[2.0 add logs in history table for BoxMutation](https://trello.com/c/FcwB42Az/953-20-add-logs-in-history-table-for-boxmutation)

PR #422, opened 2022-10-05
Update test plan

pylipp:
This PR goes in combination with an update of the [test plan](https://docs.google.com/spreadsheets/d/1sDhSsaVwNxAhGn1VYACDPVepZpvlwTkgy9nmeTaRy2Q/edit#gid=51293453), esp. the 2.0 Back-end endpoint tests.

It looks like a couple of changes from the number of changed files but it actually is not :wink: 

- The only functional change is a bug fix in the resolver of the `locations` query
- most other changes are comments referencing test case IDs from the test plan
- I added more tests for creating/updating boxes which is most important for the current milestone
- the rest is about moving tests around to have more consistent order
- I've put info about the test plan structure into the backend readme

FYI @aerinsol


pylipp:
[2.0 Return only classic locations for locations query](https://trello.com/c/v0qm6gQl/970-20-return-only-classic-locations-for-locations-query)

back/README.md
```diff
@@ -218,7 +218,36 @@ Run the test suite on your machine by executing
...
+  - the expected behavior in these categories is
+      1. the response holds the requested (queried/created/modified/deleted) data resource
+      2. the response holds a Forbidden error
+      3. the response holds a BadUserRequest error
+      4. the response holds a BadUserRequest error
+  - due to the nature of GraphQL APIs all responses (successful and erroneous) have HTTP status code 200. The content of the "data" and "errors" fields in the JSON response has to be inspected
+  - test cases for queries are formulated as "*Client requests single X by ID*" or "*Client requests list of Xs*"
+  - test cases for mutations are formulated as "*Client requests operating on X*"
+
+</details>
```
aerinsol, line 248: Super clear, amazing.

aerinsol (approved):
Sweet! We haven't had this problem before, but I'll add something to onboarding to make sure people don't mess with the test case numbering in the code. Maybe we should mention that in the public readme too.

PR #402, opened 2022-09-23
Use Best Practices and Clean Up the FE Configuration Files (tsconfig, eslintrc, prettierrc, etc.)

vahidbazzaz:
In this PR, we clean up tsconfig, prettierrc, and eslintrc, apply best practices, write down folder structures (at least for components), file names conventions, variable names conventions, type creation/enforcement, where is business logic vs view logic, global variables (apollo reactive variable), and explain what packages in package.json do.

PR #436, opened 2022-10-16
Apply linting/formatting to BoxView

vahidbazzaz:
This PR applies formatting and linting to BoxView as well as adding changes to the eslint and prettier configuration

Follow up: https://github.com/boxwise/boxtribute/pull/402


vahidbazzaz:
[2.0 Apply linting/formatting to relevant code in FE](https://trello.com/c/qnk6KSZa)

react/.eslintrc
```diff
@@ -5,15 +5,29 @@
...
+        "import/resolver": {
+            "node": {
+                "extensions": [".ts", ".tsx"],
+                "moduleDirectory": ["src", "node_modules"]
+            }
+        }
+    },
     "rules": {
-        "max-len": ["warn", { "ignoreComments": true }],
+        "max-len": ["warn", { "ignoreComments": true, "code": 200, "tabWidth": 4 }],
```
pylipp, line 21: Hej @vahidbazzaz , can you explain why you made the adjustments to the rules? Did it turn out it does not match our requirements?

vahidbazzaz: @pylipp Thank you for reviewing the PR. To make it more readable, I think we should set `code` length to `100` and reduce the `tabWidth` to 2. What do you think? I will update the documents if you agree with this change

pylipp: I won't work in the FE codebase much, so I don't have that much of an opinion. Makes sense what you're saying though, go ahead!

pylipp (approved):

PR #455, opened 2022-10-26
Make resource based permission being enforced for bases by default, part I

pylipp:
This is the first of a series of at least two PRs related to enforcing base-specific permissions. I'm breaking it down because after the authz logic was changed, there are quite some effects on the resolvers (mostly because base-specific permissions were (lazily) not enforced, or because the permission level was clarified, e.g. for transfer agreement to also operate on base-level). This means a couple of tests are failing but I'll fix them in subsequent PRs which will be merged here one by one. Eventually this branch can go into master.

You can read the background in the attached Trello card.

The main functional change is in cdb1bac4.

### DIFF
--- a/back/boxtribute_server/authz.py
+++ b/back/boxtribute_server/authz.py
@@ -137,3 +137,12 @@ def agreement_organisation_filter_condition():
     return (TransferAgreement.source_organisation == g.user.organisation_id) | (
         TransferAgreement.target_organisation == g.user.organisation_id
     )
+
+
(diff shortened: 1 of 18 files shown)


pylipp:

pylipp:

HaGuesto (approved):
@pylipp I went through the trello, Google Doc, and the PR.

Looks fine from my side. Maybe I just have one or two questions later.

PR #456, opened 2022-10-28
BoxEdit: Add Tags Form Field

vahidbazzaz:
In this PR, tags are added to the box edit form.


react/src/views/BoxEdit/BoxEditView.tsx
```diff
@@ -19,6 +19,11 @@ export const BOX_BY_LABEL_IDENTIFIER_AND_ALL_PRODUCTS_QUERY = gql`
       }
       numberOfItems
       comment
+      tags {
+        value: id
+        label: name
```
pylipp, line 24: Isn't it confusing down the line to have variable names that are different from the original GraphQL field names?

react/src/views/BoxEdit/components/BoxEdit.tsx
```diff
@@ -263,6 +277,36 @@ function BoxEdit({
...
+            <Controller
+              control={control}
+              name="tags"
+              render={({
+                field: { onChange, onBlur, name, value, ref },
+                fieldState: { error },
+              }) => (
+                <FormControl isInvalid={!!error} id="tags">
+                  <FormLabel>Tags</FormLabel>
+                  <Box border="2px">
```
pylipp, line 290: Is it true that you don't use the tag color here?

vahidbazzaz: @pylipp Yes, I want to use tag color and display tags with background colors in the tags selection input, but that turns out to be more complex and outside the scope of this PR, but I keep it on query so we can implement it later.

react/src/views/BoxEdit/BoxEditView.tsx
vahidbazzaz: @pylipp The alias is used since we just need to display the results as select options

react/src/views/BoxEdit/BoxEditView.tsx
```diff
@@ -42,6 +47,12 @@ export const BOX_BY_LABEL_IDENTIFIER_AND_ALL_PRODUCTS_QUERY = gql`
       }
     }
 
+    tags {
```
HaGuesto, line 50: @vahidbazzaz Can't you use a fragment here? I do not like it in general if I see the same code lines twice.

react/src/views/BoxEdit/BoxEditView.tsx
```diff
@@ -110,14 +123,19 @@ function BoxEditView() {
     // eslint-disable-next-line no-console
```
HaGuesto, line 123: @vahidbazzaz please remove the console logs

react/src/views/BoxEdit/BoxEditView.tsx
```diff
@@ -110,14 +123,19 @@ function BoxEditView() {
...
+      ? boxFormValues?.tags?.map((tag) => parseInt(tag.value, 10))
+      : [];
+
     updateContentOfBoxMutation({
       variables: {
         boxLabelIdentifier: labelIdentifier,
         productId: parseInt(boxFormValues.productId, 10),
         numberOfItems: boxFormValues.numberOfItems,
-        comment: boxFormValues.comment,
+        comment: boxFormValues?.comment ? boxFormValues?.comment : "",
```
HaGuesto, line 135: Why are you changing this? @vahidbazzaz

react/src/views/BoxEdit/BoxEditView.tsx
```diff
@@ -110,14 +123,19 @@ function BoxEditView() {
...
         sizeId: parseInt(boxFormValues.sizeId, 10),
         locationId: parseInt(boxFormValues.locationId, 10),
+        tagIds,
       },
     })
       .then((mutationResult) => {
```
HaGuesto, line 141: @vahidbazzaz I hope to see a PR where you also change the error handling away from the console using your global component.

react/src/views/BoxEdit/components/BoxEdit.tsx
HaGuesto: This looks perfect and easy for adding color to tags:
https://github.com/anubra266/choc-autocomplete#multi-select-with-tags

react/src/views/BoxEdit/BoxEditView.tsx
vahidbazzaz: @HaGuesto I think it would be better to do this as part of another PR to clean up and refactor GraphQL queries.

react/src/views/BoxEdit/BoxEditView.tsx
vahidbazzaz: @HaGuesto It is actually outside the scope of this PR, but I will change it in the follow-up PRs.

react/src/views/BoxEdit/components/BoxEdit.tsx
vahidbazzaz: @HaGuesto Sure, I'll check this out, but I think it could also be done as a separate PR

react/src/views/BoxEdit/BoxEditView.tsx
HaGuesto: Just wanted to remind you :)

react/src/views/BoxEdit/BoxEditView.tsx
HaGuesto: @vahidbazzaz I see what you are saying, but the thing is that by not doing it now you adding scope to a later refactor. I'm fine, but in general, please to write code that is clean.
We already gave up on dropapp, I do not want to end up in the same scenario with this front end code.

HaGuesto (approved):
functionally it works.

react/src/views/BoxEdit/BoxEditView.tsx
aerinsol: I agree with Hans, we should follow the DRY principle

PR #468, opened 2022-11-06
Update permissions enforcement part II

pylipp:
This PR builds on #455, fixing a couple of tests that were broken in that PR.

I especially updated resolvers for shipment. To this end it was required to extend the `authorize()` function to authorize the user against several bases (when accessing a shipment, the user must be a member of either source or target base).

In a third PR I will fix the remaining tests corresponding to transfer agreement resolvers.


pylipp:

HaGuesto (approved):

PR #479, opened 2022-11-11
Change allowed mutations for beta user scope

pylipp:
Pretty sure that `updateBox` is the only mutation involved in BoxView/BoxEdit, please correct me if wrong.

### DIFF
--- a/back/boxtribute_server/authz.py
+++ b/back/boxtribute_server/authz.py
@@ -150,6 +150,14 @@ def authorize_for_organisation_bases():
     _authorize(permission="base:read", ignore_missing_base_info=True)
 
 
+ALL_ALLOWED_MUTATIONS = {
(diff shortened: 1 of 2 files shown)


pylipp:
[2.0 Grant beta access for updateBox to all beta_scope=0 users](https://trello.com/c/fQyLZDkA/1011-20-grant-beta-access-for-updatebox-to-all-betascope0-users)

pylipp:
@HaGuesto @vahidbazzaz what do you think about removing the `mobile_distro_feature_flag_check` in favor of the `beta_user` metadata? All users from `MOBILE_DISTRO_ALLOWED_USER_IDS` should be given a dedicated scope, e.g. `beta_user=99`. OTOH the mobile-distro features won't be accessible anyways when the navigation menu is eventually trimmed down. But I'd like to clean up regardless.

PR #493, opened 2022-11-17
Fix scrap and lost changing state and few minor bugs

vahidbazzaz:
In this PR, the following FE bugs are fixed:
- The Scrap and Lost toggle problems have been fixed, as well as the empty comment
- Lost and Scrap are removed from the drop-down menu for the location


vahidbazzaz:
[2.0 BoxView: empty comments are shown in BoxView](https://trello.com/c/kaSb3TH1)
 [2.0 (regression) BoxEdit: remove LOST and SCRAP from locations menu](https://trello.com/c/0m61pWsS)
 [2.0 (regression) BoxView: scrap or lost cannot be unticked.](https://trello.com/c/dUult8Vh)

react/src/views/BoxEdit/BoxEditView.tsx
```diff
@@ -150,10 +157,15 @@ function BoxEditView() {
   const productAndSizesData = data?.products;
   const allTags = data?.tags || null;
 
-  const allLocations = data?.box?.location?.base?.locations.map((location) => ({
-    ...location,
-    name: location.name ?? "",
-  }));
+  const allLocations = data?.box?.location?.base?.locations
```
pylipp, line 160: Is it correct that this is a list of all locations with defaultBoxState different from Lost/Scrap?

vahidbazzaz: @pylipp yes indeed, this all the locations retrieved from the query and in the next line filter out the Scrap and Lost according to the defaultBoxState.

react/src/views/Box/components/BoxDetails.tsx
```diff
@@ -221,7 +219,14 @@ function BoxDetails({
                 isChecked={boxData.state === BoxState.Scrap}
                 data-testid="box-scrap-btn"
                 isFocusable={false}
-                onChange={onScrap}
+                onChange={() =>
+                  onStateChange(
+                    // toggle the box state
```
pylipp, line 224: Can you put a more precise comment, like "If the current box state 'Scrap' is toggled, set the defaultBoxState of the box location"? If my understanding is correct

react/src/views/BoxEdit/BoxEditView.tsx
pylipp: Okay, might it make sense to mention this in a short comment

vahidbazzaz:
> Could not review it functionally but the code looks straightforward. @vahidbazzaz what happens when LOST is active and the user toggles SCRAP?

@pylipp It will inactive the Lost and activate the Scrap, so that should work

pylipp (approved):

PR #401, opened 2022-09-19
Set SENTRY_ENVIRONMENT variable when building front-end

pylipp:
Depending on the CircleCI context, the value is
staging/demo/production. See also
https://docs.sentry.io/platforms/javascript/guides/react/configuration/options/#environment

PR #496, opened 2022-11-19
Prefix react app env variable to be recognized

pylipp:
https://create-react-app.dev/docs/adding-custom-environment-variables/

Follow-up to #302 and #401.


pylipp:
[2.0 Add Sentry to new FE](https://trello.com/c/zVs8Ci8l/923-20-add-sentry-to-new-fe)

PR #499, opened 2022-11-21
Box history GraphQL API

pylipp:
This PR brings an extension of the GraphQL API such that the history of a Box can be displayed in the FE.

For orientation I took the content of the history entries in Dropapp:

    Dev Head of Operations on Monday 14 November 2022, 18:32
    Dev Head of Operations changed the number of items from 15 to 17;

So the FE needs to know about the name of the user who committed the change, the change date, and the change message.

### DIFF
--- a/back/boxtribute_server/models/crud.py
+++ b/back/boxtribute_server/models/crud.py
@@ -15,7 +15,9 @@ from .definitions.beneficiary import Beneficiary
 from .definitions.box import Box
 from .definitions.history import DbChangeHistory
 from .definitions.location import Location
+from .definitions.product import Product
 from .definitions.qr_code import QrCode
+from .definitions.size import Size
 from .definitions.tag import Tag
 from .definitions.tags_relation import TagsRelation
 from .definitions.x_beneficiary_language import XBeneficiaryLanguage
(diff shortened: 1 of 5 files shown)


pylipp:
[2.0 Box history GraphQL API](https://trello.com/c/fst6BLCr/1015-20-box-history-graphql-api)

back/boxtribute_server/models/crud.py
```diff
@@ -435,3 +438,48 @@ def create_qr_code(box_label_identifier=None):
...
+            old_size = Size.get_by_id(raw_entry.from_int)
+            new_size = Size.get_by_id(raw_entry.to_int)
+            changes = f"changed size from {old_size.label} to {new_size.label};"
+
+        elif changes.startswith("comments"):
+            changes = changes.replace("comments changed", "changed comments")
+
+        elif changes.startswith("Record"):
+            changes = changes.replace("Record created", "created record")
+
```
vahidbazzaz, line 482: @pylipp, you can also make a history recode for changes to the state of the box. This is also part of this [PR](https://github.com/boxwise/dropapp/pull/502) in the DropApp. This can, however, be done as a follow-up PR once the mentioned PR has been merged.

vahidbazzaz (approved):

back/boxtribute_server/models/crud.py
pylipp: I see! I don't even take into account that changing the state is logged in `save_update_to_history`. The relevant bit in the dropapp PR seems to be [this](https://github.com/boxwise/dropapp/pull/502/files#diff-d3e74dca303d72c858762488d8e0381e3382a6b2dddad344931aeb342317f716R91-R94), though I don't fully understand the difference of the two messages (why is the information text-only and not using `from_int` and `to_int` fields of the history table?)

PR #500, opened 2022-11-21
Fix Products/Tags/Locations Queries of Multiple Bases Returned in BoxView/BoxEdit

vahidbazzaz:
The PR contains fixes for the products, tags, and locations returned for the multiple bases, as well as the refactoring of BoxView and BoxEdit queries using GraphQL fragments.


vahidbazzaz:
[2.0 Front-end: product/locations/tags queries of multiple bases returned in BoxView/BoxEdit](https://trello.com/c/BH50KP16)

react/src/views/BoxEdit/BoxEditView.tsx
```diff
@@ -2,80 +2,62 @@ import { gql, useMutation, useQuery } from "@apollo/client";
...
-        }
+        ...ProductFields
       }
       location {
         ... on ClassicLocation {
           defaultBoxState
         }
         id
         name
-        base {
```
pylipp, line 46: Good catch moving this part :+1:

react/src/views/BoxEdit/BoxEditView.tsx
```diff
@@ -109,15 +91,17 @@ export const UPDATE_CONTENT_OF_BOX_MUTATION = gql`
...
+    BoxByLabelIdentifierAndAllProductsWithBaseIdQuery,
+    BoxByLabelIdentifierAndAllProductsWithBaseIdQueryVariables
+  >(BOX_BY_LABEL_IDENTIFIER_AND_ALL_PRODUCTS_WITH_BASEID_QUERY, {
     variables: {
+      baseId,
       labelIdentifier,
     },
   });
-  const { baseId } = useParams<{ baseId: string }>();
+  // const { baseId } = useParams<{ baseId: string }>();
```
pylipp, line 104: Can be fully removed

react/src/utils/fragments.ts
```diff
@@ -0,0 +1,134 @@
...
+      }
+      base {
+        locations {
+          id
+          name
+          ... on ClassicLocation {
+            defaultBoxState
+          }
+        }
+        distributionEventsBeforeReturnedFromDistributionState {
```
pylipp, line 121: Do you think the distro-event parts can be simply removed (if so, in a follow-up PR but only after the first release I'd say)? Overall it's overhead queries for the BE, and the response data is not used atm in the UI.

pylipp (changes requested):
Thanks for the quick work @vahidbazzaz ! Using fragments seems to slim down the code a lot, glad you figured it out (though a separate PR would've made the changes easier to distinguish).
How did you solve the issue about the graphql-types not being generated?

react/src/utils/fragments.ts
vahidbazzaz: > Thanks for the quick work @vahidbazzaz ! Using fragments seems to slim down the code a lot, glad you figured it out (though a separate PR would've made the changes easier to distinguish). How did you solve the issue about the graphql-types not being generated?

Thanks @pylipp, it was a little strange. Instead of altering the previous GraphQL query, I created a new one with the `baseId` parameter, then the type generator created new types, and I removed the old one.

vahidbazzaz: > Do you think the distro-event parts can be simply removed (if so, in a follow-up PR but only after the first release I'd say)? Overall it's overhead queries for the BE, and the response data is not used atm in the UI.

@pylipp We may need to consult with @aerinsol and @spaudanjo on this, but I don't have any objections.

PR #501, opened 2022-11-22
Add optional filtering for Base.tags by resource type

pylipp:
The FE should have the ability to filter tags acc. to a resource type.

E.g. in the BoxEdit page the queries should be adjusted like:

```
base(id: $baseId) {
    ...
    tags(resourceType: Box) { ... }
    ...
}
```

Otherwise tags that are only applicable to beneficiaries could be selected.


pylipp:
[2.0 Add filtering for Base.tags field by object type](https://trello.com/c/D9u4c8Y6/1026-20-add-filtering-for-basetags-field-by-object-type)

vahidbazzaz (approved):

PR #503, opened 2022-11-22
Fix Tags Visible After Having Removed from Box

vahidbazzaz:
The PR fixes the badge still being visible after having removed the tag from the box by refetching the data after the update mutation. 

Although I tried adding the necessary fields back after the box mutation update in an effort to avoid refetching, it did not work. This is a quick fix.


vahidbazzaz:
[2.0 Tag badge still visible after having removed tag from box](https://trello.com/c/amrhKEXW)

pylipp (approved):
Maybe @spaudanjo has an idea why the original approach did not work out?

PR #502, opened 2022-11-22
Filter deleted products and locations for base

pylipp:
See attached card.

PR #504, opened 2022-11-22
Work-around for zero datetime products.deleted field in database

pylipp:
See commit message for details.

Follow-up to #502.


pylipp:
[Check if the datetime fields are designed correctly (nullable, etc.) and fit for purpose, esp for MySQL 8 requirements #database](https://trello.com/c/GEuf2nne/752-check-if-the-datetime-fields-are-designed-correctly-nullable-etc-and-fit-for-purpose-esp-for-mysql-8-requirements-database)

vahidbazzaz (approved):

PR #507, opened 2022-11-24
Set Sentry traces sample rate

pylipp:
This PR sets Sentry traces sample rate for back-end and front-end.

I'll set these values in CircleCI contexts:

| context | value |
| ------- | ----- |
| staging | 1.0   |
| demo    | 1.0   |
| production | 0.25 |

When debugging in staging and demo, we want all transactions to be sampled. For production, every forth transactions should be sampled (hopefully enough for debugging but not too much such that it becomes noisy). What do you think?


pylipp:
[2.0 Configure Sentry traces sample rate](https://trello.com/c/nIcmNMs7/1032-20-configure-sentry-traces-sample-rate)

vahidbazzaz (approved):

PR #509, opened 2022-11-25
Update logos/icons and serve them through GAE

pylipp:
- Update logos and favicons once more
- Make GAE serve PNG and SVG files


pylipp:
[2.0 Fix manifest.json file](https://trello.com/c/LF0sDV0s/1033-20-fix-manifestjson-file)

PR #519, opened 2022-11-28
Log box state change and format it for box history

pylipp:
See also dropapp#502


pylipp:
[2.0 Log box state change in history](https://trello.com/c/bDMQIgUG/1037-20-log-box-state-change-in-history)

PR #523, opened 2022-11-30
BoxCreate: Base BoxCreate View on BoxEdit

vahidbazzaz:
This PR changes the box create component to be consistent with the box edit component, and also adds tags and comment fields and sorts locations based on `seq`.

The next PR will focus on making a global component for both box create and box edit.


vahidbazzaz:
[2.0 BoxCreate: Base BoxCreate View on BoxEdit](https://trello.com/c/3gAH6Ryz)

pylipp (approved):
I could actually test this on my machine :sunglasses: Looks good, the only suggestion I have is to auto-select the size if there is only one.

pylipp:
[2.0 BoxView/BoxEdit: Sort location options according to seq](https://trello.com/c/yYclaWYK/1031-20-boxview-boxedit-sort-location-options-according-to-seq)

PR #532, opened 2022-12-08
Fixes and Improvements in QRReader/QrReaderOverlay

vahidbazzaz:
This PR contains changes needed to make the QR reader ready for release. As part of this PR error handling was improved and a problem was fixed locating box label identifiers starts with zeros. The interfaces and types were also refactored. In addition, the bulk QR reader is disabled, since it is not yet ready.

Summary of changes: 
- apply linter to QrReader components
- refactor and clean up QR reader 
- make separate component for QrReaderButton 
- add extractQrCodeFromUrl into helpers 
- enable QR Reader on mobile ui - disable bulk mode for current release 
- replace toast with reactive notification component 
- remove console logs 
- refactor QrResolvedValue and introduced QrResolverResultKind enum 
- change the number input to text input to fix casting labelIdentifier starting with zero


vahidbazzaz:
[2.0 QrReader/QrReaderOverlay: implement logic when to forward to BoxView (for existing Box) or BoxCreate (for new Box)](https://trello.com/c/EO9JHWgg)

react/src/components/QrReader/QrReader.tsx
```diff
@@ -25,13 +22,22 @@ export type OnResultFunction = (
...
+
 const isMediaDevicesAPIAvailable = () => {
-  const isMediaDevicesAPIAvailable =
-    typeof navigator !== "undefined" && !!navigator.mediaDevices;
+  // eslint-disable-next-line no-shadow
+  const isMediaDevicesAPIAvailable = typeof navigator !== "undefined" && !!navigator.mediaDevices;
 
+  // eslint-disable-next-line no-empty
   if (!isMediaDevicesAPIAvailable) {
   }
```
pylipp, line 42: Should we delete this block in total if there's no handling?

pylipp (commented):
Functional tests performed:
- box finder with non-existing box: error notification shown :heavy_check_mark: 
- box finder with existing box: BoxView shown :heavy_check_mark: 
- QR scanner: error notification shown :question: regardless of the QR code I try it with (even as admin user)

@vahidbazzaz can you attach an actual QR code that you tested the QR scanner with?
-

react/src/components/QrReaderOverlay/QrReaderOverlayContainer.tsx
```diff
@@ -50,127 +44,163 @@ const boxDataToSuccessQrValue = (boxData: BoxDetailsQuery["box"]) => {
...
+              query: GET_BOX_LABEL_IDENTIFIER_BY_QR_CODE,
+              fetchPolicy: "no-cache",
+              variables: { qrCode },
+            },
+          )
+          .then(({ data, errors }) => {
             const boxLabelIdentifier = data?.qrCode?.box?.labelIdentifier;
-            if (boxLabelIdentifier == null) {
+
+            if ((errors?.length || 0) > 0) {
```
pylipp, line 184: There needs to be more distinction of the error code. 
1. If the QR code is not assigned to a box, looking up the box fails, and the error formatter in `exceptions.py` sets the `BAD_USER_REQUEST` GraphQL error code (unfortunately this is only the behavior for production mode, not for development mode). This should be the `NOT_ASSIGNED_TO_BOX` result.
2. If the QR code is assigned to a box but the user is not authorized to access that box, a `FORBIDDEN` error is returned. This should be the `NOT_AUTHORIZED` result.

vahidbazzaz: @pylipp  Thanks for your suggestion, I handled the error in the FE, and also wrote on more cases for unhandled exceptions on BE (if any)

react/src/components/QrReaderOverlay/QrReaderOverlayContainer.tsx
```diff
@@ -50,127 +44,144 @@ const boxDataToSuccessQrValue = (boxData: BoxDetailsQuery["box"]) => {
...
+              const errorCode = errors ? errors[0].extensions.code : null;
+              if (errorCode === "FORBIDDEN") {
+                onScanningDone([
+                  { kind: QrResolverResultKind.NOT_AUTHORIZED, qrCodeValue: qrCode },
+                ]);
+              } else if (errorCode === "BAD_USER_INPUT") {
+                onScanningDone([
+                  { kind: QrResolverResultKind.LABEL_NOT_FOUND, qrCodeValue: qrCode },
```
pylipp, line 170: In `HeaderMenuContainer.tsx` there is no handling of `LABEL_NOT_FOUND`, is that on purpose?

pylipp: I don't see a way of solving this except for changing the box resolver in the BE.

vahidbazzaz: @pylipp Likewise, I think that it makes sense to differentiate between these two cases. However, I can handle it differently when BAD_USER_INPUT returns in the query belong to the QR code, showing a different error case, and vice versa.

pylipp:
@vahidbazzaz one conflict after @HaGuesto has merged #542 :sweat_smile:

react/src/components/HeaderMenu/HeaderMenuContainer.tsx
```diff
@@ -106,39 +107,39 @@ const HeaderMenuContainer = () => {
             break;
           }
           case QrResolverResultKind.NOT_BOXTRIBUTE_QR: {
-            notificationVar({
+            createToast({
```
pylipp, line 110: I think here and in lines 118, 126, 134 we could use the `triggerError` utility rather to also have the error logged to Sentry.

pylipp (approved):

react/src/components/HeaderMenu/HeaderMenuContainer.tsx
```diff
@@ -98,44 +97,65 @@ const HeaderMenuContainer = () => {
...
+            break;
+          }
+          case QrResolverResultKind.NOT_ASSIGNED_TO_BOX: {
+            createToast({
+              title: "QR Code",
+              type: "info",
+              message: "Scanned QR code is not assigned to a box yet",
+            });
+
+            navigate(`/bases/${baseId}/boxes/create?qrCode=${singleResolvedQrValue?.qrCodeValue}`);
```
pylipp, line 148: @vahidbazzaz the route has changed, it's `.../create/${singleResolvedQrValue?.qrCodeValue}`

vahidbazzaz: @pylipp good catch. I'll change it to the new route.

pylipp (approved):
Ready to be merged from my side. @HaGuesto do you want to test it too?

HaGuesto:
[Moving Session Storage to GCloud DataStore](https://trello.com/c/xAglnJW2/1173-moving-session-storage-to-gcloud-datastore)

PR #541, opened 2022-12-16
Refactor BoxCreate (Merge First)

HaGuesto:


react/src/views/BoxCreate/BoxCreateView.tsx
```diff
@@ -65,25 +69,84 @@ export const ALL_PRODUCTS_AND_LOCATIONS_FOR_BASE_QUERY = gql`
...
 
 function BoxCreateView() {
+  // Basics
+  const navigate = useNavigate();
+
+  // variables in URL
   const baseId = useParams<{ baseId: string }>().baseId!;
-  const [searchParams] = useSearchParams();
-  const qrCode = searchParams.get("qrCode") as string | undefined;
+  const qrCode = useParams<{ qrCode: string }>().qrCode!;
```
pylipp, line 77: Why was it not possible to use the QR code from the URL parameter?

react/src/views/BoxCreate/BoxCreateView.tsx
```diff
@@ -65,25 +69,84 @@ export const ALL_PRODUCTS_AND_LOCATIONS_FOR_BASE_QUERY = gql`
...
+  // Mutation after form submission
   const [createBoxMutation, createBoxMutationState] = useMutation<
     CreateBoxMutation,
     CreateBoxMutationVariables
   >(CREATE_BOX_MUTATION);
 
+  // Check the QR Code
+  useEffect(() => {
+    if (qrCodeExists.data?.qrExists === false) {
+      notificationVar({
```
pylipp, line 106: Note for later that this can be refactored with the enums that Vahid introduced in his PR.

pylipp (commented):
Still have to functionally review it but looks good, thanks for the comments too!

HaGuesto:
[2.0 Add QrCode check to BoxCreate and refactor Error Messages](https://trello.com/c/wrkI5JhN/1045-20-add-qrcode-check-to-boxcreate-and-refactor-error-messages)

react/src/views/BoxCreate/BoxCreateView.tsx
HaGuesto: In general, we should refactor this business logic and put it into a hook. @pylipp @vahidbazzaz 
This is just a quick fix.

react/src/views/BoxCreate/BoxCreateView.tsx
HaGuesto: It is possible, but with this implementation the QRCode becomes a required instead of an optional parameter.
We can roll this back when we have another use case where we create Boxes without QRCodes.

pylipp (approved):
@HaGuesto functional review:
1. using the endpoint `bases/1/boxes/create` I see an empty page (I assume that's intentional)
1. using the endpoint `bases/1/boxes/create/03727fc6712dd6c7717d7a151733a76` (this QR code is not yet associated with a box) I get the CreateBox page and can create a box
1. using the same endpoint as in (2.) I get an error message that the box could not be created which is expected (it's not possible to add another box for a QR code that is already associated with a box)

pylipp:
@HaGuesto one more thing:
4. using the endpoint `bases/1/boxes/create/149ff66629377f6404b5c8d32936855` (an unknown QR code) I get an error message "The QR code is not from boxtribute". But I stay on an empty page with the same URL. Should we auto-redirect to dropapp BoxesView?

PR #542, opened 2022-12-16
Fix/error handling infrastructure (merge 2nd)

HaGuesto:
Probs there will be some conflicts in CreateBoxView


HaGuesto:
[2.0 handle generic backend errors / create system including user notification and sentry reporting](https://trello.com/c/l1NBVFUv/1044-20-handle-generic-backend-errors-create-system-including-user-notification-and-sentry-reporting)

HaGuesto:
I merged #541 into this so you can merge the PRs one by one

react/src/utils/error-handling.ts
```diff
@@ -0,0 +1,35 @@
...
+      if (statusCode) console.error(`[${statusCode}] ${message}`);
+      else console.error(`${message}`);
+
+      toast({
+        duration: 5000,
+        isClosable: true,
+        position: "top",
+        variant: "subtle",
+        status: "error",
+        title: "Error",
```
pylipp, line 25: Should we also make the toast title customizable as before, e.g. for displaying the box label identifier?

react/src/views/BoxEdit/BoxEditView.tsx
```diff
@@ -185,19 +184,15 @@ function BoxEditView() {
     }));
 
   if (allLocations == null) {
-    notificationVar({
-      title: "Error",
-      type: "error",
-      message: "Error: No other locations are visible!",
+    triggerError({
+      message: "No locations are available!",
```
pylipp, line 188: Changing "visible" to "available" is fine for me, sounds better :+1:

react/src/views/BoxEdit/BoxEditView.tsx
```diff
@@ -185,19 +184,15 @@ function BoxEditView() {
...
     return <div />;
   }
 
   if (productAndSizesData == null) {
-    notificationVar({
-      title: "Error",
-      type: "error",
-      message: "Error: No products are visible!",
+    triggerError({
+      message: "No products are available!",
```
pylipp, line 195: I tested this message by setting `const productAndSizesData = null;` above; I then see the message three times, not sure if that's due to recent changes or if that's been a bug before.

pylipp: Also in case of an error in BoxEditView or BoxEdit, can we return to BoxView of that box?

react/src/views/Box/BoxView.tsx
```diff
@@ -232,10 +234,8 @@ function BTBox() {
     assignBoxToDistributionEventMutationStatus.error ||
     unassignBoxFromDistributionEventMutationStatus.error
   ) {
-    notificationVar({
-      title: "Error",
-      type: "error",
-      message: "Error: Could not update the box",
+    triggerError({
+      message: "Could not update the box.",
```
pylipp, line 238: Not directly related to the PR, however for `error` it would make sense to have a message like `Could not fetch box data`. For mutation errors the message above makes sense.

pylipp (commented):
Nice condensing of triggering an error into a single function call without repeated parameters. Functional review done from my end, with a couple of things to clarify.

react/src/utils/error-handling.ts
pylipp: Clarified: customizable through `createToast` hook

react/src/views/BoxEdit/BoxEditView.tsx
pylipp: Clarified: @HaGuesto integrates `useEffect` to avoid multiple notifications displayed.

HaGuesto: I added the useEffect and error messages only appear once on my side.
Regarding the forwarding I did not find a quick solution and left it for now.

react/src/views/Box/BoxView.tsx
HaGuesto: changed this

HaGuesto:
@pylipp I also moved all data checking out of BoxCreate and BoxEdit to BoxCreateView and BoxEditView since that's the place where the data should be handled.

react/src/views/BoxEdit/BoxEditView.tsx
pylipp: Also works on my end :+1:

pylipp (approved):

pylipp:
> @pylipp I also moved all data checking out of BoxCreate and BoxEdit to BoxCreateView and BoxEditView since that's the place where the data should be handled.

Thanks for the clean up, makes sense to be consistent here for separating UI and business logic.

PR #466, opened 2022-11-04
Feature/disable menu

johnsmol:
Disabled buttons on menu header for both desktop and mobile visualizations.
Added home button to navigate back to the currently selected base homepage.

PR #545, opened 2022-12-17
edit links to dropapp mobile views

HaGuesto:
Reason:
- set preference Session variable in dropapp to classic
- go to the same menu in dropapp on mobile.

### DIFF
--- a/react/src/utils/helpers.ts
+++ b/react/src/utils/helpers.ts
@@ -1,25 +1,50 @@
 export const getISODateTimeFromDateAndTimeString = (date: Date, timeString: string) => {
-  const [hours, minutes] = timeString.split(':').map(Number);
+  const [hours, minutes] = timeString.split(":").map(Number);
   const dateTime = new Date(date);
   dateTime.setHours(hours);
   dateTime.setMinutes(minutes);
   return dateTime;
-}
+};
 
 export const getISODateTimeFromDateAndTime = (date: Date, time: Date) => {
   const dateTime = new Date(date);
   dateTime.setHours(time.getHours());
   dateTime.setMinutes(time.getMinutes());
   return dateTime;
-}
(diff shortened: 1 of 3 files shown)


react/src/utils/helpers.ts
```diff
@@ -1,25 +1,50 @@
...
-}
+};
+
+export const generateDropappUrl = (
+  oldLink: String,
+  baseId: String | undefined,
+  qrCode: String | undefined,
+  labelIdentifier: String | undefined,
+) => {
+  const newLink = "https://" + oldLink + "?camp=" + baseId + "&preference=classic";
```
pylipp, line 41: `oldLink` already contains the protocol prefix, so we don't have to specify it another time. See [this comment](https://github.com/boxwise/boxtribute/pull/466#discussion_r1027113171). Also it then works when running locally (for `REACT_APP_OLD_APP_BASE_URL` I have set `http://localhost:8100`; if https:// is added the link becomes mixed up)

pylipp (changes requested):
The links work locally when removing the https:// prefix (see comment). Why is the precise redirecting for mobile only?
Can you link a Trello card?

react/src/utils/helpers.ts
HaGuesto: I removed the protocol.

HaGuesto:
> Why is the precise redirecting for mobile only? 
Only for the menuItem "Go to classic mobile" corresponding views exist in dropapp. So, only then should a different link then the homepage be created.

HaGuesto:
[2.0 iteration on forwarding to dropapp](https://trello.com/c/ZDlfoF7n/1048-20-iteration-on-forwarding-to-dropapp)

pylipp (approved):

PR #546, opened 2022-12-18
Display Box History Component for BoxView

vahidbazzaz:
This PR adds the history block to the box view, the flow up PR adds the overlay to show detail history, and also add history on the box edit


vahidbazzaz:
[2.0 BoxHistory component for BoxView and BoxEdit](https://trello.com/c/SWNWJVq9)

react/src/views/Box/components/HistoryEntries.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const historyData = [...data].sort(
+    (a, b) => new Date(b.changeDate).getTime() - new Date(a.changeDate).getTime(),
+  );
+
+  // show different icons for changed / created records also default history for the rest
+  const getHistoryIcon = (changes: string): IconType => {
+    if (changes.includes("created")) {
+      return MdSettings;
+    }
+    if (changes.includes("changed")) {
```
pylipp, line 22: If changes include both "created" and "changed", this block overrides the first icon, is that on purpose? Otherwise `else if` probably makes sense

react/src/views/Box/components/HistoryEntries.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { IconType } from "react-icons";
+import { MdCheckCircle, MdSettings, MdHistory } from "react-icons/md";
+import { HistoryEntry } from "types/generated/graphql";
+
+interface IHistoryEntriesProps {
+  data: HistoryEntry[];
+  total: number | undefined;
+}
+function HistoryEntries({ data, total }: IHistoryEntriesProps) {
+  // sort history entries in the descending order
```
pylipp, line 11: I'll create a backend PR to have the entries sorted by most recent first already, then the logic can be removed here.

pylipp (approved):
Functional review:
- display history of existing box :heavy_check_mark: 
- history updates when changing box properties in BoxView :heavy_check_mark: 
- history updates when coming from BoxEdit :heavy_check_mark: 
- no history shown for box without history :heavy_check_mark: 
- history shown for that box when changing box properties in BoxView :heavy_check_mark: 
- history shown when coming from BoxCreate :heavy_check_mark: 

Code review: Just one small remark. LGTM :slightly_smiling_face:

react/src/views/Box/components/HistoryEntries.tsx
pylipp: @vahidbazzaz I might directly push the change to this branch, already tested it locally

vahidbazzaz: @pylipp sure please push the change, that should be fine

react/src/views/Box/components/HistoryEntries.tsx
vahidbazzaz: @pylipp it was on purpose but we can change later if needed

PR #548, opened 2022-12-18
Remove the Logo Link and Add State Label on the Location Buttons

vahidbazzaz:
The PR contains minor changes before the release, including removing the link from the logo and adding box state to the locations button and drop down menu in the edit box when the location associated box state is not equal to InStock.

### DIFF
--- a/react/src/views/Box/components/BoxDetails.tsx
+++ b/react/src/views/Box/components/BoxDetails.tsx
@@ -288,7 +288,7 @@ function BoxDetails({
                     {location.defaultBoxState !== boxData.state && (
                       <>
                         {" "}
-                        - Boxes are &nbsp;<i> {location.defaultBoxState}</i>
+                        - Boxes are&nbsp;<i> {location.defaultBoxState}</i>
                       </>
                     )}
                   </Button>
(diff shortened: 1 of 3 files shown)


react/src/views/Box/components/BoxDetails.tsx
```diff
@@ -285,6 +285,12 @@ function BoxDetails({
                     border="2px"
                   >
                     {location.name}
+                    {location.defaultBoxState !== boxData.state && (
+                      <>
+                        {" "}
+                        - Boxes are &nbsp;<i> {location.defaultBoxState}</i>
```
pylipp, line 291: 2022-12-18-135851_618x547_scrot

The space between "Boxes are" and the state name could be smaller IMHO.

pylipp (approved):
Makes sense to not direct to the page root in the new app when clicking the logo.

If the changes on the location buttons are agreed with @aerinsol it's fine for me.

react/src/views/Box/components/BoxDetails.tsx
pylipp: Removing the space before `nbsp` looks better I find

pylipp:
@vahidbazzaz can you link a Trello card for this?

vahidbazzaz:
[2.0 Remove the Logo Link and Add State Label on the Location Buttons](https://trello.com/c/l66dzr9r)

vahidbazzaz:
Screenshot 2022-12-18 at 6 14 20 PM

HaGuesto (approved):

PR #554, opened 2022-12-19
BoxView: Disabling Buttons on State Lost and Scrap

vahidbazzaz:
In this PR, the buttons are disabled if the box state is lost or scrapped.


vahidbazzaz:
[2.0 BoxView: Disabling Buttons on State Lost and Scrap](https://trello.com/c/DZlLJEBS)

pylipp (approved):

pylipp:
Agreement: we still display the "Move box to ..." container to (a) not confuse the user (they might wonder whether the container did not render) and (b) show the current location of the box.

PR #559, opened 2022-12-22
BoxView: validate input of AddItemsOverlay TakeItemsOverlay

vahidbazzaz:
In this PR, the items overlay is validated and the input field is changed to a number.

### DIFF
--- a/react/src/views/Box/BoxView.tsx
+++ b/react/src/views/Box/BoxView.tsx
@@ -286,6 +286,7 @@ function BTBox() {
       boxFormValues.numberOfItems > 0 &&
       (boxData?.numberOfItems || boxData?.numberOfItems === 0)
     ) {
+      // The number of items must be less than the maximum MySQL signed integer value
       if ((boxData.numberOfItems || 0) + boxFormValues.numberOfItems > 2147483647) {
         createToast({
           title: `Box ${boxData.labelIdentifier}`,
--- a/react/src/views/Box/BoxView.tsx
+++ b/react/src/views/Box/BoxView.tsx
@@ -248,27 +248,35 @@ function BTBox() {
 
   const onSubmitTakeItemsFromBox = (boxFormValues: IChangeNumberOfItemsBoxData) => {
     if (boxFormValues.numberOfItems && boxFormValues.numberOfItems > 0 && boxData?.numberOfItems) {
-      updateNumberOfItemsMutation({
-        variables: {
-          boxLabelIdentifier: labelIdentifier,
-          numberOfItems: (boxData?.numberOfItems || 0) - (boxFormValues?.numberOfItems || 0),
-        },
-      })
-        .then((res) => {
-          createToast({
-            title: `Box ${boxData.labelIdentifier}`,
-            type: res.errors ? "error" : "success",
-            message: res.errors
-              ? "Error: Could not remove items from the box"
-              : `Successfully removed ${boxFormValues?.numberOfItems} items from box`,
-          });
-          onMinusClose();
-        })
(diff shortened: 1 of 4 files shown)


vahidbazzaz:
[2.0 BoxView: validate input of AddItemsOverlay TakeItemsOverlay](https://trello.com/c/AMrWZTrl)

react/src/views/Box/BoxView.tsx
```diff
@@ -248,27 +248,35 @@ function BTBox() {
...
+        createToast({
+          title: `Box ${boxData.labelIdentifier}`,
+          type: "error",
+          message: `Could not remove more than ${boxData?.numberOfItems} items`,
+        });
+      } else {
+        updateNumberOfItemsMutation({
+          variables: {
+            boxLabelIdentifier: labelIdentifier,
+            numberOfItems: (boxData?.numberOfItems || 0) - (boxFormValues?.numberOfItems || 0),
```
pylipp, line 261: I think in line 250 it's already checked for values being defined, so the `|| 0` fallback might be redundant here? Or does eslint complain?

vahidbazzaz: @pylipp yes indeed eslint complain

pylipp (changes requested):
I did a couple of tests trying to break the validation, it did survive :laughing: I like that it's not possible by default to enter anything but numbers.

1. I let the back-end raise an error when updating a box. When submitting on the TakeItemsForm, I then see two error messages ("Could not remove items from the box", which is expected) but also "Something went wrong", and then an empty page (but the URL in the browser shows `/bases/2/boxes/629029`). I would expect the BoxView to be shown correctly
2. In BoxEdit we use zod and small texts below the input fields to give hints about validation errors. Here error toasts are used, so it's feels a bit inconsistent to me. Should probably be @aerinsol who decides.

vahidbazzaz:
> I did a couple of tests trying to break the validation, it did survive 😆 I like that it's not possible by default to enter anything but numbers.
> 
> 1. I let the back-end raise an error when updating a box. When submitting on the TakeItemsForm, I then see two error messages ("Could not remove items from the box", which is expected) but also "Something went wrong", and then an empty page (but the URL in the browser shows `/bases/2/boxes/629029`). I would expect the BoxView to be shown correctly
> 2. In BoxEdit we use zod and small texts below the input fields to give hints about validation errors. Here error toasts are used, so it's feels a bit inconsistent to me. Should probably be @aerinsol who decides.

@pylipp Thank you for reviewing the PR.

1.  Regarding the error message, this is something that was introduced with the recent changes to error messages. One global error handler is implemented in the Apollo Provider, and the other is captured in the boxview. I am not sure why the page did not reload, but it might be because the Apollo client did not fetch box data when the error occurred.
2. I use toast message here because there is only one field, and showing error in the field does not look right. However, I do not have any objections to changing the way the error appears in the overlay.

pylipp:
@vahidbazzaz if you have some time left, could you check out why (1.) happens (in a new PR). I don't think it's user-friendly to show an empty page after an error (if I remember correctly, there was no navigation visible then). Maybe @HaGuesto knows more about how the global error handling works and why this happens.

pylipp (approved):

pylipp:
> @vahidbazzaz if you have some time left, could you check out why (1.) happens (in a new PR). I don't think it's user-friendly to show an empty page after an error (if I remember correctly, there was no navigation visible then). Maybe @HaGuesto knows more about how the global error handling works and why this happens.

As discussed there will be a separate PR to work on not returning empty `<div>` in case of errors.

PR #560, opened 2022-12-23
Add heap.io on Frontend

vahidbazzaz:
This PR integrates Heap Analytics into the front-end by adding scripts to index HTML during deployment. In response to these changes, an environment variable HEAPIO_APP_ID is created in the CircleCI to store the heap app ID.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -183,8 +183,8 @@ jobs:
       - run:
           name: Inject Heap Analytics Script to index.html
           command: |
-            sed -i 's/HEAPIO_APP_ID/${HEAPIO_APP_ID}/' public/heapanalytics.js
-            sed -i 's|</head>| i <script src="/heapanalytics.js"></script>|' public/index.html
+            sed -i 's/HEAPIO_APP_ID/${HEAPIO_APP_ID}/' ../react/build/heapanalytics.js
+            sed -i 's|</head>| i <script src="/heapanalytics.js"></script>|' ../react/build/index.html
       - run:
           name: Copy react build into flask app for deployment
           command: mv ../react/build react-build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,6 +66,8 @@ jobs:
               REACT_APP_SENTRY_ENVIRONMENT=${SENTRY_ENVIRONMENT}
               REACT_APP_SENTRY_TRACES_SAMPLE_RATE=${SENTRY_TRACES_SAMPLE_RATE}
             " > .env
+            sed -i 's/HEAPIO_APP_ID/${HEAPIO_APP_ID}/' public/heapanalytics.js
+            sed -i 's|</head>| i <script src="/heapanalytics.js"></script>|' public/index.html
       # following https://circleci.com/docs/2.0/yarn/
       - restore_cache:
           name: Restore Yarn Package Cache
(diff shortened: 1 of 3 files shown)


vahidbazzaz:
[2.0 Add heap.io on frontend](https://trello.com/c/WRuTISIy)

.circleci/config.yml
```diff
@@ -66,6 +66,8 @@ jobs:
               REACT_APP_SENTRY_ENVIRONMENT=${SENTRY_ENVIRONMENT}
               REACT_APP_SENTRY_TRACES_SAMPLE_RATE=${SENTRY_TRACES_SAMPLE_RATE}
             " > .env
+            sed -i 's/HEAPIO_APP_ID/${HEAPIO_APP_ID}/' public/heapanalytics.js
+            sed -i 's|</head>| i <script src="/heapanalytics.js"></script>|' public/index.html
```
pylipp, line 70: Can you check that this does not cause the tests to send data to heap.io? I think that would be confusing. If that happens we have to add the index.html update into the deploy step.

react/public/heapanalytics.js
```diff
@@ -0,0 +1,33 @@
...
+          "track",
+          "unsetEventProperty",
+        ],
+        o = 0;
+      o < p.length;
+      o++
+    )
+      heap[p[o]] = n(p[o]);
+  });
+heap.load(HEAPIO_APP_ID);
```
pylipp, line 33: Can you add a comment like: "App ID will be substituted during deploy by CircleCI"
And another one at the top of the file "This file is only integrated for deploys"

.circleci/config.yml
pylipp: Also @vahidbazzaz I think it might be required to have 
```
- url: /(.*\.js)$
  secure: always
  static_files: react-build/\1
  upload: react-build/(.*\.js)$
```
in app.yaml to actually provide the heapanalytics file.

pylipp (changes requested):
Thanks for researching, left some comments :)

pylipp:
Is there a way that after the user has authenticated, `heap.identify(userId)` is called?

.circleci/config.yml
vahidbazzaz: @pylipp As you said, this is only required on a deployed site, so putting it in deploy makes sense to me

vahidbazzaz:
> Is there a way that after the user has authenticated, `heap.identify(userId)` is called?

@pylipp I am also still unsure how to handle it because we do not have server-side rendering in the FE, but I think it would be possible to update it when a user logs in, which would be ideal.

.circleci/config.yml
```diff
@@ -182,6 +180,11 @@ jobs:
           at: ~/
       # install gcloud orb
       - gcp-cli/install
+      - run:
+          name: Inject Heap Analytics Script to index.html
+          command: |
+            sed -i 's/HEAPIO_APP_ID/${HEAPIO_APP_ID}/' public/heapanalytics.js
+            sed -i 's|</head>| i <script src="/heapanalytics.js"></script>|' public/index.html
```
pylipp, line 187: The working directory of the job is `back/`, and we want to modify the build artifact, so the file path should be `../react/build/public/index.html` I think.

pylipp: Same for the JS script

back/app.yaml
```diff
@@ -24,3 +24,7 @@ handlers:
   secure: always
   static_files: react-build/index.html
   upload: react-build/index.html
+- url: /(.*\.js)$
```
pylipp, line 27: Can you switch this block with the fallback handler above? Otherwise I think it's never reached

vahidbazzaz: @pylipp I have changed the route with a more secure one that follows a more specific path

pylipp (approved):

PR #561, opened 2022-12-23
Integrate FE style-checks with pre-commit and CircleCI

pylipp:
This PR integrates FE style-checks and the generation of GraphQL Typescript defs with pre-commit, and runs the checks in CircleCI.

It assumes the developer has a yarn installation by having run `docker-compose build react` once.

Now, whenever a commit includes Typescript or Javascript files, pre-commit will invoke eslint and prettier.
Also, when the GraphQL schema is changed, the Typescript types are auto-generated during pre-commit stage (stage them and try to commit again). This avoids discrepancy between the schema and the definitions.

Note that in CircleCI I found it easier to simply invoke yarn rather than dealing with installing pre-commit into the build-and-test-react job first.

For testing, check out this branch and
1. run `pre-commit run --all-files`
2. do some illegal change in a TS file, stage it, and try to commit
3. change something in the `back/boxtribute_server/graph_ql/*.graphql` files, stage it, and try to commit


pylipp:
[2.0 Integrate FE style-checks with pre-commit and CircleCI](https://trello.com/c/UckS9MNf/1064-20-integrate-fe-style-checks-with-pre-commit-and-circleci)

vahidbazzaz (approved):
Thanks @pylipp I have tested this and it seems to work fine. However, I wonder if preventing commits when the local build fails would be helpful, specifically for FE builds.

pylipp:
> 

@vahidbazzaz wouldn't that be a rather expensive operation? Running the pre-commit checks should not interrupt the workflow too much, otherwise I think it becomes annoying.

I tried integrating it locally. On my machine it takes 9s and then fails with a permission error (I think because the system tries to modify files that were previously compiled through the Docker container). I don't think we're missnig a lot when not having this check. `yarn build` is also run during in a CircleCI job, and also continuously from the Docker service `react` if it's running during development. For completeness here's the snippet anyways:

```yaml
      - id: yarn-build
        name: Build front-end package
        entry: yarn
        files: ^react/.*
        types_or: [javascript, jsx, ts, tsx]
        language: system
        args: [--cwd=react, build]
        pass_filenames: false
        require_serial: true
```

PR #569, opened 2022-12-27
Correctly configure CORS in back-end

pylipp:
When running the back-end tests a warning "unknown option passed to
flask-cors: origin" could be seen. It turns out that the option
correctly has to be named 'origins'. But even then the server would
accept any origins for making requests.
More research revealed that the flask-cors offers several ways of
configuring the extension: using the `CORS` wrapper, and using the
`cross_origin` decorator. Since we used both, the former took
precedence, and it's default is to allow any origins, methods, and
headers (questionable defaults, however that's exactly what the docs
state).
As a solution, the `CORS` wrapper is removed, and the `cross_origin`
decorator is applied to the one route that matters, with the correct
options passed in.

FYI @aerinsol

### DIFF
--- a/back/boxtribute_server/app.py
+++ b/back/boxtribute_server/app.py
@@ -3,7 +3,6 @@ import os
 
 import sentry_sdk
 from flask import Flask
-from flask_cors import CORS
 from graphql.error import GraphQLError
 from sentry_sdk.integrations.flask import FlaskIntegration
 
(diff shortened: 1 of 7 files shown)


pylipp:
[2.0 Correctly configure CORS](https://trello.com/c/u9lVwjxT/1065-20-correctly-configure-cors)

pylipp:
@vahidbazzaz I've adjusted the allowed URL origins as per your suggestion :)

vahidbazzaz (approved):

vahidbazzaz:
@pylipp Would you mind adding a note to the readme about this configuration and domain update so that anyone deploying it is aware of these changes. Thank you

pylipp:
@vahidbazzaz that's a good suggestion, I think I'll put it into the readme with parts of [the release instructions](https://docs.google.com/document/u/1/d/1zfqQ4yRUf4C-fbhMWfzKzBWPoDAvt4tF8nVI8fIrQrY/edit#heading=h.5z1kbpap2015)

PR #570, opened 2022-12-30
Update loader-utils Package to Non-vulnerable Version

vahidbazzaz:
This PR is aimed at upgrading the loader-utils package to a version that is not vulnerable. As part of the PR, the storybook package has been updated to the latest version, and the MSW script has also been updated.


pylipp (approved):

pylipp:
Thanks! For reference, [this](https://github.com/boxwise/boxtribute/security/dependabot/52) was the alert that is now fixed, and other should be as well (54-57).

PR #577, opened 2023-01-02
BoxEdit/BoxView/BoxCreate: Add Colors to the Tag Options

vahidbazzaz:
The PR adds colors to tag options for tags that already have colors.

### DIFF
--- a/react/src/components/Form/SelectField.tsx
+++ b/react/src/components/Form/SelectField.tsx
@@ -1,7 +1,7 @@
 import { FormControl, FormErrorMessage, FormLabel } from "@chakra-ui/react";
 import { Select, OptionBase } from "chakra-react-select";
 import { Controller } from "react-hook-form";
-import { getContrastYIQ } from "utils/helpers";
+import { colorIsBright } from "utils/helpers";
 
 export interface IDropdownOption extends OptionBase {
   value: string;
@@ -67,8 +67,8 @@ function SelectField({
               multiValue: (provided, state) => ({
                 ...provided,
                 border: "1px",
-                borderColor: getContrastYIQ(state.data?.color) ? "gray.300" : state.data?.color,
-                color: getContrastYIQ(state.data?.color) ? "black" : "white",
+                borderColor: colorIsBright(state.data?.color) ? "gray.300" : state.data?.color,
+                color: colorIsBright(state.data?.color) ? "black" : "white",
                 background: state.data?.color || "gray.100",
                 borderRadius: "20",
               }),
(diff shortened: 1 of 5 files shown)


vahidbazzaz:
[2.0 BoxEdit: give tags color](https://trello.com/c/2McAE5b1)

pylipp (changes requested):
Code review and functional review done, two remarks in comments :)

react/src/components/Form/SelectField.tsx
```diff
@@ -62,9 +63,10 @@ function SelectField({
                 borderRadius: "0",
                 borderColor: "black",
               }),
-              multiValue: (provided) => ({
+              multiValue: (provided, state) => ({
                 ...provided,
                 border: "1px",
+                background: state.data?.color || "gray.100",
```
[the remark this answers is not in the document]
vahidbazzaz: @pylipp That makes total sense, as the user just selects the background color, then we adjust the font color using contrast ratio math. I'll push the changes.

react/src/components/Form/SelectField.tsx
```diff
@@ -62,9 +63,10 @@ function SelectField({
                 borderRadius: "0",
                 borderColor: "black",
               }),
-              multiValue: (provided) => ({
+              multiValue: (provided, state) => ({
                 ...provided,
                 border: "1px",
```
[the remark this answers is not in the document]
vahidbazzaz: @pylipp If the background color is so bright, a border with higher contrast would make sense. I will push the changes, and you can try them out.

react/src/utils/helpers.ts
```diff
@@ -61,3 +61,12 @@ export const generateDropappUrl = (
 export const redirectToExternalUrl = (url) => {
   window.location.replace(url);
 };
+
+export const getContrastYIQ = (hex) => {
```
pylipp, line 65: Can you add a comment, e.g.
Return true if given RGB hex color is bright, false if dark.
This is achieved by converting the RGB value into YIQ and then checking the Y component (luma). Cf. https://stackoverflow.com/a/946734/3865876

pylipp: Maybe the function name `colorIsBright` would suit better then?

react/src/views/BoxCreate/components/BoxCreate.tsx
```diff
@@ -128,7 +128,7 @@ function BoxCreate({
   const tagsForDropdownGroups: Array<IDropdownOption> | undefined = allTags?.map((tag) => ({
     label: tag.label,
     value: tag.value,
-    // colorScheme: "red",
+    color: tag.color,
```
pylipp, line 131: Good catch :+1:

pylipp (approved):
Works and brings UX enhancements, nice! :tada:

PR #590, opened 2023-01-14
Sort history entries by ID instead of change date

pylipp:
The issue was that we use ordering by change datetime to sort log
entries, and then select the first one as the newest one. However the
datetime has second precision, hence when two changes occur within one
second, it's not necessarily the newest one that comes first in the
ordering. The newest entry always has the largest ID, so sorting by ID
is used now.

@aerinsol I'll merge this such that you can test on staging again.

### DIFF
--- a/back/boxtribute_server/warehouse/box/crud.py
+++ b/back/boxtribute_server/warehouse/box/crud.py
@@ -214,4 +214,4 @@ def get_box_history(box_id):
 
         raw_entry.changes = changes
         entries.append(raw_entry)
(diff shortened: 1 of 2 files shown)


pylipp:
[2.0 BUG: Most recent Box history entry isn't always accurate #sprint-jan-19](https://trello.com/c/1qPAsYwf/1083-20-bug-most-recent-box-history-entry-isnt-always-accurate-sprint-jan-19)

aerinsol (approved):

PR #604, opened 2023-01-19
Add source organisation ID as input argument when creating agreement

pylipp:


pylipp:
[2.0 Add source organisation ID as input argument for creating agreement](https://trello.com/c/PxYGNMjx/1095-20-add-source-organisation-id-as-input-argument-for-creating-agreement)

PR #605, opened 2023-01-20
CreateAgreement: page/view for the new transfer agreement creation

vahidbazzaz:
The PR creates a page/view for creating transfer agreements, and adds a global radio group and date field component as well. The PR made the following changes:

- Create a component for the date and radio groups
- Set the default color for the select field
- Fix radio and checkbox pointer events
- Remove transfer files from eslintignore to apply linter
- Create a page and route for transfer agreements
- Show an alert when initial data cannot be loaded
- Validation of the date field

Tests are added for the create transfer agreement in the follow-up PR: https://github.com/boxwise/boxtribute/pull/621


vahidbazzaz:
[2.0 CreateAgreement: create page/view](https://trello.com/c/XTHv8t56)

react/src/App.tsx
```diff
@@ -64,7 +65,7 @@ const useLoadAndSetAvailableBases = () => {
       const organisationId = decodedToken["https://www.boxtribute.com/organisation_id"];
       dispatch({
         type: "setOrganisationId",
-        payload: organisationId,
+        payload: organisationId ?? 1,
```
pylipp, line 68: I assume this is for the god user who does not have an organisation ID set. Is it possible to use 0 as a value? 1 is ambiguous (even though it's a test org).

react/src/components/Form/DateField.tsx
```diff
@@ -0,0 +1,69 @@
+import { FormControl, FormErrorMessage, FormLabel, Input, Text } from "@chakra-ui/react";
+import { Controller } from "react-hook-form";
+// import { Controller } from "react-hook-form";
```
pylipp, line 3: comment can be removed

react/src/components/Form/SelectField.tsx
```diff
@@ -67,8 +67,10 @@ function SelectField({
               multiValue: (provided, state) => ({
                 ...provided,
                 border: "1px",
-                borderColor: colorIsBright(state.data?.color) ? "gray.300" : state.data?.color,
-                color: colorIsBright(state.data?.color) ? "black" : "white",
+                borderColor: colorIsBright(state.data?.color ?? "#fff")
```
pylipp, line 70: I'm don't fully understand what this does. If `state.data.color` is undefined, you pass `#fff` to `colorIsBright`? How is `#fff` a valid color code? Also if `#fff` stands for white color, it will be assessed as bright color anyways.

vahidbazzaz: @pylipp this is actually needed as when we have multiple choice selects it assumes you set the default background color, causing issues, so as a quick fix I added the default colors (background color to grey and font color to white)

react/src/views/Transfers/CreateTransferAgreement/CreateTransferAgreementView.tsx
```diff
@@ -0,0 +1,206 @@
...
+  );
+
+  const partnerOrganisationsWithTheirBasesData = allOrgsAndTheirBases?.filter(
+    (organisation) => organisation.id !== globalPreferences.selectedOrganisationId?.toString(),
+  );
+
+  // check data for form
+  useEffect(() => {
+    if (!allFormOptions.loading) {
+      if (allOrgsAndTheirBases === undefined) {
```
pylipp, line 96: Isn't this error handling a duplication to line 184?

react/src/views/Transfers/CreateTransferAgreement/components/CreateTransferAgreement.tsx
```diff
@@ -0,0 +1,269 @@
...
+//       const dateDiff = data.validUntil.getTime() - data.validFrom.getTime();
+//       return dateDiff > 0;
+//     }
+//     return true;
+//   },
+//   {
+//     message: "Please enter greater date for the valid until",
+//     path: ["validUntil"], // path of error
+//   },
+// );
```
pylipp, line 74: Comment can probably removed?

pylipp: Validating that the second date is later than the first should be done in the back-end, too.
If you want, you can add an item to check it in the FE to the "General enhancements" trello card.

react/src/App.tsx
vahidbazzaz: @pylipp actually it ensures that only available organization are selected for god user. If it is set to 0, then the source organisation name and base information cannot be shown.

react/src/views/Transfers/CreateTransferAgreement/components/CreateTransferAgreement.tsx
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    if (validFrom) {
+      setValidUntilMinDate(
+        new Date(new Date(validFrom).setDate(new Date(validFrom).getDate() + 1))
+          .toJSON()
+          .split("T")[0],
+      );
+    } else {
+      setValue("validFrom", new Date().toISOString().substring(0, 10));
```
pylipp, line 166: Can you add a comment `Use today as default for "validFrom"` ?

react/src/views/Transfers/CreateTransferAgreement/components/CreateTransferAgreement.tsx
```diff
@@ -0,0 +1,269 @@
...
+        {
+          label: currentOrganisation?.bases[0].name,
+          value: currentOrganisation?.bases[0].id,
+        },
+      ]);
+    }
+
+    if (validFrom) {
+      setValidUntilMinDate(
+        new Date(new Date(validFrom).setDate(new Date(validFrom).getDate() + 1))
```
pylipp, line 161: Why are there three `new Date` calls? :sweat_smile:

pylipp (changes requested):
Code review done, still have to do functional review

react/src/views/Transfers/CreateTransferAgreement/components/CreateTransferAgreement.tsx
vahidbazzaz: @pylipp 😅 Let me refactor this with the `addDays` function of `date-fns`

vahidbazzaz:
> * Select partner base(s). Empty means all will be selected.

Thanks @pylipp for reviewing the PR. 

Regarding the suggested changes:
- it is not possible to change the format of the standard date input form. We will have to replace it with a custom date picker or component. However, I will try to see if there is a quick solution to the situation you mentioned. 
- Also, I think long text cannot be used as placeholder text as it will not display correctly and may cause confusion. Perhaps an Alert component can be used instead under the partner base field. @aerinsol What do you think?

vahidbazzaz:
@pylipp @HaGuesto I've applied the transfer agreement enum changes to the PR

react/src/views/Transfers/CreateTransferAgreement/CreateTransferAgreementView.tsx
```diff
@@ -100,45 +100,28 @@ function CreateTransferAgreementView() {
...
     let transferType: TransferAgreementType;
-
     switch (createTransferAgreementData.transferType) {
       case "Sending to":
-        sourceOrganisationId = globalPreferences?.selectedOrganisationId;
-        sourceBaseIds = currentOrgBaseIds;
-        targetOrganisationId = createTransferAgreementData.partnerOrganisation.value;
-        targetBaseIds = partnerBaseIds;
-        transferType = TransferAgreementType.Unidirectional;
+        transferType = TransferAgreementType.SendingTo;
```
pylipp, line 106: @vahidbazzaz it's still necessary to have the old assignment of source/target org/bases because I didn't change the API in that regard: the idea is still that one org acts as source, and one as target, instead of one being the "creating" and the other the "accepting" one. Sorry for not explicitly mentioning it in #623, I thought it was clear due to the small code change.

pylipp: yes, exactly

pylipp: As discussed, I revert my request and instead implement this logic in the BE, also updating the authorization.
In the meantime the "Receiving-From" agreements-creation will be broken though.

One final thing @vahidbazzaz can you merge master one more time (it now holds some FE refactoring changes by @HaGuesto and I want to be certain everything's still okay)

pylipp (approved):

react/src/views/Transfers/CreateTransferAgreement/CreateTransferAgreementView.tsx
pylipp: @HaGuesto @vahidbazzaz here we agreed (and we just confirmed) that the source-target-switch logic does not happen in the FE but in the BE (implemented through #636 ).
Also I'll update the naming to talk about "initiating" and "partner" org when creating an agreement.

aerinsol:
The placeholder text was supposed to be "all bases" as per the design I believe. @vahidbazzaz @pylipp

PR #622, opened 2023-01-27
Sort locations in BoxEditView, BoxCreateView and BoxView

johnsmol:
Updated frontend query with 'seq' parameter in order to sort the locations in the dropdown menus of BoxEditView, BoxCreateView and BoxView


pylipp:
[2.0 BoxView/BoxEdit: Sort location options according to seq](https://trello.com/c/yYclaWYK/1031-20-boxview-boxedit-sort-location-options-according-to-seq)

HaGuesto (changes requested):
Hey Giovanni, thanks for the PR!

I have a few change requests:
1. I believe the tests fail since you should update the mocked data with the seq field. Maybe you even want to write an additional test to check if your changes work.
2. in BoxView I do not see an implementation of .sort() and there the locations are not sorted.

PR #623, opened 2023-01-27
Introduce SendingTo and ReceivingFrom agreement types

pylipp:


pylipp:
[2.0 Validate that user who accepts agreement is not the one who created the agreement #sprint-jan-30](https://trello.com/c/YpaFhMnR/1107-20-validate-that-user-who-accepts-agreement-is-not-the-one-who-created-the-agreement-sprint-jan-30)

vahidbazzaz (approved):
Thanks @pylipp

aerinsol:
Nice @pylipp . I see that we are keeping the definition of source and target base consistent with shipment direction as a way of dealing with the fact that we're not using `creating_base` and `accepting_base`. 

Do you think it would make sense to switch to the second implementation? @vahidbazzaz  @HaGuesto

PR #634, opened 2023-01-31
align db seed of dropapp and v2 and auth0

HaGuesto:
I did the following things:
- Qr-Codes pictures in the Boxtribute repo should work
- add transfer Agreement Data to the seeds to have data to show in AgreementsOverview
- add shipment data to the seed  to have data to show in ShipmentOverview
- add users to each user group in boxcare to more easily test transfer agreement/shipment actions


HaGuesto:
[align db seed of dropapp and v2. Qr-Codes pictures in the repo are not in the db seed in the new app #sprint-jan-30](https://trello.com/c/kyxhpmHm/1067-align-db-seed-of-dropapp-and-v2-qr-codes-pictures-in-the-repo-are-not-in-the-db-seed-in-the-new-app-sprint-jan-30)

back/init.sql
```diff
@@ -1905,6 +1905,7 @@ CREATE TABLE `transfer_agreement` (
 
 LOCK TABLES `transfer_agreement` WRITE;
 /*!40000 ALTER TABLE `transfer_agreement` DISABLE KEYS */;
+INSERT INTO `transfer_agreement` VALUES (1,1,2,'UnderReview','Bidirectional','2023-01-24 07:25:25',8,NULL,NULL,NULL,NULL,'2023-01-24 07:25:25',NULL,NULL),(2,2,1,'UnderReview','Bidirectional','2023-01-24 07:27:49',8,NULL,NULL,NULL,NULL,'2023-01-24 07:27:49',NULL,NULL),(3,1,2,'UnderReview','Unidirectional','2023-01-24 07:29:24',8,NULL,NULL,NULL,NULL,'2023-01-24 07:29:24',NULL,NULL),(4,2,1,'UnderReview','Unidirectional','2023-01-24 07:30:05',8,NULL,NULL,NULL,NULL,'2023-01-24 07:30:05',NULL,NULL);
```
pylipp, line 1908: The value `Unidirectional` does not exist anymore. For ID 3 it should be `SendingTo`, for ID 4 `ReceivingFrom`

pylipp (changes requested):
In `back/test/conftest.py` is a line with `setup_box_transfer_models()` which was used to insert transfer models into the dropapp_dev database in the context of integration testing. There now is an error because the seed already contains agreements. So I think the line can be removed. But even then I expect errors in other integration tests that rely on QR codes and box label identifiers from the seed.

back/init.sql
pylipp: Can you rebase on latest master?

pylipp:
@HaGuesto Also I'd like to understand how these new values were generated and whether one must be careful in the future when creating a dump from the staging DB and using it as the seed for the dev DB.

HaGuesto:
@pylipp This is ready to review again.

back/test/integration_tests/test_permissions.py
```diff
@@ -97,15 +97,15 @@
         [
             "another_dev_coordinatorr@boxcare.org",
             BOX_CARE_ID,
-            [THESSALONIKI_ID],
-            [SAMOS_ID, ATHENS_ID],
+            [ATHENS_ID],
```
HaGuesto, line 100: @pylipp I just realized that in other PRs these tests fail, too. This fix should solve it. 
The problem is that another_dev_coordinatorr and another_dev_volunteer are associated with base Athens in the db seed, but with base Thessaloniki in Auth0. 
I fixed this and synced it.

pylipp (approved):

PR #647, opened 2023-02-10
Refactor Qr Reader

HaGuesto:
Things done:

- move handling of scanning Qr Codes and searching labelIdentifier out of the HeaderMenu and move it into QrReaderComponent
- reduce QrReaderOverlay to controlling modals.
- add hooks for resolving qr codes and labelIdentifiers
- rename all hooks to adhere to standard practise
- re-organize graphQL fragments and fix corresponding circular dependencies
- create QrReaderView and NotFoundView to prepare for installable PWA
Test this by requesting `localhost:3000/qrreader`


HaGuesto:
[2.0 Frontend: The logic to check if an QrCode exists or is connected to a Box is implemented in two places in the front-end and should be deduplicated/refactored into a Hook #sprint-feb-09](https://trello.com/c/tbLp7J7B/1101-20-frontend-the-logic-to-check-if-an-qrcode-exists-or-is-connected-to-a-box-is-implemented-in-two-places-in-the-front-end-and-sh)

react/src/hooks/useLabelIdentifierResolver.ts
```diff
@@ -0,0 +1,80 @@
...
+import { BoxDetailsQuery, BoxDetailsQueryVariables } from "types/generated/graphql";
+import { BOX_DETAILS_BY_LABEL_IDENTIFIER_QUERY } from "queries/queries";
+
+// eslint-disable-next-line no-shadow
+export enum ILabelIdentifierResolverResultKind {
+  SUCCESS = "success",
+  FAIL = "fail",
+  NOT_FOUND = "notFound",
+  NOT_AUTHORIZED = "notAuthorized",
+  // TODO: implement the following two edge cases
```
pylipp, line 12: Usual request for a Trello card :wink:

react/src/views/AutomaticBaseSwitcher/AutomaticBaseSwitcher.tsx
```diff
@@ -2,35 +2,43 @@ import { useAuth0 } from "@auth0/auth0-react";
...
   useEffect(() => {
     const bases = globalPreferences.availableBases;
-    if (bases != null) {
+    if (bases && Array.isArray(bases)) {
       if (bases.length > 0) {
         const newBaseId = bases[0].id;
-        navigate(`/bases/${newBaseId}`);
+        // try to prepend /bases/:baseId to state.origin
+        navigate(
+          state && state.origin ? `/bases/${newBaseId}${state.origin}` : `/bases/${newBaseId}`,
```
pylipp, line 24: Can I test this somehow?

react/src/App.tsx
```diff
@@ -133,9 +135,11 @@ function App() {
                 </Route>
               </Route>
             </Route>
+            <Route path="qrreader" element={<QrReaderView />} />
           </Route>
         </Route>
       </Route>
+      <Route path="/*" element={<NotFoundView />} />
```
pylipp, line 142: When using the route `/bases` only, this works (I suppose) by redirecting me to `/bases/1`. However when I go to `/bases/1/foo` I end up with `/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/bases/1/foo` so some recursion happens. Can you remove the usage of NotFoundView here because it's rather out-of-scope and fix it later?

react/src/views/AutomaticBaseSwitcher/AutomaticBaseSwitcher.tsx
HaGuesto: just request any route like /qrreader or /test in the Browser and it should prepend /bases/1

react/src/App.tsx
HaGuesto: lol. ok yes. can do

pylipp (changes requested):
### Code review
I understand most of it :sunglasses: a bit confused with the naming of QrReader and QrReaderScanner (does scanning mean getting the image, and reading means "comprehending" it, i.e. extracting the URL?).
Thanks for the big breakdown of the QrReaderOverlay(Container), and the cleanup for fragments.

### Functional review
- QR scanning works as expected for most cases except when I use [this code](https://github.com/boxwise/boxtribute/blob/refactor/QrResolver/docs/qr/code-not-in-db-f660f96618eaa81e16b7869aca8d67d.png), then I get an endless notification stream (logged in as dev_volunteer@boxaid.org)
- label finding incl validation works too
- NotFoundView has a small bug, see comment

react/src/views/AutomaticBaseSwitcher/AutomaticBaseSwitcher.tsx
pylipp: thanks, works :+1:

HaGuesto:
> I understand most of it sunglasses a bit confused with the naming of QrReader and QrReaderScanner (does scanning mean getting the image, and reading means "comprehending" it, i.e. extracting the URL?). Thanks for the big breakdown of the QrReaderOverlay(Container), and the cleanup for fragments.

@pylipp yes, you are correct.

> * QR scanning works as expected for most cases except when I use [this code](https://github.com/boxwise/boxtribute/blob/refactor/QrResolver/docs/qr/code-not-in-db-f660f96618eaa81e16b7869aca8d67d.png), then I get an endless notification stream (logged in as [dev_volunteer@boxaid.org](mailto:dev_volunteer@boxaid.org))

The problem underneath it is that the QrReaderScanner triggers the callback once it finds a QR-Code in its image. I already refactored it in a way that the QrReaderScanner can only trigger one callback at a time and has to wait for the response from the server before triggering another.

However, if it is a non Boxtibute Qr-Code the callback resolves immediately and no request is sent to the server. Therefore, the QrReaderScanner triggers error after error.

To resolve this I probably have to implement a wait period that the QrScanner can only trigger the callback every two seconds, but that feels hacky. I'm open for suggestions. @vahidbazzaz @pylipp @aerinsol

react/src/App.tsx
HaGuesto: fixed it.

pylipp:
@HaGuesto why is there no "queue" of errors being processed?

HaGuesto:
> @HaGuesto why is there no "queue" of errors being processed?

I do not 100% understand what you mean.

PR #648, opened 2023-02-10
BoxView Fix for When Legacy Lost Boxes are Found

vahidbazzaz:
This PR adds a warning to the box view when a box is located on the legacy Lost/Scrap location, and allows it to be moved to the instock location


vahidbazzaz:
[BUG: 2.0 BoxView fix for when legacy lost boxes are found](https://trello.com/c/Y20IxvvO)

aerinsol:
As far as I can tell, this issue with not being able to move LOST boxes in legacy locations also exists in Dropapp. Does this cover that case?

react/src/views/Box/BoxView.tsx
```diff
@@ -228,16 +235,40 @@ function BTBox() {
         boxLabelIdentifier: labelIdentifier,
         newState,
       },
+      // optimisticResponse: {
```
pylipp, line 238: Can you remove this and/or move it to a separate branch?

pylipp (approved):
I'm still seeing that a couple of error cases don't seem to be covered but that's not part of this PR. Thanks for implementing the alert :+1:

aerinsol:
@vahidbazzaz confirming that there's no issue in Dropapp on the backend so in theory this PR should resolve the issue.

PR #656, opened 2023-02-14
First version of ADR for database history

pylipp:
Proposal: go for history/shadow tables.

Read the ADR as rendered markdown [here](https://github.com/boxwise/boxtribute/blob/00f78b9cb3ffe065d22546cc10ca1e9e6b34b251/docs/adr/adr_database-history.md).

This ADR supersedes the previous [Google doc draft](https://docs.google.com/document/d/1eqcS7iYN0oIorl9JcvNm2eoY4_ogUFyLh-wQLD53SOQ/edit#).

### DIFF
--- a/docs/adr/adr_database-history.md
+++ b/docs/adr/adr_database-history.md
@@ -36,14 +36,14 @@ A variation is to have another table to record details for each change.
 
 *Advantages*
 
-- changes of data are recorded
+- straightforward to see information about what *changed* at any point in time
 - simple and extensible
 - can be combined with database triggers to record any change on DB level
 
 *Drawbacks*
 
-- at a later point in time, the historical state of the data cannot be clearly derived: one can see that a field of a row has changed but other fields might have changed as well
-- dealing with data type conversion when storing/reading change payload ("old"/"new" value)
+- since the audit table records changes it becomes involved to derive the state of the database at a certain point in the past
+- dealing with different data types requires additional effort when storing/reading change payload ("old"/"new" value)
 
 ### History tables (or: shadow tables)
 
@@ -52,6 +52,7 @@ For every relevant table create a history table with identical schema and an add
 *Advantages*
 
 - change date and user are recorded
+- data versions are recorded and can be easily retrieved
 - simple, especially querying (select by highest value of `effective` date column)
 - separation of data and history
 - can be combined with database triggers to record any change on DB level (no change in application code necessary then)
@@ -61,10 +62,11 @@ For every relevant table create a history table with identical schema and an add
 
 - extensibility costs: every change in the original table requires an update in the corresponding history table
 - the state of foreign key fields (i.e. tables that a history table depends on) has to be recorded as well to still be correct later
+- since shadow tables record versions of data it becomes more involved to derive the actual change (i.e. the values before/after a change)
 
 ### Insert-only tables
 
-Whenever a row would change, it is marked as inactive and copied into a new, active row.
+Whenever a row would change, it is marked as inactive and copied into a new, active row with an `effective` date set.
 
 *Advantages*
 
@@ -108,11 +110,13 @@ Every change to the state of an application is captured in an event object. Thes
 
 There are several popular audit logging techniques, but none of them serve every purpose. The most effective ones are often expensive, resource intensive, or performance degrading. Others are cheaper in terms of resources but are either incomplete, cumbersome to maintain, or require a sacrifice in design quality.
 
-I propose to use history tables. It meets the requirements for simplicity and for knowing historical state. Since we don't have a lot of tables to be shadowed (stock, products, locations, people, transactions), the maintenance cost is low.
+I propose to use history tables. It meets the requirements for simplicity and for knowing historical state. Since we don't have a lot of tables to be shadowed (stock, products, locations, people, tags, qr; maybe some for agreements/shipments), the maintenance cost is low.
 
 Quoting the first reference below,
 
 > If you only want to store logs for a few tables, shadow tables may be the most convenient option.
+
+If we want to record data *changes* rather than *versions* of data, an audit table is more suitable.
 
 ## Consequences
 
@@ -120,8 +124,8 @@ See advantages/drawbacks above.
 
 We have to think about
 
-- whether we still want `created_at`, `modified_at`, etc. fields in tables that are shadowed
-- whether to "convert" the current `history` table or to remove it
+- ~~whether we still want `created_at`, `modified_at`, etc. fields in tables that are shadowed~~ (keep for now because otherwise dropapp code needs to be modified)
+- ~~whether to "convert" the current `history` table or to remove it~~ (keep for now because otherwise dropapp code needs to be modified)
 - how to design the GraphQL API for data analysis
 
 ## References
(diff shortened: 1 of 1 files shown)


pylipp:
[2.0 Take a stab at box-flow #sprint-jan-19 #sprint-jan-30 #sprint-feb-09](https://trello.com/c/KQpZaHqC/1074-20-take-a-stab-at-box-flow-sprint-jan-19-sprint-jan-30-sprint-feb-09)

docs/adr/adr_database-history.md
```diff
@@ -0,0 +1,150 @@
...
+
+*Advantages*
+
+- changes of data are recorded
+- simple and extensible
+- can be combined with database triggers to record any change on DB level
+
+*Drawbacks*
+
+- at a later point in time, the historical state of the data cannot be clearly derived: one can see that a field of a row has changed but other fields might have changed as well
```
HaGuesto, line 45: I do not get this point.

docs/adr/adr_database-history.md
```diff
@@ -0,0 +1,150 @@
...
+There are five options with an impact on different levels of the software.
+
+### Audit table
+
+This is a record of changes, tracking change type, change time, change description, name of the table that the change was made in, ID of the changed record, and possibly a structured payload. The current `history` table implements this concept.
+A variation is to have another table to record details for each change.
+
+*Advantages*
+
+- changes of data are recorded
```
HaGuesto, line 39: why is this an advantage? I thought this might be a necessity/requirement?

docs/adr/adr_database-history.md
```diff
@@ -0,0 +1,150 @@
...
+- simple and extensible
+- can be combined with database triggers to record any change on DB level
+
+*Drawbacks*
+
+- at a later point in time, the historical state of the data cannot be clearly derived: one can see that a field of a row has changed but other fields might have changed as well
+- dealing with data type conversion when storing/reading change payload ("old"/"new" value)
```
HaGuesto, line 46: also not sure about this. I do not think we have data type conversions in the history table, but it is annoying to store different types in the same table.

docs/adr/adr_database-history.md
```diff
@@ -0,0 +1,150 @@
...
+*Drawbacks*
+
+- complex extension of codebase required (twice for dropapp and v2 back-end)
+- new developers might need to familiarize themselves with the concept first
+
+## Decision
+
+There are several popular audit logging techniques, but none of them serve every purpose. The most effective ones are often expensive, resource intensive, or performance degrading. Others are cheaper in terms of resources but are either incomplete, cumbersome to maintain, or require a sacrifice in design quality.
+
+I propose to use history tables. It meets the requirements for simplicity and for knowing historical state. Since we don't have a lot of tables to be shadowed (stock, products, locations, people, transactions), the maintenance cost is low.
```
HaGuesto, line 111: I do not think we have to shadow transactions. It is already a "transactional" db table and not a CRUD table. We can figure out the distributed items in the free shop just by counting/summing the rows between two dates. 
One can delete transactions though.

docs/adr/adr_database-history.md
```diff
@@ -0,0 +1,150 @@
...
+- history tables should be insert-only; they cannot be modified then
+
+*Drawbacks*
+
+- extensibility costs: every change in the original table requires an update in the corresponding history table
+- the state of foreign key fields (i.e. tables that a history table depends on) has to be recorded as well to still be correct later
+
+### Insert-only tables
+
+Whenever a row would change, it is marked as inactive and copied into a new, active row.
```
HaGuesto, line 67: I guess we would also need to add an `effective` column in the related tables, right?

HaGuesto (commented):
Hey @pylipp, well done.

Here, my first thoughts:
- In the best case we should implement something that we can do on the db level so that we do not have to touch both the php and the v2 app. --> Event sourcing is out for me.
- Snapshots feel too much data to be stored and too much of a cost while being inaccurate. --> for me out, too.
- Insert-only tables feel scary to implement due to foreign key and we would probably have to touch php to change the db commands or can you change how an `UPDATE` call is rendered in the db? --> Since I get stomach aches reading this option, I would not do it.
- History (Shadow Tables) and an Audit Table seem both feasible solutions.

docs/adr/adr_database-history.md
HaGuesto: I went through the db. In the best case we should implement shadow tables for all db tables that the user can edit and especially the CRUD ones.
This means:

- locations
- people
- products
- qr (since it is INSERT only, we do not need to shadow it, but a created_by column would be nice here)
- stock 
- tags (this is the only N-to-N relationship, which probably means we have to shadow tags_relations, too)
- transactions (since it is INSERT DELETE only, we do not need to shadow it)

not sure what to do about:
- distro_events
- transfer_agreements
- shipments

Stuff we do not want to mirror:
- anything connected to borrow_, laundry_, library_ because we discontinue support for these features.
- Anything user related.

docs/adr/adr_database-history.md
pylipp: you're right. I meant to express that changes are first-level objects in this concept, whereas with shadow tables, "versions" of a row are first-level, and in order to find out the underlying change, you'd have to compare to consecutive versions.

So if it's most important to know about what _changed_ at any point in time, the audit table provides a direct solution.

If it's most important to know about a _version_ of the data at any point in time, shadow tables provide a direct solution.

However with some extra work, it's possible to derive the respective opposite in each concept.

docs/adr/adr_database-history.md
pylipp: okay, I should have put it like this: dealing with different data types requires additional effort

docs/adr/adr_database-history.md
pylipp: yep

docs/adr/adr_database-history.md
```diff
@@ -0,0 +1,150 @@
...
+
+> If you only want to store logs for a few tables, shadow tables may be the most convenient option.
+
+## Consequences
+
+See advantages/drawbacks above.
+
+We have to think about
+
+- whether we still want `created_at`, `modified_at`, etc. fields in tables that are shadowed
```
HaGuesto, line 123: If we remove them, we might have to change code in the php app. Therefore, I would keep them and remove them once we are rid of the php app :).

docs/adr/adr_database-history.md
```diff
@@ -0,0 +1,150 @@
...
+- whether to "convert" the current `history` table or to remove it
```
HaGuesto, line 124: same as for `created_at`, `modified_at`

docs/adr/adr_database-history.md
pylipp: Yeah I struggled to find a clear explanation indeed.
I explained in a different reply that it's harder to obtain a version of the data at a certain point in time using an audit table.
E.g. if you recorded a change in number-of-items for a box, you do not know right away the product that was in the box.

HaGuesto:
@pylipp I suggest as a next step that we create a first history(shadow) table for stock with a few entries (moved boxes, changed product/size, change in number of items) and we write a query 
- to sum up the outflow of stock between two time points and 
(- to show a stock summary at a current point in time)

If we still want to consider audit tables, we can do the same queries based on the history table.

aerinsol:
I have a preference for history tables rather than audit tables. Audit tables, as we've seen now, appear to be more difficult to work with compared to history tables for feature adjustments and overall logging.

aerinsol (approved):

PR #684, opened 2023-03-02
Updating Packages and Resolving Dependencies Vulnerabilities

vahidbazzaz:
This PR updates some dependencies to the latest version as follows: 

- upgrade json5 to the fixed version 2.2.3
- update packages 
- update json5 to fixed version 2.2.3 
- update msw 
- update storybook 
- update craco 
- update graphql-codegen 
- update prettier 
- update regenerator-runtime
- update glob-parent to fixed version 5.1.2
- update trim to fixed version 0.0.3
- update trim-newline to fixed version 3.0.1
- update nth-check to fixed version 2.0.1


react/yarn.lock
```diff
@@ -15735,7 +15582,7 @@ remark-parse@8.0.3:
     parse-entities "^2.0.0"
     repeat-string "^1.5.4"
     state-toggle "^1.0.0"
-    trim "0.0.1"
+    trim "0.0.3"
```
aerinsol, line 15585: @vahidbazzaz isn't the first fixed version 3.0.1, not 0.0.3?

vahidbazzaz: @aerinsol  It appears from the [dependent bot message](https://github.com/boxwise/boxtribute/security/dependabot/36) that the first fixed version of `trim` is `0.0.3`. I think you are referring to `trim-newline`, which I corrected by upgrading dependent packages so that they already included the patch and as you mentioned, the fixed version of `trim-newline` is `3.0.1`.

aerinsol (approved):

react/src/index.tsx
```diff
@@ -14,8 +14,7 @@ import { theme } from "./utils/theme";
 
 if (process.env.NODE_ENV === "development") {
```
pylipp, line 15: Can you delete this entire block? It's stale code and the GraphQL mock query might not even work

react/src/mocks/browser.ts
```diff
@@ -1,5 +1,7 @@
+/* eslint-disable no-unused-vars */
```
pylipp, line 1: Please delete the comments, imports and the lines that are offensive for eslint, we can pull that from git history.

pylipp (changes requested):
thanks for the updates @vahidbazzaz can you run the FE tests locally ones to confirm everythings okay (because CI is flaky...)

react/src/mocks/browser.ts
vahidbazzaz: @pylipp I should then delete the file since we no longer use any of the embedded libraries

vahidbazzaz:
> thanks for the updates @vahidbazzaz can you run the FE tests locally ones to confirm everythings okay (because CI is flaky...)
I have no problems running tests locally

pylipp (approved):

PR #691, opened 2023-03-10
Bug: Fix Box-field Fragment for Shipment Overview Query

vahidbazzaz:
This PR fixes an authorization error when querying the box location from the target organization's shipment overview page.


vahidbazzaz:
[2.0 Bug: fix box-field fragment for shipment overview query](https://trello.com/c/QSOficjd)

HaGuesto (approved):
Works as it should @vahidbazzaz

react/src/queries/fragments.ts
```diff
@@ -137,6 +137,32 @@ export const BOX_FIELDS_FRAGMENT = gql`
...
+export const BOX_WITHOUT_LOCATION_FIELDS_FRAGMENT = gql`
+  ${PRODUCT_BASIC_FIELDS_FRAGMENT}
+  ${SIZE_FIELDS_FRAGMENT}
+  ${TAG_FIELDS_FRAGMENT}
+  ${BASE_BASIC_FIELDS_FRAGMENT}
+  ${HISTORY_FIELDS_FRAGMENT}
+  fragment BoxWithoutLocationFields on Box {
+    labelIdentifier
+    state
+    product {
```
pylipp, line 149: FYI @HaGuesto @vahidbazzaz seeing this fragment I was surprised that the query previously failed with insufficient permission for the location field but not for the product and tags fields that are both base-specific, too.
I checked the resolvers in `.../box/fields.py` and the difference is that the resolvers for tags and product use a loader. The loader works by collecting all requested IDs, and then loads the rows in a single batch but also will filter for rows (i.e. tags or product) that the user is authorized for. If an ID for a non-permitted resource was collected, the corresponding returned value and hence the GraphQL field will be None (so instead of failing with a Forbidden exception, the tags/product field resolver silently returns None).
However the location resolver has the `authorize()` call which noisely fails the entire request if the user has no access to the location's base.
_Why_ does the location resolver not work the same way as the others? Probably because I was too lazy to align it. If the location resolver would also use a loader, then the original issue would not have appeared. I'm considering updating the BE in that regard and removing the `BOX_WITHOUT_LOCATION` fragment again.

aerinsol: Agree with @pylipp's solution, it seems more durable and less hacky. I'm going to request a fix here prior to production release @vahidbazzaz

PR #704, opened 2023-03-17
Fix/Refactor how we use the baseId from the url throughout the app

HaGuesto:
### What I did

- moved `useLoadAndSetAvailableBases` from Apps.tsx to a separate hook named `useLoadAndSetGlobalPreferences`
- added selectedBaseId to the GlobalPreference Provider and set it in the new `useLoadAndSetGlobalPreferences` hook. I did this to validate the baseId provided in the url against `availableBases`. It is also needed to implement base switching.
- I replaced all instances where the baseId is directly read from the url by `gloabelPreferences.selectedBaseId`. This is to prevent any possibility for injection of the baseId in some way. @vahidbazzaz I recommend to do the same in the future if u need to query the current baseId a user is in.
- I refactored how prepending of `bases/<baseId>` to the url and I ripped out the hacky AutomaticBaseSwitcher.
- I resolved the issue of queries without authorization headers by implementing an error and isLoading state in App.tsx. Now the app components are only rendered if the globalPreferences are properly provided.
- I also had to create a context/provider for the last point in ApolloAuth0Provider.tsx


HaGuesto:
@pylipp This is ready for a code review. I will fix the failing test, when the flaky tests are resolved on master.

react/src/hooks/useLoadAndSetGlobalPreferences.ts
```diff
@@ -0,0 +1,94 @@
...
+import { GlobalPreferencesContext } from "providers/GlobalPreferencesProvider";
+import { ApolloAuth0WrapperContext } from "providers/ApolloAuth0Provider";
+import { BasesQuery } from "types/generated/graphql";
+
+export const useLoadAndSetGlobalPreferences = () => {
+  const { user } = useAuth0();
+  const navigate = useNavigate();
+  const location = useLocation();
+  const { globalPreferences, dispatch } = useContext(GlobalPreferencesContext);
+  const { isAccessTokenInHeader } = useContext(ApolloAuth0WrapperContext);
```
pylipp, line 14: Is this the key to the 401s due to missing tokens in the header that we saw?

pylipp, line 14: Is this the key to the 401s due to missing tokens in the header that we saw?

pylipp: okay yep apparently: https://github.com/boxwise/boxtribute/pull/704/files#diff-dff76fc64a9d7628c33c39e139bc4fd76c3f60a138b28ba14d1a23d4f0c98802R101

react/src/providers/ApolloAuth0Provider.tsx
```diff
@@ -81,14 +89,25 @@ function ApolloAuth0Provider({ children }: { children: ReactNode }) {
 
   const client = new ApolloClient({
     cache,
-    // HINT: Ideally, only set this temporary to true for local debugging
+    // TODO: Ideally, only set this temporary to true for local debugging
```
pylipp, line 92: Is it possible to do `connectToDevTools: process.env.NODE_ENV === "development",` ?

pylipp (approved):
### Functional review
I logged in as dev_volunteer@boxaid.org:
- no 401s when loading the page
- route redirecting works when I omit `bases/1` from the URL
- I see an error message when I manipulate the URL to go to `bases/2` and the logout button works

I logged in as dev_headofops@boxcare.org
- I'm directed to `/bases/2` as default
- I can go to `bases/3` without errors
- manipulating the URL as `bases/1` gives an error message

### Code review
Solution with new hook looks clean, just one small remark.

react/src/providers/ApolloAuth0Provider.tsx
HaGuesto: fixed it

docker-compose.yml
```diff
@@ -1,81 +1,81 @@
...
+      MYSQL_DB: dropapp_dev
+      MYSQL_PORT: 3306
+      ENVIRONMENT: ${ENVIRONMENT:-development}
+      CI: "true"
+  react:
+    build:
+      context: ./react
+    environment:
+      NODE_ENV: development
+      REACT_APP_SENTRY_ENVIRONMENT: ${ENVIRONMENT:-development}
```
pylipp, line 38: Why is `REACT_APP_ENVIRONMENT` not set here?

pylipp:
thanks for cleaning up the env variables. I assume you set `ENVIRONMENT` in the circleci contexts to staging/demo/production. :sparkles:

PR #715, opened 2023-03-20
Use correct name for shipment ID variable

pylipp:
Otherwise the following error happens:
"Variable '$id' of required type 'ID!' was not provided."


vahidbazzaz (approved):
Thanks Philipp for fixing this!

PR #716, opened 2023-03-21
BoxEditView: reset box's product if it is soft deleted

serchtul:
When we edit a Box, we now check if the currently assigned product has been soft-deleted. If it has we reset the Product input, so the user will need to re-input a new (valid) product.


back/boxtribute_server/graph_ql/types.graphql
```diff
@@ -86,6 +86,7 @@ type Product {
   createdOn: Datetime
   lastModifiedBy: User
   lastModifiedOn: Datetime
+  deleted: Datetime
```
serchtul, line 89: I modified this file manually. I went through the documentation but I wasn't sure whether this was meant to be manually edited, or whether this is auto-generated. Can you help me with some insights on this? @pylipp

serchtul, line 89: Also, I was thinking of just having a boolean `isDeleted` field via a resolver, instead of making the whole Datetime available to the final user. Would that be a better course of action for this PR? What do you think?

back/test/endpoint_tests/test_product.py
```diff
@@ -30,6 +31,7 @@ def test_product_query(read_only_client, default_product, default_size, another_
         "price": default_product["price"],
         "gender": "Women",
         "createdBy": {"id": str(default_product["created_by"])},
+        "deleted": None,
```
serchtul, line 34: Should it be worth adding a test case for a soft-deleted product somewhere?

react/src/views/BoxEdit/components/BoxEdit.tsx
```diff
@@ -127,6 +127,10 @@ function BoxEdit({
     tags: boxData?.tags || [],
   };
 
+  if (boxData?.product?.deleted !== null) {
```
serchtul, line 130: I wasn't sure where to add typing information about this new property. I see these types get pulled directly from the generated graphql.ts file, but that file auto-generates it as `any`.

Is there a way to be more specific about the types? @vahidbazzaz

serchtul (commented):
Here are some preliminary thoughts on this PR. This works locally but I still need to update some FE tests to account for the changes I made.

back/boxtribute_server/graph_ql/types.graphql
pylipp: Hej @serchtul , this file is part of the definition of our GraphQL API and is manually edited.
By adding this field it's available through the API, you did the correct thing :+1: 
On the FE a TypeScript type definition file is derived by `yarn generate-graphql-ts-types` but I can see you already figured that out. :slightly_smiling_face:

back/test/endpoint_tests/test_product.py
pylipp: Sure, there's already a deleted product in the test data (`back/test/data/product.py`). In the first test in the current file you can add another request to query the deleted product and it's `deletedOn` field. :muscle:

aerinsol:
Quick comment from me: we should probably also add some sort of user notification informing them what is happening. I will try and add the notification requirement to our test cases at some later date.

back/boxtribute_server/graph_ql/types.graphql
serchtul: Done!

serchtul (commented):
I think I am done with this functionality, I had some issues testing some changes I made into the backend locally, so it took me way longer than I originally expected.

Let me know if there are any questions or concerns.

serchtul:
[2.0 BoxEdit: if the current product is soft-deleted, the input should be left empty such that the user is forced to select a non-deleted product (same behavior as dropapp)](https://trello.com/c/uKCV85uc/1167-20-boxedit-if-the-current-product-is-soft-deleted-the-input-should-be-left-empty-such-that-the-user-is-forced-to-select-a-non-de)

react/src/views/BoxEdit/components/BoxEdit.tsx
```diff
@@ -127,6 +129,16 @@ function BoxEdit({
     tags: boxData?.tags || [],
   };
 
+  if (boxData?.product?.deletedOn !== null) {
```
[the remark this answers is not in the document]
aerinsol: Great thought @serchtul. I'm not sure if I mentioned we do TDD for all non-proof of concept PRs but I'm glad it's at the front of your mind in any case! You can log your test cases [here](https://docs.google.com/spreadsheets/d/1sDhSsaVwNxAhGn1VYACDPVepZpvlwTkgy9nmeTaRy2Q/edit?usp=sharing) and use the 3.1 series of tests as "what good looks like" examples. I've tagged you on the relevant cell. 

As for the dropdowns, IIRC we actually use a non-Chakra implementation. I'm sure we logged which library we use for both the dropdown and the testing of it but I can't seem to find it in Slack ... @vahidbazzaz @HaGuesto ?

react/src/views/BoxEdit/components/BoxEdit.tsx
```diff
@@ -127,6 +129,16 @@ function BoxEdit({
...
+    defaultValues.productId = { label: "", value: "" };
+    if (!toastIsActive!()) {
+      createToast({
+        message: "The product assigned to this box no longer exists. Please select another one.",
+        type: "warning",
+      });
+    }
```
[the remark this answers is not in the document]
aerinsol: Requesting two changes: 
1. I would actually use a `FormErrorMessage` here rather than a toast, if possible. https://v1.chakra-ui.com/docs/components/form/form-control
2. Change the form error message to: "$deletedproductname has been deleted as a product type. Please select a different product."

NB: If it's not possible to do a `FormErrorMessage`, I'd like to change the toast type from `warning `to `info`. We try to avoid using the warning type for UX reasons.

aerinsol (changes requested):
See comments on testing and error messages

pylipp (commented):
On the BE side it looks solid, thank you :+1: 

For the FE side I can't tell (I remember the repeated rendering of toasts has been plaguing us) and I'll leave that to @HaGuesto

react/src/hooks/useNotification.ts
```diff
@@ -2,7 +2,7 @@ import { ToastPositionWithLogical, useMediaQuery, useToast } from "@chakra-ui/re
 import { useState, useEffect, useCallback } from "react";
 import { INotificationProps } from "./hooks";
 
-export const useNotification = () => {
+export const useNotification = (toastName?: string) => {
```
HaGuesto, line 5: @serchtul even if we switch the notification to the input field, I would definitely let this stay in. Quite a nice solution to check if a notification is active.

react/src/views/BoxEdit/components/BoxEdit.tsx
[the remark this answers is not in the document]
HaGuesto: So, actually looking into this, I realize that you are directly touching a bit of a sensitive spot here :)

For the select field in the form we use the following stack:
- chakra-react-select for the input field itself with the search capability (u can find the definition of the SelectField in /components/Form). There might be also a helpful link in this file.
- zod for the validation of the form
- react-hook-form for the control over the form

react/src/views/BoxEdit/components/BoxEdit.tsx
```diff
@@ -127,6 +129,16 @@ function BoxEdit({
     tags: boxData?.tags || [],
   };
 
+  if (boxData?.product?.deletedOn !== null) {
+    defaultValues.productId = { label: "", value: "" };
```
HaGuesto, line 133: @serchtul This should probably be null or undefined to flush the value in the product field properly. I can imagine that this will cause a type error based on the type definition through zod. Try resolving this type error and we can maybe talk about it tomorrow.

HaGuesto: in line 65 you have more information on the problem there.

react/src/views/BoxEdit/components/BoxEdit.tsx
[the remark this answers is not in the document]
HaGuesto: @serchtul To make it a FormErrorMessage you will either have to adjust the zod validation in line 65 or you create a separate error state just for this and pass it as an error besides the errors from react-hook-form to the SelectField in line 228.

I can imagine that it might be a bit of a tough nut and do not waste to much on it. If we do not find a solution, stay with the toast.

HaGuesto (commented):
@serchtul There is a few things you can try out, but I'm happy if you change the default productId to null and leave the toast.

Let me know if you need help.

react/src/views/BoxEdit/components/BoxEdit.tsx
aerinsol: @HaGuesto

aerinsol:
Can we merge this? @vahidbazzaz ? @HaGuesto ?

HaGuesto:
@aerinsol I fixed it and will merge now.

Here, a screencast of the current state:
https://www.loom.com/share/8080659800f844669e6f113de679ddb1?sid=9d9a01ee-ef0e-4426-8e19-3d37856402ee

PR #717, opened 2023-03-24
Fix/boxcreate updates qr in cache

HaGuesto:
I found a small bug and fixed it here:

1. Scan a qr unassociated to a box in MultiBoxTab of QrReader
2. Scan the same qr in SingleBoxTab of QrReader
3. create Box
4. Scan the same qr in MultiBoxTab of QrReader

The error message that no Box associated to this QrCode still appears since the cache for the QrCode is not updated after the Box is created.


pylipp (approved):

HaGuesto:
[2.0 BoxCreate: Add box to cache after Box was created](https://trello.com/c/ZfQ659cu/1217-20-boxcreate-add-box-to-cache-after-box-was-created)

PR #720, opened 2023-03-25
QrReaderMultiBox: Assign Boxes To Shipment

HaGuesto:
The specs are clearly defined in the Tests 3.4.5 in the [Test Plan](https://docs.google.com/spreadsheets/d/1sDhSsaVwNxAhGn1VYACDPVepZpvlwTkgy9nmeTaRy2Q/edit?usp=sharing)

Ready for functional and code review 

### Missing:
- tests


HaGuesto:
[2.0 QrOverlayMultiTab: create tests + view](https://trello.com/c/PvoAqpuF/1136-20-qroverlaymultitab-create-tests-view)

pylipp (commented):
### Functional review
@HaGuesto The following things worked:
- in /qrreader multi-box I scanned a QR code without associated, a QR code with a box from a non-authorized base, and a Lost box; the correct respective alerts were shown
- in /qrreader multi-box I scan a QR code with an InStock box associated. I can select a shipment and click "Assign All" and I'm forwarded to the Shipment View (see below for an issue that builds onto this use case)
- I've scanned one InStock box but killed the BE docker-compose service. On clicking 'Assign All' is the the expected network error message
- I scanned one InStock box but before assigning it to a shipment I manually sent the shipment via playground already. When clicking "Assign All" I see the message that the shipment is not in preparing state.
- I scanned two InStock boxes but before assigning them to a shiment I manually set one box's state to Lost via playground. I see a success message that one box could be added, and a hint that adding one box failed. I can click to remove the failed box from the list, and the counter resets to 0
- I scanned one box but before clicking "assign all" I had tampered the resolver to raise an authz error. When clicking "assign all" I see the correct error message about missing permissions.
- sceletons instead of the radio buttons are shown when I add a delay into the shipments query resolver
- I let the shipments query resolver raise an error, then I see the note the shipments query data could not be fetched
- I let the shipments query resolver return an empty list. The assign-shipment radio button and assign-all button are not shown.
- I create a second shipment in Preparing state. Now the dropdown shows two shipments

### Issues
- when I use my browser navigation to go back from the /shipments/1 page to /qrreader (after having assinged boxed), the box assigned to the shipment is still in the list but now has MarkedForShipment state, hence the message about the incorrect state is shown. This is fixed by a page refresh (box list then has 0 elements)

Thanks for providing such detailed test cases. I could use them well to verify the correct behavior :slightly_smiling_face:

pylipp (approved):

PR #753, opened 2023-04-26
Make error detection for re-used QR code less strict

pylipp:
Since the last prod deploy, the MySQL error message for when a used QR
code is added to a box is along 

    Duplicate entry '153659' for key 'qr_id_unique'

but the error detection was looking for an exact match

    Duplicate entry '153659' for key 'stock.qr_id_unique'

and, not finding a match, would leave this error unhandled.
A couple of occurrences have been reported by Sentry recently.

The error detection is now made less strict since we only have to
distinguish between an assignment of a duplicated QR code, a duplicated
box label identifier, or another unknown error.

This fix is important because the detection of a used QR code would
trigger an unknown error without telling the user what went wrong
(hence also the uncaught exceptions in Sentry). Now they're correctly
notified again about the QR code already being in use.

### DIFF
--- a/back/boxtribute_server/business_logic/warehouse/box/crud.py
+++ b/back/boxtribute_server/business_logic/warehouse/box/crud.py
@@ -82,11 +82,12 @@ def create_box(
             # peewee throws the same exception for different constraint violations.
             # E.g. failing "NOT NULL" constraint shall be directly reported
             prefix = "Duplicate entry"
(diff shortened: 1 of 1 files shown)


aerinsol (approved):

pylipp:
[2.0 Unknown error shown instead of "QR code already in use"](https://trello.com/c/4MfM1PxC/1197-20-unknown-error-shown-instead-of-qr-code-already-in-use)

PR #771, opened 2023-05-11
Fix Logout Issue for V2 that Doesn't Logout DropApp

vahidbazzaz:
This PR fixes the logout issue for V2 apps that don't log out DropApp.

This [PR](https://github.com/boxwise/dropapp/pull/548) must be merged and deployed before this one


vahidbazzaz:
[IMPORTANT BUG: Logout from V2 does not create a corresponding logout from Dropapp](https://trello.com/c/sbIJYHFF)

PR #789, opened 2023-05-25
Box Reconciliation: Create Overlay UI component

vahidbazzaz:
PR adds global component for box reconciliation based on mock Figma.


HaGuesto:
[Enable Tag filtering dropdown for Stock Overview](https://trello.com/c/E7zjZ5tD/1118-enable-tag-filtering-dropdown-for-stock-overview)

react/src/components/BoxReconciliationOverlay/BoxReconciliationOverlay.tsx
```diff
@@ -0,0 +1,232 @@
...
+            if (mutationResult?.errors) {
+              triggerError({
+                message: "Error: Could not change state of the box.",
+              });
+            } else {
+              onOverlayClose();
+              createToast({
+                title: `Box ${labelIdentifier}`,
+                type: "success",
+                message: "Boxe marked as undelivered",
```
pylipp, line 143: typo: Box

pylipp (changes requested):
## Functional review
### working as expected
- I can reconcile a box by selecting product&gender, number of items and location. After clicking "Save and move instock" I checked BoxView and the box is indeed instock.

### not working as expected
- When I've marked a box as lost, it still shows in the list, and the count does not update
- closing the BoxReconciliation overlay when I click "Save and Move Instock"

pylipp:
@vahidbazzaz as a said above, there's the following issue:
1. log in as dev_coordinator@boxaid.org
2. go to shipment 5. I see `Remaining 23/26`
3. I click the first box in the list and click the bin
**Actual behaviour**: the box is still in the list and the counter shows 23/26
**Expected behaviour**:the box is removed from the list and the counter shows 22/26 (even if I refresh the page the counter does not update)

and another one in the same Receiving-Shipment view:
1. I click a box for reconciliation
2. I select all fields in valid way and click `Save and move Instock`
**Acutal behaviour": the overlay does not close. When closing it manually, the counter shows 23/26
**Expected behaviour": the overlay auto-closes. The counter shows 22/26 (even if I refresh the page the counter does not update)

So I think there might be an issue with the way the counter is derived.
Maybe @aerinsol can comment on whether lost/reconciled boxes should still be in the list; and whether the overlay should auto-close after successful reconciliation.

aerinsol:
Yes, auto-close should happen once the reconciliation  / marking as lost is successful. Counter should absolutely update in line with **Expected Behavior** which was correctly described by @pylipp

vahidbazzaz:
Thank you @pylipp and @aerinsol for your comments, I should mention that this PR does not fully implement the reconciliation backend, and we still need to work on a few things as part of the next PR. Regarding the issue you mentioned, I also noticed that the first attempt does not work and that when you click on the button again, the overlay closes and the box is updated. I will make the necessary changes to the PR.

react/src/components/BoxReconciliationOverlay/BoxReconciliationOverlay.tsx
```diff
@@ -0,0 +1,233 @@
...
+
+  const { loading, error, data } = useQuery<
+    ShipmentByIdWithProductsAndLocationsQuery,
+    ShipmentByIdWithProductsAndLocationsQueryVariables
+  >(SHIPMENT_BY_ID_WITH_PRODUCTS_AND_LOCATIONS_QUERY, {
+    variables: {
+      shipmentId: boxReconciliationOverlayState.shipmentId || "",
+      baseId: baseId || "",
+    },
+    skip: !boxReconciliationOverlayState.shipmentId || !baseId,
```
HaGuesto, line 66: @vahidbazzaz why are you checking for this? Is this a possible scenario that the BoxReconciliation is called without shipmentId or baseId?
If so, we should probably do some error handling for this case. It should not be possible to call the overlay without the two values

vahidbazzaz: @HaGuesto yes indeed, as I used a reactive variable as soon as the page overly renders it will trigger the query. I used `skip` to only run the query when both `shipmentId` and 'baseId' are set. But I agree we should add more error handling and this is what we are planning to work on for the next PR which is more focus on the backend implementation and validation of reconciliation form.

react/src/components/BoxReconciliationOverlay/components/BoxReconciliationAccordion.tsx
```diff
@@ -0,0 +1,106 @@
...
+            onLocationSpecified={setLocationSpecified}
+            allLocations={allLocations}
+            onSubmitReceiveLocationForm={(receiveLocationFormData: IReceiveLocationFormData) => {
+              setLocationSpecified(true);
+              setAccordionIndex(-1);
+              boxReconciliationLocationFormDataVar({
+                locationId: parseInt(receiveLocationFormData.locationId.value, 10),
+              });
+
+              onBoxDelivered(shipmentDetail.box.labelIdentifier);
```
HaGuesto, line 99: This line is only called for the onSubmitReceiveLocation.

What happens if someone fills out the location form first?
Probably either the submit button should be disabled if the product is not matched yet or this callback should be added in the Product Form submit

react/src/components/BoxReconciliationOverlay/BoxReconciliationOverlay.tsx
```diff
@@ -0,0 +1,233 @@
...
+            if (mutationResult?.errors) {
+              triggerError({
+                message: "Error: Could not change state of the box.",
+              });
+            } else {
+              const locationName = allLocations?.find(
+                (location) =>
+                  location.id === boxReconciliationLocationFormState?.locationId?.toString(),
+              )?.name;
+              onOverlayClose();
```
HaGuesto, line 193: mmh this should actually close the modal. not sure why it is not?

HaGuesto (approved):
in general, the code review passes. There are some error handling cases that need to be fixed later.
In general, the use of ReactiveVar is atm not set up in the correct to cache the BoxReconciliation FormData.

You are using the ReactiveVar as a State for the form, but caching would mean that you save the selected product and location for the boxes in the cache even if the modal is closed inbetween until the form is successfully submitted

HaGuesto:
@pylipp The problems in this PR is fixed and ready to merge

@Vahid can u resolve the problems with the tests, please? Please also merge then directly afterwards

@aerinsol once this is merged you can start reviewing again.

react/src/components/BoxReconciliationOverlay/BoxReconciliationOverlay.tsx
```diff
@@ -0,0 +1,214 @@
...
+    ShipmentByIdWithProductsAndLocationsQueryVariables
+  >(SHIPMENT_BY_ID_WITH_PRODUCTS_AND_LOCATIONS_QUERY, {
+    variables: {
+      shipmentId: boxReconciliationOverlayState.shipmentId || "",
+      baseId: baseId || "",
+    },
+    skip: !boxReconciliationOverlayState.shipmentId || !baseId,
+  });
+
+  const [updateShipmentWhenReceiving, updateShipmentWhenReceivingStatus] = useMutation<
```
HaGuesto, line 54: @vahidbazzaz In this mutation u either need to update the cache after the mutation or run a refetchquery. Otherwise, the list and counter in ReceiveShipment does not update.

vahidbazzaz: @HaGuesto Since it already contains the required information, it does not need to be changed

PR #830, opened 2023-06-30
BoxView: Fix Tests and Update GraphQL Error Policy for BoxView Query

vahidbazzaz:
- fix boxview tests - add cache
- add support for partial response to boxview


vahidbazzaz:
[2.0 BUG: There are some BoxView tests that fail due to cache, and BoxView fails to load due to partial GraphQL response](https://trello.com/c/7YyWlrlo/1233-20-bug-there-are-some-boxview-tests-that-fail-due-to-cache-and-boxview-fails-to-load-due-to-partial-graphql-response)

pylipp:
Thanks for the fixes :raised_hands: @vahidbazzaz

PR #855, opened 2023-07-21
Add BoxReconciliation Overlay to BoxView

vahidbazzaz:
- add to optional props to box reconcilation to set closeOnOverlayClick and closeOnEsc
- add sourceLocation to shipment detail of box field fragment
- displays box reconciliation overlay if box sate equals to receiving
- disable remove shipment button when box state equals to receiving


vahidbazzaz:
[2.0 Add BoxReconciliation Overlay to BoxView](https://trello.com/c/MQMkHsGE/1226-20-add-boxreconciliation-overlay-to-boxview)

vahidbazzaz:
@HaGuesto @pylipp  I am merging this PR to master. I added one case and it increased the code coverage, but I absolutely want to add more cases in the following PR. Please review it

react/src/views/Box/BoxView.tsx
```diff
@@ -200,6 +201,22 @@ function BTBox() {
 
   const boxData = allData.data?.box;
 
+  useEffect(() => {
+    setCurrentState(boxData?.state);
+    const shipmentId = boxData?.shipmentDetail?.shipment.id;
+    // open reconciliation overlay if the box state is receiving
+    if (shipmentId && boxData?.state === BoxState.Receiving) {
```
HaGuesto, line 208: I think here should also be a check if you are currently in the target base. Otherwise, the reconciliation ui also can be added by the sending org or maybe it is reconciled to a wrong base by a user who has access to multiple bases.

I will post about this on slack and start a discussion

react/src/views/Box/components/BoxCard.tsx
```diff
@@ -62,6 +62,27 @@ function BoxCard({
     return color;
   };
 
+  const product =
```
HaGuesto, line 65: can you please move this to BoxView. Data prep should not happen in the ui component, but in the container component, e.g. where you also infered the location from the shipmentDetail

HaGuesto (commented):
looks good, just a few small change requests to be consistent where we put which part of the code.

The bug that the reconciliation overlay opens for a user from the sending org is an edge case, but should be fixed.

Probably, based on the discussion in slack, we will need to do a small iteration here.

https://boxwise.slack.com/archives/CB1UYDGLU/p1690555540392109

one thing regarding the test plan: I think u defined the tests twice. Can u remove 3.1.11

PR #856, opened 2023-07-22
Feature/move multi box

HaGuesto:
This PR enables the moving of multiple Boxes at once with the Boxtribute Qr-Reader.

[please check tests 3.4.6 as reference](https://docs.google.com/spreadsheets/d/1sDhSsaVwNxAhGn1VYACDPVepZpvlwTkgy9nmeTaRy2Q/edit?usp=sharing)

@pylipp I have not validated if a Box which was scanned is already in the location that you want to move it to. What happens in the backend if a send a graphQL mutation with updateBox and the new location id is what the box already has?


HaGuesto:
[2.0 QrOverlayMultiTab: view for move Box action](https://trello.com/c/ShhAsfd4/1178-20-qroverlaymultitab-view-for-move-box-action)

pylipp:
Functional review: works :+1: (I did not test the error cases)

> @pylipp I have not validated if a Box which was scanned is already in the location that you want to move it to. What happens in the backend if a send a graphQL mutation with updateBox and the new location id is what the box already has?

I tested this case and I can see the notification (a box was successfully moved). The BE executes the change on DB level but won't write an entry to the history table.

pylipp (approved):

HaGuesto:
@aerinsol Do you think that we should handle this case or are you happy?

PR #867, opened 2023-07-24
Add dedicated fcuntion for user-mocking in tests

pylipp:
Hide away the use of the jose library


pylipp:
[2.0 Add dedicated function for user-mocking in tests](https://trello.com/c/4MPrkHf4/1238-20-add-dedicated-function-for-user-mocking-in-tests)

HaGuesto (commented):
The tests are passing. I do not want to dig deeper. I'm happy wiht it.

PR #872, opened 2023-07-28
Add Breadcrumbs and menus transfer views

HaGuesto:
Included in this PR:

- I had to extend the authentication query for bases and organisations to also include their names. Therefore, I had to change 37 files.
- breadcrumbs in all transfer related views.
- transfer related menus without checking of ABP: I left the authorization function out since the menu should be refactored before adding this.

[I summarised what I think we should do next with the menu here](https://trello.com/c/m0JT1wpo)


HaGuesto:
image
I do not know why the codecov check is failing, but on the website the codecov increased by .11% Therfore, I am going to merge this now

pylipp:
Looks nice @HaGuesto however nothing happens when I click the breadcrumbs, not sure if that's intentional

PR #879, opened 2023-07-31
Update history table during shipment (box state)

pylipp:
- Create list when querying shipment details instead of appending
- Update history test data
- New function for bulk updating box state during shipments and tracking the change


pylipp:
FYI overall coverage decreases because I removed covered lines (i.e. I made the code more concise by removing code duplication). The patch coverage is 100%.

PR #881, opened 2023-08-03
Enhance query performance using batch-loading

pylipp:
Make more use of batch-loading for user, base, and organisation fields, hence avoiding excessive DB queries.


pylipp:
[2.0 Enhance query performance](https://trello.com/c/Z5LTysZt/1242-20-enhance-query-performance)

PR #882, opened 2023-08-05
prevent creating identical agreements

pylipp:
When creating an agreement, now it is first checked whether an identical accepted agreement exists, i.e. same set of organisations and same (sub)set of bases involved, and overlapping validity periods. If so, the GraphQL response contains an error with `BAD_USER_INPUT` code.


pylipp:

HaGuesto (approved):
approved

PR #883, opened 2023-08-06
use timezone from user jwt

pylipp:
Timezone info is needed for the validity times input when creating an agreement. This PR pulls the info from the user JWT (put there by Auth0), and also drops a Python package dependency.


pylipp:
[2.0 Use timezone info from user JWT](https://trello.com/c/qOWNZ8rI/1256-20-use-timezone-info-from-user-jwt)

HaGuesto (approved):
ok cool. passes my tests. Just I do not understand or remember what was the reason to work on it

Issue #888, opened 2023-08-09
Research data preparation in backend for first user story


pylipp:
## user story

I, as a warehouse coordinator, want to see how many boxes and items I have moved from a warehouse location to a donated location like a free shop in a time range, possibly filterable by product, size, gender. The diagram should be a bar plot where each bar represents a customizable time period.

## Task description
@pylipp figure out what we would have to do to prepare the data and pass it to the frontend.
Please start a doc where we collect the additional queries we have to create for the visualizations with their requirements, variables, user stories, and the data structure that is returned.

Please consider at least two options: write the query in the current data-structure and in the perfect data structure for the query.
What would be the trade-offs in performance, cost and implementing the data structure/query

Update 2023-07-24: list of boxes grouped by location for one product in a given time range
Update August 2023: first user story shall be about beneficiary demographics.

pylipp:
## User story: beneficiary demographics

We'd like to obtain the number of beneficiaries (fact) for all sets of gender, age(-bin), and creation date (dimensions).

### BE implementation

Such data can be obtained by first querying beneficiary rows, then transforming it into a pivot table (e.g. using the pandas framework; `id` as values, `gender, created_on, age` as index, `count` as aggfunc), and finally transforming this table into single data points again.

However due to the simplicity of the aggregation, the data can be obtained directly from the DB via 

```python
    bin_width = 5
    gender = fn.IF(Beneficiary.gender == "", "D", Beneficiary.gender)
    created_on = db.database.truncate_date("day", Beneficiary.created_on)
    age = fn.FLOOR((date.today().year - Beneficiary.date_of_birth.year) / bin_width)

    demographics = (
        Beneficiary.select(
            gender.alias("gender"),
            created_on.alias("created_on"),
            age.alias("age"),
            fn.COUNT(Beneficiary.id).alias("count"),
        )
        .where(Beneficiary.deleted.is_null())  # add base IDs
        .group_by(SQL("gender"), SQL("age"), SQL("created_on"))
    )
```

### Data format

The resulting data format is a list of objects with keys `age`, `gender`, `created_at`, and `count`.

### GraphQL interface

To return the data via GraphQL the following extension is proposed:

```graphql
type Metrics {
  ...
  beneficiaryDemographics(baseIds: [Int]): [BeneficiaryDemographicsResult]
}

type BeneficiaryDemographicsResult {
  age: Int
  gender: HumanGender
  createdOn: Datetime
  count: Int
}
```

[...]

PR #891, opened 2023-08-09
initial visx population chart with random data

MaikNeubert:
First version of the demographic StatViz using random data for demonstration

Test that the chart displays the correct values (you can use the browser debugger to check what the random generated values are)

other features to test
Tooltip on the Bars shows the exact values
Adjusting height and width in "PeopleByAgeChart" const chart scales the chart

Tests will be added with the backend integration


react/src/views/Statistics/components/PeopleByAgeChart.tsx
```diff
@@ -0,0 +1,77 @@
+import BarChartCenterAxis, { IBarChartCenterAxis } from "components/graphs/BarChartCenterAxis";
+import { random, uniq } from "lodash";
+
+const dataCube = {
+  factTable: [
+    { gender: "M", age: 1, created: "2021-06-22", people: 3 },
```
pylipp, line 6: @MaikNeubert [this post](https://github.com/boxwise/boxtribute/issues/888#issuecomment-1670889773) shows the data format provide by the BE. 
1. `gender` will be one of Male/Female/Diverse
2. `created` will be `createdOn`
3. `people` will be `count`
(1.) cannot be changed since it's dictated by the GraphQL enum HumanGender. (2.) can be changed but would be inconsistent (datetime fields in our GraphQL schema always have suffix `On`). (3.) can be changed :slightly_smiling_face:

package.json
```diff
@@ -0,0 +1,5 @@
+{
```
pylipp, line 1: I think this file is not needed because `yarn` should be executed in the `react/` folder, and the relevant `package.json` lives there.

MaikNeubert: Yes, i removed this package.json

react/src/views/Statistics/components/PeopleByAgeChart.tsx
MaikNeubert: fields on the frontend where updated to match the BE

react/src/components/graphs/BarChartCenterAxis.tsx
```diff
@@ -0,0 +1,239 @@
...
+              xScale={scaleXLeft}
+              yScale={scaleY}
+            />
+          </Group>
+          <Group>
+            {fields.dataXl.map((element) => {
+              const barWidth = halfWidth - scaleXLeft(element.x);
+              const x = scaleXLeft(minX) - barWidth;
+              const y = scaleY(element.y);
+              const tooltip = `${fields.labelY} ${element.y}, ${fields.labelXl} ${element.x}`;
```
pylipp, line 148: When building locally I get an eslint warning here: `'tooltipData' is already declared in the upper scope on line 68`, same for line 178

pylipp: Nvm, latest commits resolved this :+1:

pylipp (approved):
## Functional review
I could run the code and view the graph at `/bases/1/statistics` for a boxaid user. I checked the "responsiveness" of the graph when adjusting the hard-coded width. Nice work :sunglasses:

react/src/components/graphs/BarChartCenterAxis.tsx
```diff
@@ -0,0 +1,239 @@
...
+  if (!fields.settings) {
+    fields.settings = {};
+  }
+  const settings = {
+    ...defaultSettings,
+    ...fields.settings,
+  };
+
+  const chartWidth = fields.width - (marginLeft + marginRight);
+  const chartHight = fields.height - (marginTop + marginBottom);
```
pylipp, line 90: typo: it's chartHeight, also barHeight :see_no_evil:

PR #899, opened 2023-08-10
Connect backend "beneficiaryDemographics" with the frontend visualization

MaikNeubert:
This PR connects the backend for the Demographics chart.
In order to connect the backend, the statviz frontend was seperated from the boxtribute app frontend


react/tsconfig.json
```diff
@@ -30,5 +30,5 @@
     "pretty": true,
     "sourceMap": true
   },
-  "include": ["src"]
+  "include": ["src", "../statviz/src/views/Statistics", "../statviz/src/components/graphs"]
```
HaGuesto, line 33: @MaikNeubert please remove the latter two paths. They are the reason for the build failure. statviz has its own tsconfig and this way they are intersecting where they should not.

statviz/.vite/deps_temp_4475ae61/package.json
```diff
@@ -0,0 +1,3 @@
+{
```
pylipp, line 1: not sure if this is needed, looks like a temp file

statviz/src/components/Form/DateField.tsx
```diff
@@ -0,0 +1,68 @@
+import { FormControl, FormErrorMessage, FormLabel, Input, Text } from "@chakra-ui/react";
+import { Controller } from "react-hook-form";
+
+export interface IDateFieldProps {
+  fieldId: string;
+  fieldLabel: string;
+  errors: object;
+  control: any;
+  register: any;
+  // eslint-disable-next-line react/require-default-props
```
HaGuesto, line 10: i note down that we should turn off this rule in eslint

statviz/src/components/graphs/BarChartCenterAxis.tsx
```diff
@@ -48,14 +48,14 @@ const defaultSettings = {
...
 
-const marginTop = 20;
-const marginLeft = 45;
+const marginTop = 0;
+const marginLeft = 70;
 const marginRight = 40;
-const marginBottom = 40;
+const marginBottom = 70;
 
 const labelProps = {
```
HaGuesto, line 56: in the future, I would love if we could somehow just load it from the customized chakra theme.

statviz/src/views/Statistics/StatisticsView.tsx
```diff
@@ -0,0 +1,45 @@
...
+    beneficiaryDemographics {
+      facts {
+        count
+				createdOn
+				age
+				gender
+      }
+      dimensions {
+        tag {
+          __typename
```
HaGuesto, line 16: IN general, u do not have to add _typename. This is a default option of the InMemoryCache from Apollo

statviz/src/components/graphs/BarChartCenterAxis.tsx
HaGuesto: No ACTION ITEM atm. Just something to consider for the future

statviz/src/views/Statistics/StatisticsView.tsx
```diff
@@ -0,0 +1,45 @@
...
+`
+
+export default function StatisticsView() {
+
+  const { data, loading, error } = useQuery<
+    BeneficiaryDemographicsQuery,
+    BeneficiaryDemographicsQueryVariables
+  >(DEMOGRAPHIC_QUERY);
+
+  if (loading || data?.beneficiaryDemographics === undefined) {
```
HaGuesto, line 32: The second part of the if does not make much sense. If an error happens, data?.beneficiaryDemographics could be undefined and we would show the loading state.

ACTION ITEM: add error handling.

pylipp (approved):
**Functional review**
I started the webapp and statviz services and `localhost:5173/demographic` shows the graph as expected. I tinkered with the age-bin in the back-end (currently hard-coded to 5; you might want to set it to 1 here), and the graph updates correctly.
Thanks for the good work, the first step is always the hardest! also @HaGuesto

statviz/src/views/Statistics/components/DemographicChart.tsx
```diff
@@ -0,0 +1,88 @@
+import { useState } from "react";
+import { Heading } from "@chakra-ui/react";
+import BarChartCenterAxis from "../../../components/graphs/BarChartCenterAxis";
+import { range } from "lodash";
+import FilterCreatedOn from "./filter/FilterCreatedOn";
+import { HumanGender } from "../../../types/generated/graphql";
+
+export interface IDemographicFact {
```
HaGuesto, line 8: It is a bit weird that you have to define these types. Should the not also be generated by the codegen tool?

Something to check out in the future.

statviz/.vite/deps_temp_4475ae61/package.json
HaGuesto: agreed. 

ACTION ITEM: Please remove and add .vite to gitignore

HaGuesto (approved):
@MaikNeubert just a few minor change requests in the code. Functionally everything works as expected.

I already approved it so that you can merge it yourself. I trust you that you will do the changes and then merge when you are ready. No further review needed from my side.

statviz/src/views/Statistics/components/DemographicChart.tsx
MaikNeubert: Yes, the problem with the generated types is, that fields are nullable.
Since most fields there aren't null in any case i used this interface to not be forced to do null checks everywhere

PR #945, opened 2023-08-17
CreateTransferAgreementView: Alerts on Identical Agreements

vahidbazzaz:
This PR add alert on identical agreement and implement test case 4.1.5.

### DIFF
--- a/react/src/views/Transfers/CreateTransferAgreement/CreateTransferAgreementView.test.tsx
+++ b/react/src/views/Transfers/CreateTransferAgreement/CreateTransferAgreementView.test.tsx
@@ -74,6 +74,33 @@ const mutationNetworkError = {
   error: new Error(),
 };
 
+const mutationIdenticalAgreementError = {
+  request: {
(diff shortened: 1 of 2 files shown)


vahidbazzaz:

HaGuesto (approved):

pylipp:
@vahidbazzaz @HaGuesto I tested this and it does not work as expected. On v2-staging, I log in as boxaid dev-coordinator, and create a new agreement with BoxCare (all bases, bidirectional). On submitting the form two error messages appear: "something went wrong" and "Your changes could not be saved". But not "Can’t create agreement, an active identical agreement exists"

HaGuesto:
@pylipp weird. It is working for me locally and the test is fine, but it also does not work on staging. I take a look.

HaGuesto:
There is no consistent behavior atm. I managed to make it work with the correct error message, but I also was able to reproduce the state in staging without knowing exactly how. @vahidbazzaz

PR #946, opened 2023-08-17
[1st viz] Add beneficiary tag IDs to demographics query

pylipp:
See comment in #888 for the updated data format.

The FE can now filter for a combination of tag IDs in the result. A dimensions table is provided such that human-readable tag names can be used.


HaGuesto (approved):

HaGuesto:
@pylipp where is the connected ticket in the github project?

Otherwise, code and functional review are passing.

PR #948, opened 2023-08-18
fix transferagreement caching error

HaGuesto:


HaGuesto:

HaGuesto:
@pylipp we fixed the problem. Feel free to check it on staging now.

PR #950, opened 2023-08-20
ADR for backend of DSEE 100x project with Hermine "Statviz"

HaGuesto:
[Rendered version](https://github.com/boxwise/boxtribute/blob/adr/statviz_backend/docs/adr/adr_backend_etl_and_hosting_for_StatViz.md)

### DIFF
--- a/docs/adr/adr_backend_etl_and_hosting_for_StatViz.md
+++ b/docs/adr/adr_backend_etl_and_hosting_for_StatViz.md
@@ -154,9 +154,9 @@ We learned (see PR #1013) that
 
   --> the data structure is compatible to respond to (probably) all queries.
 
-- However, it comes at a great cost (see PR #1013). Even with a small sample size the latency of recreating old states of the data with the transactional data from the "history" table lead to an unacceptable latency of >10s.
+- However, it comes at a great cost when the computation-heavy part is executed in pure Python (see PR #1013). Even with a small sample size the latency of recreating old states of the data with the transactional data from the "history" table lead to an unacceptable latency of >1s.
 
-  -->**We need to add a cache or extend the schema** to make the analytical queries run faster.
+  -->**We need to add a cache, or extend the schema, or use optimized SQL queries** to make the analytical queries run faster.
 
 ### B.1 Extending the DB schema
 
@@ -247,6 +247,19 @@ The question when and how the cache is populated is similar to B.1.3.
 
 - probably each query needs a cache
 
+### B.3 Pure SQL queries
+
+SQL alone is a very powerful to extract and transform data, especially with appropriate indexing of tables. The largest tables by far are the history table (1.5M rows) and the stock table (100k rows) which is considered fairly small for any production database.
+
+**Pros**:
+
+- queries can be integrated in pure SQL (previously tested via SQL IDE or similar)
+
+**Cons**:
+
+- advanced SQL knowledge required
+- query performance might deteriorate by time (increasing data size)
+
 ## Decisions
 
 1. **[DONE] Data Format - Incorporation of User Filters**: We will integrate User Filters as dimensions in the returned data.
@@ -257,23 +270,24 @@ The question when and how the cache is populated is similar to B.1.3.
 
    Visualizations are typically embedded via URLs. This design means users won't have dynamic control over these filter values within embedded visualizations. Consequently, embedding URL filters as query variables suffices without burdening the system with additional dimensions that offer no direct user benefit.
 
-3. **[DONE] Data structure - We need to add caching or extend the db schema to speed up certain analysis queries**:
+3. **[DONE] Data structure - We need a way to speed up certain analysis queries**:
 
    See section B.
 
-4. **[OPEN] We rather extend the db schema than cache the result of each query in a cache storage**:
+4. **[DONE] ~~We rather extend the db schema than cache the result of each query in a cache storage~~ We use advanced SQL queries**:
 
-   Extending the db schema is the solution that creates less duplication of data and is less complex to implement and maintain. File-system and Redis caching would additionally increase complexity and probably cost.
-   We currently work on a concrete proposal for the data schema, but it will most likely be additional shadow tables for stock and people with inter-linking to the "history" table.
+   We avoid the additional complexity that would be introduced by adjusting the software architecture and/or database structure.
+   ~~Extending the db schema is the solution that creates less duplication of data and is less complex to implement and maintain. File-system and Redis caching would additionally increase complexity and probably cost. We currently work on a concrete proposal for the data schema, but it will most likely be additional shadow tables for stock and people with inter-linking to the "history" table.~~
 
 5. **[DONE] Hosting - We will implement read-only replicas and host the public endpoint on a separate service**:
 
-   With our visualizations accessible via a public endpoint and the anticipation of embedding them on public channels, there's potential for a rapid increase in backend requests. This might peak especially during promotional campaigns by our users. To prevent any disruption to the Boxtribute app (most important decision driver):
+   ~~With our visualizations accessible via a public endpoint and the anticipation of embedding them on public channels, there's potential for a rapid increase in backend requests. This might peak especially during promotional campaigns by our users.~~ To prevent any disruption to the Boxtribute app (most important decision driver):
 
-   - We host the public GraphQL API endpoint on a distinct Google service, separate from the Boxtribute backend.
+   - we make visualizations accessible only to Boxtribute users. Public exposure is postponed, also for data privacy concerns.
+   - ~~We host the public GraphQL API endpoint on a distinct Google service, separate from the Boxtribute backend.~~
    - we will query analytical data only from a read-only replica
 
-6. **[OPEN] Security measurements to separate public from private personal data**:
+6. **[POSTPONED] ~~Security measurements to separate public from private personal data~~**:
 
    During the creation of this ADR, we also touched how we can ensure that no personal data - data that identifies a person - is passed through the public endpoint:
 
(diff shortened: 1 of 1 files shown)


HaGuesto:
@aerinsol this is the ADR we should talk with David about

docs/adr/adr_backend_etl_and_hosting_for_StatViz.md
```diff
@@ -0,0 +1,237 @@
...
+
+3. **Cost Efficiency**: Implementing the ETL process and serving the data should be done with minimal additional financial impact.
+
+4. **Server Response Latency**: The time it takes for the server to respond with the requested data should be minimized to ensure a seamless user experience.
+
+5. **Manageable Data Size**: The size of the returned data should be optimized so that frontend applications can perform operations like dynamic filtering and sorting without performance degradation.
+
+6. **Utilization of Existing Data**: We should target that we can already leveraging the data already present in Boxtribute.
+
+7. **Scalability**
```
pylipp, line 41: more thoughts here? ;)

docs/adr/adr_backend_etl_and_hosting_for_StatViz.md
```diff
@@ -0,0 +1,237 @@
...
+This is straight-forward solution and how we started for the first visualizations - Directly transform operational data from Boxtribute in response to incoming graphQL requests.
+
+**Pros**:
+
+- you only need to write the query
+- live data
+
+**Cons**:
+
+- query with many reads directly from the Boxtribute db
```
pylipp, line 132: another con (maybe you see it implicitly covered): historically accurate data needs to be reconstructed

docs/adr/adr_backend_etl_and_hosting_for_StatViz.md
```diff
@@ -0,0 +1,237 @@
...
+  --> no huge increase in cost (depending on the cache), low latency
+
+**Cons**:
+
+- no live data
+- more code writing nightly transformation and caching
+
+#### B.3 Reactive Transformation on Boxtribute Data Updates
+
+Instead of preparing the data for analytics once a night, we could also directly update the cache when a change happens to the Boxtribute operational db. We have not really considered this option, but it would be possible.
```
pylipp, line 164: This would imply either
- implementing the cache update logic in both dropapp and v2, or
- use implementation on DB level
First one is double work; second one is too raw, and also inconsistent to have logic on DB level in my opinion. All in all B.3 is a no-go in my opinion

docs/adr/adr_backend_etl_and_hosting_for_StatViz.md
```diff
@@ -0,0 +1,237 @@
...
+
+3. **[Data Transformation] Nightly Transformations**:
+
+   With our visualizations accessible via a public endpoint and the anticipation of embedding them on public channels, there's potential for a rapid increase in backend requests. This might peak especially during promotional campaigns by our users. To prevent any disruption to the Boxtribute app:
+
+   - We propose hosting the public GraphQL API endpoint on a distinct Google service, separate from the Boxtribute backend.
+   - Data will undergo nightly transformation and be cached independently. Consequently, requests to the public endpoint will bypass resources that the Boxtribute app directly accesses.
+
+4. **[Data Transformation] Caching in DB**:
+   We aim to cache this data in dedicated database tables to avoid the added complexity of filesystem or Redis caching. This is cost-effective in comparison to Redis caching. A lingering question, however, is whether to house this data in a distinct database or an isolated MySQL instance on Google Cloud SQL to guarantee no performance disruptions to the Boxtribute app.
```
pylipp, line 216: From a BE perspective it's easier to have all data in one database. Otherwise data fetching becomes complicated: at least two DB connections have to be managed (fetch cache data for facts; fetch data for dimensions, e.g. product names). Sure we can duplicate certain tables for dimensions in the cache DB but that's prone to inconsistencies.

docs/adr/adr_backend_etl_and_hosting_for_StatViz.md
```diff
@@ -0,0 +1,237 @@
...
+
+6. **[Data Structure] No additional db tables are needed in the Boxtribute DB**
+
+   The history table is sufficiently detailed, enabling us even to recreate past data snapshots. Since we are aiming for a nightly transformations, transforming the db (e.g. recreating past snapshots) do not have to be optimized for runtime.
+
+   While there's an option to introduce shadow tables for core tracked data like stock and people, transformations would still be essential to tailor this data for varied visualizations. Thus, there's no compelling reason to modify the operational database structure of Boxtribute.
+
+   Nevertheless, we should do an audit to ascertain that all transactions are tracked within the history table. For instance, operations like assigning or removing tags from stock or people currently bypass the history table.
+
+## Next steps
```
pylipp, line 229: After some of these are clarified, we should also update the C4 diagrams to have an architectural overview :)

pylipp (approved):
Thanks for the detailed ADR, I think most of our thoughts of the last weeks are summarized well. I also like that you thought of integrity tests of the transformed data.

docs/adr/adr_backend_etl_and_hosting_for_StatViz.md
```diff
@@ -0,0 +1,237 @@
...
+
+## Status
+
+proposed
+
+## Context
+
+In the DSEE 100x project titled "StatViz", our goal is to construct data visualizations derived from Boxtribute's data. This will allow our users to showcase their contributions and impact to the public through various mediums, such as their websites or newsletters.
+
+Currently, data meant for these visualizations is planned to be served through a public GraphQL endpoint that is implemented using a stack of MySQL, Peewee, Flask, and Ariadne.
```
pylipp, line 15: Since performance is also of concern below, it might be worth explaining why we go for a GraphQL API here instead of REST GET endpoints (I think it might be less overhead to directly serve the data as JSON over REST rather having all the resolver overhead. OTOH we'll miss out on the implicit schema and type definitions)

docs/adr/adr_backend_etl_and_hosting_for_StatViz.md
HaGuesto: good point. haven't thought of that yet.

docs/adr/adr_backend_etl_and_hosting_for_StatViz.md
HaGuesto: and What is your reasoning then?

I took the graphQL endpoint as a given :) just because we are accustomed to it.

pylipp: I'm also in favor of using GraphQL because it's better known to us, and because the schema is defined.

docs/adr/adr_backend_etl_and_hosting_for_StatViz.md
```diff
@@ -0,0 +1,237 @@
...
+
+4. **Server Response Latency**: The time it takes for the server to respond with the requested data should be minimized to ensure a seamless user experience.
+
+5. **Manageable Data Size**: The size of the returned data should be optimized so that frontend applications can perform operations like dynamic filtering and sorting without performance degradation.
+
+6. **Utilization of Existing Data**: We should target that we can already leveraging the data already present in Boxtribute.
+
+7. **Scalability**
+
+8. **Data Freshness**: There isn't a requirement for real-time data. A one-day lag in data freshness is acceptable, which might open up opportunities for batch processing or scheduled updates.
```
aerinsol, line 43: one day lag is not as precise as an x hour lag. I'd also be interested in how much of a lock-in this requirement is. If we want to move to, say, a 3 hour lag, what's the difficulty / implication of that?

pylipp (approved):
Thanks for the update :+1:

aerinsol (changes requested):
Does not reflect my concerns on security or even reference our discussion there.

Conclusion on the performance issues does not reflect my detailed commentary in the test PR on the inaccuracy of the way optimization was done, and is therefore I believe a premature conclusion. 

Retracting approval as this is no longer a first draft

docs/adr/adr_backend_etl_and_hosting_for_StatViz.md
```diff
@@ -114,124 +116,185 @@ beneficiaryDemographics {
...
 
-This is straight-forward solution and how we started for the first visualizations - Directly transform operational data from Boxtribute in response to incoming graphQL requests.
+  -->**We need to add a cache or extend the schema** to make the analytical queries run faster.
+
+### B.1 Extending the DB schema
+
+Especially querying historical snapshots of data is read-intensive with the transactional "history" table.
+Therefore, to reduce the latency of such queries we consider adding data structures to the schema that make querying snapshots easier.
+
+#### B.1.1 Adding Bi-temporal / Shadow Tables
```
pylipp, line 166: After reading [this blog post](https://martinfowler.com/articles/bitemporal-history.html) I don't think bitemporal tables solve the issue we have, I think we want plain shadow/temporal tables.

aerinsol:
Logging August 26, 2023 first feedback provided by @jdcsgh after reading the first iteration of this PR. Feedback was provided directly to me over email. 

***

> On-the-fly Transformation During GraphQL Requests:
> Cons: increase of cost & latency, not scalable, interference with the Boxtribute db

These are strong claims. But it's not immediately obvious that they're true. For decision-critical factors like this, it might be helpful to try an adversarial challenge process to get to the truth of the matter: E.g., Engineer #1 claims that on-the-fly translation from GraphQL to SQL is too slow, and exhibits a sample query with timings to demonstrate that. Engineer #2 sees that the query is suboptimally written, and shows how it can be optimized to run in a reasonable time, etc. After some back and forth, you should be able to converge on the truth.

My intuition tells me that Boxtribute only has a modest amount of data and users, and on-the-fly transformation should be totally doable. But I would need to see actual queries to make the call.

> Scheduled Nightly Transformation
> Cons:
> no live data

From personal experience working on projects where we did this, this is a very large con:

- Your users will get frustrated because the data is stale. You'll get a large volume of support requests, which will take up a lot of developer time to investigate, and at the end of the process, you'll have to reply to your users with "please try again tomorrow," which makes no one happy.

- Your data visualization users will be using a different database than your developers, which leads to the "out-of-sight, out-of-mind" problem. Many of your junior developers will not touch the nightly transformation pipeline at all, or even know about it, which means that they will introduce changes that will break it in subtle ways. And your developers are constantly looking at the main DB, but not the transformation cache, so bad data in the transformation cache might not be noticed for days or weeks.

I've seen this cause resentment in both directions. The data visualization users will think, "ugh, why are the developers so sloppy and constantly breaking my dashboards?" And the developers will think, "another stupid bug in the transformation pipeline -- the data in the main DB is perfectly correct, but now I have to waste my time tracing through a transformation pipeline I barely understand to figure out how to fix their dashboards."

Our current method of tracking changes to database records, like stock or people, employs a centralized history table.
This data structure efficiently captures transactional data changes, but extracting historical snapshots (like warehouse stock at a particular time) is more complex.

For this use case, please consider "bitemporal"/"temporal" tables for tracking history. There's a lot of literature about this, but a very simplistic implementation could look like this:

// change location from WH1 to Free Shop
(..., box_id, time_lower, time_upper, box_location, ...)
(...,  "box1", 2023-08-01 00:00, 2023-08-07 12:00, "WH1", ...)
(...,  "box1", 2023-08-07 12:00, infinity, "Free Shop", ...)

Basically, each row asserts a fact that is true for the duration of time_lower to time_upper. So in this example, box1 was in WH1 until 2023-08-07 12:00, and then it moved to Free Shop at 2023-08-07 12:00.

This schema makes it extremely efficient to query for a snapshot at any instant in time. To see where a box was at time t, you would just say:

SELECT ... FROM ... WHERE time_lower <= t AND time_upper > t;

This query will be extremely fast with proper indexes on the time_lower and time_upper columns. We used this design extensively at my last job.

> Decisions

Speaking purely on intuition, and **not knowing anything about your data**, my inclination is to query the main DB directly, and not do a nightly transformation. (bold added by @aerinsol, as this relates to current feedback by @jdcsgh ).

To isolate your data visualizations from adversely affecting your main DB, one common design is to run your DB with replicas:

- The main DB is read/write, and is used by your core systems that need to update the data.
- Then you run one or more read-only replicas, which your data visualization and dashboards can query. Because these are separate replicas, even if the queries completely bog down the read-only replica, it won't hurt your core systems.

This is a very popular configuration, and DB servers like PostgreSQL have built-in support for it. support for it.

docs/adr/adr_backend_etl_and_hosting_for_StatViz.md
HaGuesto: Hey @pylipp I read the link and I believe what David means with bi-temporal table is not the same as in the blog post. Bi-temporal means that it saves two dates. For David the two dates are "from" and "to" datetime so that we only have to compare columns and not rows. In the blog post it is about record date and actual date which I agree is not a concept that helps us.

HaGuesto:
@pylipp can we update the ADR for the statviz BE so that it reflects what we did at the end and why?

pylipp:
@HaGuesto putting it on my todo-list

HaGuesto:
@aerinsol I checked this and from my side we can merge it

aerinsol:
Approved

Issue #903, opened 2023-08-10
[3rd viz] create user story / spec for 3rd viz


HaGuesto:

MaikNeubert:
example
https://github.com/kdivringi/visx-rodeo/blob/main/src/Sankey.tsx

pylipp:
@aerinsol can you define the user story, or link one from the [spreadsheet](https://docs.google.com/spreadsheets/d/1BGRHsgQsOMS8QahhsivA7yB_ayR9TJiGn5AJF2uJgX0/edit#gid=431000046). IIRC it's about flow of boxes from one location to others, comparable to what was originally written [here](https://github.com/boxwise/boxtribute/issues/888#issue-1842716458) or in #878 .

aerinsol:
Currently holding off on creation on Sankey or similar flow-based diagrams as we are awaiting more finalized input from @seinig .

Instead, the feature request will now be to create a pie chart similar to the [sample shown here](https://docs.google.com/spreadsheets/d/1JE2Of1Yj_pd3_JAsIHxivkBxT7PQmpd8Cxb15Fn7jJo/edit?usp=sharing) with a ranked list:

Backend @pylipp : Top Ranked Products given a Custom Time Frame.
Distribution: Showcase which were the top 5 product distributed (checked out) with 1) ranking and 2) total item numbers in that time frame frame.
Inventory - corresponds to user story 3.1: Showcase which were the top 5 products - with both 1) ranking and 2) total item numbers 3) with the option to filter/group by `size` - moved to DONATED state within that time frame; OR created during that time frame.

Front End @MaikNeubert : Create a pie chart that groups by product categories all items checked out within a specified time frame AND/OR all all items CREATED within a certain time frame. Simultaneously show next to it a ranked list of top items checkout / created during that time period.

Additional comment - @pylipp the cube that you created for the 2nd visualization should probably be updated to return ProductName and CategoryName, rather than ProductId and CategoryId - @MaikNeubert please state if you'd like to keep the current ID numbers instead of text strings for your purposes.

pylipp:
@aerinsol using ProductId and CategoryId is by intention. Using a star-like data format the corresponding names can be looked up in dimension tables as described in #950 .

HaGuesto:
@aerinsol regarding:

[...]

PR #969, opened 2023-08-22
[3rd viz] BE for top-products queries

pylipp:
This PR brings the `topProducts*` queries as suggested by #904.

For testing you can start the webapp service and run

```shell
xh POST :5005/public query='query { topProductsCheckedOut(baseId: 1) { facts { itemsCount checkedOutOn productId } } }'
```

or

```shell
curl -X POST 'http://localhost:5005/public' -H 'content-type: application/json' -H 'accept: application/json, */*;q=0.5' -d '{"query":"query { topProductsCheckedOut(baseId: 1) { facts { itemsCount checkedOutOn productId } } }"}'
```


back/boxtribute_server/business_logic/statistics/crud.py
```diff
@@ -3,17 +3,56 @@
...
+            .where(Product.id << product_ids)
+            .dicts()
+        )
+
+    if "category" in names:
+        dimensions["category"] = ProductCategory.select(
+            ProductCategory.id, ProductCategory.name
+        ).dicts()
+
+    if "tag" in names:
```
HaGuesto, line 39: I think it would be nice if you could also return the color here in general.

back/boxtribute_server/business_logic/statistics/crud.py
```diff
@@ -3,17 +3,56 @@
...
+    if "product" in names:
+        product_ids = {f["product_id"] for f in facts}
+        dimensions["product"] = (
+            Product.select(Product.id, Product.name)
+            .where(Product.id << product_ids)
+            .dicts()
+        )
+
+    if "category" in names:
+        dimensions["category"] = ProductCategory.select(
```
HaGuesto, line 35: Please add the parent_id here too.

back/boxtribute_server/business_logic/statistics/crud.py
```diff
@@ -91,16 +126,83 @@ def compute_created_boxes(base_id=None):
...
+        row["rank"] = rank
+        row["checked_out_on"] = row["checked_out_on"].date()
+
+    dimensions = _generate_dimensions("category", "product", facts=facts)
+    dimensions["size"] = None
+    return {"facts": facts, "dimensions": dimensions}
+
+
+def compute_top_products_donated(base_id):
+    """Return list of most-donated products with rank included, grouped by distribution
```
HaGuesto, line 162: Will we also count boxes assigned to shipments to it in the future?

back/boxtribute_server/business_logic/statistics/crud.py
```diff
@@ -3,17 +3,56 @@
...
+def _generate_dimensions(*names, facts):
+    """Return a dictionary holding information (ID, name) about dimensions with
+    specified names.
+    """
+    dimensions = {}
+
+    if "product" in names:
+        product_ids = {f["product_id"] for f in facts}
+        dimensions["product"] = (
+            Product.select(Product.id, Product.name)
```
HaGuesto, line 29: Here, is something we might want to talk about. A Product always has a gender. If we do not add this to the dimension, many values will have the same name.

For example:
Usually there is a product "Winter Jacket" for Male and Female. They are two separate products, but have the same name.

Proposal:
Either we concatenate the Gender name to the Product name or we create an extra dimension for gender.

If we add gender as a dimension, we can also think about grouping the facts not by product id, but by product name.

back/boxtribute_server/business_logic/statistics/crud.py
pylipp: User story in #903 does not mention this, maybe additional thoughts from @aerinsol ?

back/boxtribute_server/business_logic/statistics/crud.py
pylipp: Would it also work to return `gender` in the product dimension table? I think it gives most flexibility, and FE can then concatenate product and gender to their liking.

pylipp: addressed in #998

back/boxtribute_server/business_logic/statistics/crud.py
pylipp: addressed in #998

back/boxtribute_server/business_logic/statistics/crud.py
HaGuesto: i put a comment about this into #903

HaGuesto (approved):

PR #971, opened 2023-08-23
[MonoRepoHygiene1] Rename react folder to front and add yarn workspaces

HaGuesto:
Before doing any kind of config unification for mono-repos we have to rename the react folder.
Otherwise, yarn might have problems with resolving the react package.

I based this PR on #899 since I wanted to start with yarn workspaces from a point where a first working version of both frontends exists.


HaGuesto:
@pylipp I updated the CircleCI config to work with yarn workspaces and I would love you to focus your review on that.

.yarn/releases/yarn-1.22.5.cjs
pylipp, line 1: @HaGuesto not sure if  this 150k lines file is supposed to be committed.

package.json
pylipp, line 1: @HaGuesto this is all the magic, and now both projects can share things between them?

pylipp (approved):
I don't see much of a change in the circleci config besides switching docker images for executors. Apparently there are no other changes needed for setting up the projects? At least the READMEs don't mention anything new.

.yarn/releases/yarn-1.22.5.cjs
HaGuesto: There is actually an argument in the community about this, but I have not figured it out myself yet completely.

HaGuesto: ATM it is not important, so I will remove it, but when we upgrade yarn it might make sense to pin down the version of yarn.

package.json
HaGuesto: basically yes

aerinsol:
[2.0 Update CircleCI orbs (Slack/codecov integration)](https://trello.com/c/ecd0WUre/1081-20-update-circleci-orbs-slack-codecov-integration)

aerinsol:
We also have a request on our backlog to update CircleCI orbs as attached above. 

Is that still relevant?

PR #985, opened 2023-08-25
BoxView: Remove Extra Padding When Hiding Tags

vahidbazzaz:
This PR remove extra space between box identifier and box status when there are no tags assigned to a box.


HaGuesto (approved):
looks good

PR #1009, opened 2023-09-01
update fetch-policy of apollo to "cache-and-network" for overviews

HaGuesto:


HaGuesto:
[2.0 BUG When having ShipmentView open for Shipment A, adding new boxes in a separate window then using "Back to Shipments" button, selecting shipment A does not produce the updated list of boxes. #softlaunch #tiny](https://trello.com/c/AAO6lRjG/1263-20-bug-when-having-shipmentview-open-for-shipment-a-adding-new-boxes-in-a-separate-window-then-using-back-to-shipments-button-se)

PR #1010, opened 2023-09-01
ReconciliationView: Change "Remaining" to "Received"

vahidbazzaz:
The PR changes "Remaining" to "Received" so that the number of boxes increases as they are received rather than decreasing.


vahidbazzaz:
[2.0 ReconciliationView - Change "Remaining" to "Received", so that the number starts at 0/100 and completes at 100/100](https://trello.com/c/ihgx3w1r/1268-20-reconciliationview-change-remaining-to-received-so-that-the-number-starts-at-0-100-and-completes-at-100-100)

PR #1011, opened 2023-09-01
BoxView: History Message Should Use "box status" Instead of "box state"

vahidbazzaz:
This PR changes the history message from "box state" to "box status" in order to comply with user-facing terminology.


vahidbazzaz:
[2.0 BoxView History message should use "box status" not "box state" to be consistent with user-facing language #tiny](https://trello.com/c/ClAikFIk/1270-20-boxview-history-message-should-use-box-status-not-box-state-to-be-consistent-with-user-facing-language-tiny)

PR #1012, opened 2023-09-01
ReconciliationView: Open Accordion Upon First Render

vahidbazzaz:
The PR fixes reconciliation view accordion to open/focus on when it is first rendered.


vahidbazzaz:
[2.0 ReconciliationView: should open/focus on accordion 1 upon first render](https://trello.com/c/BWnnUwJu/1274-20-reconciliationview-should-open-focus-on-accordion-1-upon-first-render)

PR #1013, opened 2023-09-01
Add CLI to reconstruct box versions from history

pylipp:
## Goal
Reconstruct box versions from history for precise data in `createdBoxes` query.

It is necessary to find the number of items in the box *at the time of creation*. This can only be obtained by walking back all changes applied to a box.

At this point, the PR does not integrate the reconstruction in the actual `createdBoxes` BE but provides a CLI for experimenting and monitoring instead.

## Approach
As test data I got a production DB dump via

    mysqldump --gtid --tz-utc --dump-date --skip-lock-tables --disable-keys --quote-names --create-options --add-locks --extended-insert -u dropapp_selectonly --protocol=tcp -p --host 127.0.0.1 --port 3386 dropapp_production stock locations history > stock-history.sql

I used this dump as volume for the docker-compose db service.

Now the reconstruction script can be invoked via

    bwiz -P 32000 -u root -p dropapp_root -d dropapp_dev --verbose convert -l 54 -y 2017

to reconstruct all versions of boxes in location 54 created in 2017 or after.

## Result
The location above has about 3800 boxes with >20k history entries. Reconstruction takes about 13s. The results were visually inspected by comparing them to the history table, and are verified to be correct.

## Conclusion
The above reconstruction is too slow for acceptable UX. It surely can be optimized by using more sophisticated grouping, or using multiprocessing.

However bear in mind that the strategy to pre-select boxes of a certain base is not correct. This strategy erroneously finds only boxes that *currently* are in these places. What we want to know though are all boxes that were in these places *at their time of creation* which can only be obtained by processing all stock-related history entries (450k). So even if the runtime above can be optimized it will be 10-20x increased again.

Memory requirements: The resulting reconstructed data for 3800 boxes is about 5MB on disk, probably about the same in memory. In total 94k boxes are registered now, so we can estimate that >100MB of memory required for the reconstruction.

### DIFF
--- a/back/boxtribute_server/setup_wizard.py
+++ b/back/boxtribute_server/setup_wizard.py
@@ -21,7 +21,10 @@ delimiter and double-quote as quote char.
 import argparse
 import csv
 import getpass
+import json
 import logging
+import time
+from datetime import datetime
 
 from boxtribute_server.db import create_db_interface, db
 
@@ -85,6 +88,10 @@ def _parse_options(args=None):
     clone_products_parser.add_argument(
         "-t", "--target-base-id", type=int, required=True
     )
+
+    convert_parser = subparsers.add_parser("convert")
+    convert_parser.add_argument("-l", "--location-id", type=int, required=True)
+    convert_parser.add_argument("-y", "--oldest-creation-year", type=int)
 
     return vars(parser.parse_args(args=args))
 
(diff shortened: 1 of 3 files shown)


HaGuesto:
@jamescrowley would be nice if you could have a read and we could have a quick call.

HaGuesto:
@pylipp very cool that you already wrote it in such a general fashion that we most likely can reuse.

I would drop the ORDER BY in history since that could be very expensive. The thing is that it is already sorted by change_date just not DESC. Can you somehow iterate from the last to the first item instead?

Can you also try out how it improves with some additional indexes in the db? Some of them might be already there.

### Composite Index on (table_name, record_id, change_date):
This index will help with filtering by table_name, record_id, and change_date in your WHERE clause, as well as sorting the results based on change_date. The composite index will improve the query's performance by reducing the number of rows that need to be scanned and allowing efficient sorting.

`CREATE INDEX idx_history_table_record_date ON history (table_name, record_id, change_date);`

### Index on (change_date):
Since you're using change_date in your WHERE clause for filtering and ordering, having a separate index on this column can help improve query performance.

`CREATE INDEX idx_history_change_date ON history (change_date);`

### Index on (table_name, changes):
Although you're not filtering by changes.startswith("comments") directly in the WHERE clause, having an index on table_name and changes can help if your query plan includes a full table scan and then filtering.

`CREATE INDEX idx_history_table_changes ON history (table_name, changes);`

pylipp:
@HaGuesto I added the suggested indexes, and did a run without the order-by clause but there was no visible speedup.

pylipp:
@aerinsol @vahidbazzaz

pylipp:
@aerinsol @vahidbazzaz the involved queries are two simple SQL selects. The reconstruction is Python is the expensive part.

Box-select raw SQL (with placeholders)
```sql
'SELECT `t1`.`id`, `t1`.`box_state_id`, `t1`.`created`, `t1`.`deleted`, `t1`.`items`, `t1`.`location_id`, `t1`.`product_id`, `t1`.`size_id` FROM `stock` AS `t1` WHERE (
(`t1`.`created` > %s) AND (`t1`.`location_id` = %s))', [datetime.datetime(2017, 1, 1, 0, 0), 54]
```

History select
```sql
'SELECT `t1`.`changes`, `t1`.`changedate`, `t1`.`from_int`, `t1`.`record_id` FROM `history` AS `t1` WHERE ((((`t1`.`tablename` = %s) AND (`t1`.`record_id` IN (%s, .......), ([long list of box IDs,  datetime.datetime(2017, 1, 1, 0, 0), 'comments%']
```

```sql

**Comment by aerinsol** [comment]
@pylipp I thought you said there was a join somewhere? I think it's possible to optimize the 2nd query with a CTE or sub-select on BoxIDs, and maybe doing a join.

**Comment by aerinsol** [comment]
@pylipp what are the current table indexes again? I agree with some of @HaGuesto's suggestions, but I strongly disagree with adding all of them, as that will slow performance down by a lot (see reference [here](https://dev.mysql.com/doc/refman/8.0/en/multiple-column-indexes.html) on multi-column indexes - too many of them will mean too many comparisons).  I'm going to guess that if indexes will actually improve performance it will be on JUST `record_id` (and maaaaaybe `change_date`).

https://dev.mysql.com/doc/refman/8.0/en/index-btree-hash.html

**Comment by aerinsol** [comment]
To really see if my guess is right on SQL optimization I will probably need an EXPLAIN plan (paired with a trace, to see if it's really not the SQL.)

**Comment by aerinsol** [comment]
Last comment - can maybe test the performance of different indexes using this feature: https://dev.mysql.com/doc/refman/8.0/en/invisible-indexes.html

File: back/boxtribute_server/setup_wizard.py [hunk 1]
```diff
@@ -192,14 +199,117 @@ def _clone_products(*, source_base_id, target_base_id):
...
+        Box.location,
+        Box.product,
+        Box.size,
+    ).where(
+        Box.created_on > start,
+        Box.location == location_id,
+    )
+    LOGGER.debug(f"Selected {len(boxes)} boxes.")
+
+    # Obtain history entries for selected boxes
```
aerinsol, line 249: requested a structural change to the history table, including a more complete performance trace.

aerinsol (changes requested):
See comments above.

back/boxtribute_server/setup_wizard.py
aerinsol: Github won't let me leave a requested changes review without a line-level comment, so this is where it is.

aerinsol: @HaGuesto @pylipp

pylipp: @aerinsol current indexes of the history table are:
```sql
  KEY `user_id` (`user_id`),
  KEY `recordid_changedate_composite_index` (`record_id`,`changedate`),
  KEY `tablename_recordid_changedate_composite_index` (`tablename`,`record_id`,`changedate`),
```
This means there's already a composite index with `record_id` as leftmost index. I added an index on the `record_id` column nevertheless but there was no change in runtime.

Maybe I should stress again that the expensive part is **not** the DB select (takes about 2-3s) but the reconstruction in Python which takes 14s.

aerinsol: > This means there's already a composite index with record_id as leftmost index. I added an index on the record_id column nevertheless but there was no change in runtime.

@pylipp To be clear, what I was actually proposing was **dropping all other indices except for `record_id`**, OR a composite index of `record_id `and `changedate`, as having multiple concurrent indexes creates performance issues. 

May I ask if you know where the delay in Python is coming from, if not from the processing of the SQL? If there is some other issue I don't believe that this will be solved using shadow tables as proposed in:  docs/adr/adr_backend_etl_and_hosting_for_StatViz.md

pylipp: When reconstructing, I iterate over all boxes, and then iterate over the entire history to find related entries. But obviously on average only 5 entries out of 20k are related to the box of the outer loop, resulting in a large overhead of cancelled runs of the inner loop.

This is not yet tied together with a full data computation for e.g. the `createdBoxes` query, so it's not a definite counter-argument against shadow tables.

aerinsol:
Discussed offline with @pylipp . 

Conclusion: I'm not that concerned with the 20s delay on the initial `SELECT ` statement for full table scan, since I'm quite sure that can be optimized with some indexing / CTE on the query. 

Where I do see concern is the historical walkback - as far as I can tell is that can't be optimized further (though @jamescrowley @jdcsgh and others more experienced in Python can disagree with me). 

The load also can't be moved into SQL with a windowing function since `history` table tracks snapshots of absolute values, not differences. 

As a result, I now agree with the proposal to have a shadow/temporal table as suggested by @jdcsgh and supported by @pylipp and index search on the time.

HaGuesto:
@pylipp is this still needed. I think we can close it, right?

pylipp:
yes @HaGuesto the reconstruction is executed in pure SQL in #1190 , I'm closing this

PR #1014, opened 2023-09-01
ReconciliationView: Change Number of Received Boxes and Total Boxes Count

vahidbazzaz:
- remove all locations other than those setting box state to InStock
- fix received and total boxes count by checking box state


vahidbazzaz:
[2.0 ReconciliationView - Change "Remaining" to "Received", so that the number starts at 0/100 and completes at 100/100](https://trello.com/c/ihgx3w1r/1268-20-reconciliationview-change-remaining-to-received-so-that-the-number-starts-at-0-100-and-completes-at-100-100)

PR #1016, opened 2023-09-03
Fix load bases data before auth0token in apollo header

HaGuesto:
This is the network error pop up that shows up when reloading a page on staging.


HaGuesto:
[BUG: When any page is loaded a NetworkError happens](https://trello.com/c/vr2jcFqI/1275-bug-when-any-page-is-loaded-a-networkerror-happens)

pylipp (approved):
## Functional review
I tested locally and didn't see any 401 error when refreshing a page.

PR #1024, opened 2023-09-04
Add GraphQL boxes query

pylipp:
Test by running the new query with various filters, e.g.

```graphql
query { boxes(baseId: 1) { totalCount } }
query { boxes(baseId: 1, filterInput: {productId: 1}) { totalCount } }
query { boxes(baseId: 1, filterInput: {productId: 1, states: [InStock]}) { totalCount } }
```

and verifying the numbers via direct querying from MySQL.


back/boxtribute_server/graph_ql/filtering.py
```diff
@@ -73,9 +79,34 @@ def derive_box_filter(filter_input):
...
+    if tag_ids is not None:
+        selection = selection.join(
+            TagsRelation,
+            src=Box,
+            on=(
+                (TagsRelation.object_type == TaggableObjectType.Box)
+                & (TagsRelation.object_id == Box.id)
+                & (TagsRelation.tag << tag_ids)
+            ),
+        )
```
pylipp, line 107: todo: check if this needs distinct

HaGuesto: Yes, you will need to change something here. This query returns duplicate results:
```
{boxes(
  baseId:1, 
  filterInput:{
    tagIds: [4,5]
    
  }) {
    elements {id}}}
    ```

HaGuesto: There are two ways to solve this. See case 4.) in this [link](https://learnsql.com/blog/get-rid-of-duplicates-sql-join/)

HaGuesto (approved):
@pylipp code review passes. There is only issue with tag filtering as you expected yourself regarding feature review. I also tried to extract Boxes from another base and it did not work.

I'll merge this already because we need this for the BoxesView and I'm asking you to create a new ticket and PR for the tag filtering.

pylipp:
Follow-up: https://github.com/boxwise/boxtribute/pull/1062

PR #1032, opened 2023-09-15
Set ShipmentDetail.received_by/on immediately after box reconciliation

pylipp:
Can be tested by reconciling a box via the GraphQL playground.

### DIFF
--- a/back/boxtribute_server/business_logic/box_transfer/shipment/crud.py
+++ b/back/boxtribute_server/business_logic/box_transfer/shipment/crud.py
@@ -302,6 +302,7 @@ def _update_shipment_with_received_boxes(
         for i in shipment_detail_update_inputs or []
     }
 
+    now = utcnow()
     updated_box_fields = [
         Box.product,
         Box.location,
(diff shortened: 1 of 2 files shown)


pylipp:
[2.0 Set ShipmentDetail.received_by/on immediately after box reconciliation](https://trello.com/c/Douee3Mb/1288-20-set-shipmentdetailreceivedby-on-immediately-after-box-reconciliation)

vahidbazzaz (approved):
Thank you @pylipp, this works as expected, once I reconciled the box, it set the received_on and receive_by on the database. 

@pylipp @aerinsol If you agree, I will push related FE changes to this PR so that in the reconciliation UI box counts are based on received_on rather than boxes' states

pylipp:
@vahidbazzaz is there any advantage of using box states instead of timestamps for counting? Otherwise I find it easier to read code logic that deals with box states.

vahidbazzaz:
> @vahidbazzaz is there any advantage of using box states instead of timestamps for counting? Otherwise I find it easier to read code logic that deals with box states.

@pylipp yes indeed, the count can be incorrect if the box state changes while the other boxes are still reconciling.

pylipp:
@vahidbazzaz sure go ahead!

PR #1036, opened 2023-09-18
[4th viz] Implement GraphQL interface for movedBoxes

pylipp:
This PR brings the GraphQL interface for data about moved boxes.

It can be used to generate a Sankey diagram of the number of boxes donated from a base into "donation" locations, filterable by movement day and product category.

### DIFF
--- a/back/boxtribute_server/business_logic/statistics/crud.py
+++ b/back/boxtribute_server/business_logic/statistics/crud.py
@@ -222,11 +222,19 @@ def compute_moved_boxes(base_id):
     """
     # Similar to example from
     # https://docs.peewee-orm.com/en/latest/peewee/relationships.html#subqueries
+    # Subquery to select record IDs and latest dates when box state was changed from
+    # InStock to Donated.
     LatestMoved = DbChangeHistory.alias()
     LatestMovedSubQuery = (
         LatestMoved.select(
             LatestMoved.record_id,
-            fn.MAX(LatestMoved.change_date).alias("max_change_date"),
+            fn.MAX(LatestMoved.change_date).alias("move_date"),
+        )
+        .where(
+            (LatestMoved.table_name == Box._meta.table_name),
+            (LatestMoved.changes == Box.state.column_name),
+            (LatestMoved.from_int == BoxState.InStock),
+            (LatestMoved.to_int == BoxState.Donated),
         )
         .group_by(LatestMoved.record_id)
         .alias("sq")
@@ -247,18 +255,12 @@ def compute_moved_boxes(base_id):
             LatestMovedSubQuery,
             on=(
                 (DbChangeHistory.record_id == LatestMovedSubQuery.c.record_id)
-                & (DbChangeHistory.change_date == LatestMovedSubQuery.c.max_change_date)
+                & (DbChangeHistory.change_date == LatestMovedSubQuery.c.move_date)
             ),
         )
         .join(
             Box,
-            on=(
-                (DbChangeHistory.record_id == Box.id)
-                & (DbChangeHistory.table_name == Box._meta.table_name)
-                & (DbChangeHistory.changes == Box.state.column_name)
-                & (DbChangeHistory.from_int == BoxState.InStock)
-                & (DbChangeHistory.to_int == BoxState.Donated)
-            ),
+            on=((DbChangeHistory.record_id == Box.id)),
             src=DbChangeHistory,
         )
         .join(
--- a/back/boxtribute_server/business_logic/statistics/crud.py
+++ b/back/boxtribute_server/business_logic/statistics/crud.py
@@ -4,6 +4,7 @@ from peewee import JOIN, SQL, fn
 
 from ...db import db
 from ...enums import BoxState, HumanGender, TaggableObjectType
+from ...models.definitions.base import Base
 from ...models.definitions.beneficiary import Beneficiary
 from ...models.definitions.box import Box
 from ...models.definitions.history import DbChangeHistory
@@ -47,6 +48,20 @@ def _generate_dimensions(*names, facts):
         dimensions["size"] = (
             Size.select(Size.id, Size.label.alias("name"))
             .where(Size.id << size_ids)
+            .dicts()
+        )
+
+    if "base" in names:
+        base_ids = {f["base_id"] for f in facts}
+        dimensions["base"] = (
+            Base.select(Base.id, Base.name).where(Base.id << base_ids).dicts()
+        )
+
+    if "location" in names:
+        location_ids = {f["location_id"] for f in facts}
+        dimensions["location"] = (
+            Location.select(Location.id, Location.name)
+            .where(Location.id << location_ids)
             .dicts()
         )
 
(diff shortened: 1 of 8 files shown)


back/boxtribute_server/graph_ql/definitions/public_api.graphql
```diff
@@ -84,6 +85,35 @@ type TopProductsDimensions {
   category: [DimensionInfo]
 }
 
+type MovedBoxesData implements DataCube {
+  facts: [MovedBoxesResult]
+  dimensions: MovedBoxDataDimensions
+}
+
+"""
+A box can be moved in various ways:
```
HaGuesto, line 94: I analyzed the history table a bit andd what changes actually exist in there that are stock related. Here, the result:
https://docs.google.com/spreadsheets/d/1CLqyYBgCxQ1s5R9h8ySAUPoDqCqXeyoH8dQhzmfVc18/edit?usp=sharing

(Work In Progress)
@pylipp

back/boxtribute_server/graph_ql/definitions/public_api.graphql
```diff
@@ -84,6 +85,35 @@ type TopProductsDimensions {
...
+- within a base (location ID with InStock/Donated)
+- because it's lost (Lost)
+- because it becomes scrap (Scrap)
+- because it's about to be shipped (target base ID with MarkedForShipment)
+- because it's being shipped (target base ID with InTransit/Receiving)
+"""
+type MovedBoxesResult {
+  movedOn: Date!
+  boxState: BoxState!
+  locationId: Int
```
HaGuesto, line 104: I would merge location and base into one return value(maybe called "target")

back/boxtribute_server/graph_ql/definitions/public_api.graphql
```diff
@@ -84,6 +85,35 @@ type TopProductsDimensions {
...
+A box can be moved in various ways:
+- within a base (location ID with InStock/Donated)
+- because it's lost (Lost)
+- because it becomes scrap (Scrap)
+- because it's about to be shipped (target base ID with MarkedForShipment)
+- because it's being shipped (target base ID with InTransit/Receiving)
+"""
+type MovedBoxesResult {
+  movedOn: Date!
+  boxState: BoxState!
```
HaGuesto, line 103: I don't get what this is doing here.

back/boxtribute_server/graph_ql/definitions/public_api.graphql
```diff
@@ -84,6 +85,35 @@ type TopProductsDimensions {
...
+
+"""
+A box can be moved in various ways:
+- within a base (location ID with InStock/Donated)
+- because it's lost (Lost)
+- because it becomes scrap (Scrap)
+- because it's about to be shipped (target base ID with MarkedForShipment)
+- because it's being shipped (target base ID with InTransit/Receiving)
+"""
+type MovedBoxesResult {
```
HaGuesto, line 101: maybe we should add something like a moveType to distinguish between shipments and Box moves

back/boxtribute_server/graph_ql/definitions/public_api.graphql
```diff
@@ -84,6 +85,35 @@ type TopProductsDimensions {
   category: [DimensionInfo]
 }
 
+type MovedBoxesData implements DataCube {
```
HaGuesto, line 88: I would rename it. MovedBoxes is a bit misleading. Since we only take a look at how Boxes are moved out of the warehouse and we neglect internal moves.

HaGuesto: outgoingBoxes

HaGuesto:
## Meeting with @pylipp @MaikNeubert 

AGREED:
`
type MovedBoxesResult {
  movedOn: Date!
  targetId: ID!
  categoryId: Int!
  boxesCount: Int!
}

type MovedBoxDataDimensions {
  target: [TargetDimensionInfo]
  category: [DimensionInfo]
}

type TargetDimensionInfo {
  id: 
  label: Base Name for shipment and Location Name for OutgoingLocation
  type:  enum<Shipment, OutgoingLocation>
}
`
AGREED: Keep it simple, leave out SCRAP, LOST, shipment states for now

back/boxtribute_server/business_logic/statistics/crud.py
```diff
@@ -206,3 +215,53 @@ def compute_top_products_donated(base_id):
...
+        .join(
+            Product,
+            on=((Box.product == Product.id) & (Product.base == base_id)),
+        )
+        .join(
+            Location,
+            src=Box,
+            on=((Box.location == Location.id) & (Location.base == base_id)),
+        )
+        .join(Base)
```
HaGuesto, line 255: don't you need an `on` clause here?

pylipp: peewee is able to auto-derive this because the previous model (Location) has a single FK reference to the Base model.

pylipp:
referencing discussion from [Slack](https://boxwise.slack.com/archives/CB1UYDGLU/p1696269030310409)

pylipp:
@MaikNeubert I tested it with Hermine's base. Target dimension now looks like below, let me know if this is handy or whether you want base and location name to be listed separately.
```
{
  "data": {
    "movedBoxes": {
      "dimensions": {
        "target": [
          {
            "name": "Würzburg---V85 -JNCR Italy"
          },
...
```

back/boxtribute_server/business_logic/statistics/crud.py
```diff
@@ -206,3 +215,55 @@ def compute_top_products_donated(base_id):
...
+    movement, product category, and box state.
+    """
+    # This selects only information of boxes that were moved from InStock to Donated
+    # state, and are now in the base of given base ID. It is NOT taken into account that
+    # boxes can be moved back from Donated to InStock, nor that the product or other
+    # attributes of the box change after having been donated
+    selection = (
+        DbChangeHistory.select(
+            fn.MAX(DbChangeHistory.change_date).alias("moved_on"),
+            fn.CONCAT(Base.name, TARGET_ID_SEPARATOR, Location.name).alias("target_id"),
```
HaGuesto, line 234: Why do you add the Base name for the target_id? I would just do the Location.name for Outgoing Locations and the Base.name for shipments.

back/boxtribute_server/business_logic/statistics/crud.py
```diff
@@ -206,3 +215,55 @@ def compute_top_products_donated(base_id):
...
+        .join(
+            Location,
+            src=Box,
+            on=((Box.location == Location.id) & (Location.base == base_id)),
+        )
+        .join(Base)
+    )
+    facts = selection.group_by(
+        SQL("target_id"),
+        SQL("category_id"),
```
HaGuesto, line 261: @pylipp Shouldn't you also group by "movedOn"? I'm not really familiar in peewee, but in SQL I would have to do that.

HaGuesto (changes requested):
@pylipp not sure if I agree with the target_id format. I made another proposal there and I had one additional question that you could maybe answer.

Thank you @pylipp

back/boxtribute_server/business_logic/statistics/crud.py
pylipp: makes sense :+1:

back/boxtribute_server/business_logic/statistics/crud.py
pylipp: When I do that the query errors with `1056 - Can't group on 'moved_on'`. I think this is because the `moved_on` column is derivate by the MAX aggregate.

HaGuesto: ahhh, yes @pylipp . was already wondering cause the results were also off.

So, you have to do two sequential group bys - one group by in a subquery over record_id in history to get the movedOn date and then one group by over movedOn, target_id and category_id.

MaikNeubert (approved):

back/boxtribute_server/business_logic/statistics/crud.py
HaGuesto: @pylipp almost there, but not completely. 
`moved_on` in your query is the date of the last entry of the corresponding `record_id` in history.
What you need is the last date of the corresponding `record_id` WHERE the tablename is stock, the change is Box_state_id, from_to=......

Therefore, you will need to move this ON clause  `AND (`t1`.`tablename` = %s)) AND (`t1`.`changes` = %s)) AND (`t1`.`from_int` = %s)) AND (`t1`.`to_int` = %s))`
to a WHERE clause in the subquery sq.

HaGuesto (approved):

PR #1038, opened 2023-09-21
BoxView: Add History Overlay  

vahidbazzaz:
The PR creates a global timeline component and adds a history overlay to the box view.

- create global timeline component
- create history overlay component
- make helper function for preparing box history entry
- add history overlay in the box view
- refactors the shipment history view to use a global timeline


pylipp (approved):
## Functional review
- when more than one history item is there, the button for history overlay shows and can be clicked. The overlay opens as expected, and is scrollable
- please fix the time format (e.g. 11:07 shows as 11:7)
- I found an irregularity: when I go to BoxEdit for a box that has no comment yet, and update e.g. tags there, save and return to BoxView, then the history shows "comments changed from "None" to "" " because None/null is the default in the stock table; but the FE sends "" (empty string) from the form, and then the BE detects and logs this as a change. If you know a simple way to make the FE send null instead of "" with the `updateBox` mutation, that'd be awesome; otherwise I'll adjust the BE in a separate PR

Leaving code review to @HaGuesto

HaGuesto:
## Functional review
All mentions from @pylipp were fixed.

> I found an irregularity: when I go to BoxEdit for a box that has no comment yet, and update e.g. tags there, save and return to BoxView, then the history shows "comments changed from "None" to "" " because None/null is the default in the stock table; but the FE sends "" (empty string) from the form, and then the BE detects and logs this as a change. If you know a simple way to make the FE send null instead of "" with the updateBox mutation, that'd be awesome; otherwise I'll adjust the BE in a separate PR

I saw this problem already and I believe I fixed that behavior [here](https://github.com/boxwise/boxtribute/commit/c36e4bc2262541f01f87af2152f8ec01a811fccd)
Probably we should also check BoxCreate (not in this PR though. I will create a ticket for it)

Otherwise, I have two more comments regarding functional review and maybe @aerinsol should decide here:
- The red and the blue does not look like the colors from our brand (Is that the color we want to have? What are they?)
- the points are not centered on the line. Is that wanted?
image

react/src/components/Timeline/TimelineContainer.tsx
```diff
@@ -0,0 +1,55 @@
+import { Box, Flex, Text } from "@chakra-ui/react";
+import { User } from "types/generated/graphql";
+import TimelineEntry from "./components/TimelineEntry";
+
+export interface ITimelineEntry {
+  action: string | undefined;
```
HaGuesto, line 6: I'd suggest try to get rid of these undefined items here and handle these cases before you pass the data in the component. In general, that would be just cleaner.

react/src/components/Timeline/components/TimelineEntry.tsx
```diff
@@ -3,10 +3,21 @@ import TimelineBullet from "./TimelineBullet";
 
 export interface ITimelineEntryProps {
   content: string | undefined;
-  time: string | undefined;
+  time: string | Date | undefined;
 }
 
 function TimelineEntry({ content, time }: ITimelineEntryProps) {
+  function formatTime(date: Date | string): string {
```
HaGuesto, line 10: could you pull this out into helpers. Otherwise, this function is defined in each rerender multiple times.

react/src/views/Box/BoxView.tsx
```diff
@@ -169,6 +175,38 @@ function BTBox() {
...
+  const boxLogs: ITimelineEntry[] = (allData.data?.box?.history as HistoryEntry[])?.flatMap(
+    (histories) =>
+      _.compact([
+        histories?.user && {
+          action: prepareBoxHistoryEntryText(`${histories.user.name} ${histories.changes}`),
+          createdBy: histories.user as User,
+          createdOn: new Date(histories.changeDate),
+        },
+      ]),
+  ) as unknown as ITimelineEntry[];
```
HaGuesto, line 188: could you try that the data that is returned here, does not include any unknowns, undefined or similar. As mentioned above it makes the definition of the UI component cleaner and easier.

react/src/views/Box/BoxViewHistoryOverlay.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+    },
+  },
+};
+
+// Test case 3.1.12
+describe("3.1.12 - Box HistoryOverlay on BoxView", () => {
+  beforeEach(() => {
+    // we need to mock matchmedia
+    // https://jestjs.io/docs/manual-mocks#mocking-methods-which-are-not-implemented-in-jsdom
+    Object.defineProperty(window, "matchMedia", {
```
HaGuesto, line 49: we have a helper for this somewhere already.

react/src/components/HistoryOverlay/HistoryOverlay.tsx
```diff
@@ -0,0 +1,40 @@
...
+  onClose: () => void;
+}
+
+function HistoryOverlay({ data, isOpen, onClose }: IHistoryOverlay) {
+  return (
+    <Modal
+      isOpen={isOpen}
+      onClose={onClose}
+      blockScrollOnMount={false}
+      size="3xl"
```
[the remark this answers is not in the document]
vahidbazzaz: @HaGuesto If the overlay expands, the default behavior is to scroll the whole page, but we want to scroll the overlay instead.

react/src/views/Box/BoxViewHistoryOverlay.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+        dispatch: jest.fn(),
+        globalPreferences: {
+          organisation: { id: organisation1.id, name: organisation1.name },
+          availableBases: organisation1.bases,
+          selectedBase: organisation1.bases[0],
+        },
+      },
+    });
+
+    await waitFor(async () => {
```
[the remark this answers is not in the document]
vahidbazzaz: @HaGuesto I refactored the tests here, but waitFor is used elsewhere which we'll need to refactor at some point as well.

PR #1039, opened 2023-09-21
QrOverlay: remove minus and plus buttons

vahidbazzaz:
In this PR, the minus and plus buttons are removed from the QR reader overlay.


vahidbazzaz:
[2.0 Drop the +/- button in the QR Scanner, it's confusing our users.](https://trello.com/c/hsdBdCQ4/1287-20-drop-the-button-in-the-qr-scanner-its-confusing-our-users)

pylipp (approved):

PR #1059, opened 2023-10-09
Add shipment label identifier

pylipp:
This PR adds the `Shipment.labelIdentifier` field, and integrates it in the front-end.

An identifier looks like `S078-231009-BExAT`.

I used a pragmatic approach that does not require any additional DB field. The label identifier is created from "static", neverchanging data on-the-fly when the `labelIdentifier` field is requested. Not having another DB field saves time for writing the migration, updating the seed, updating the test data, ... for information that is actually redundant.

Also I don't think we have to update any queries in the FE since they can just use the `Shipment.id` field for internal purposes like caching or value-name listing.

Let me know if you want to change anything.


pylipp:
[2.0 Add shipment identifier](https://trello.com/c/wqk9wlNG/1295-20-add-shipment-identifier)

back/boxtribute_server/business_logic/box_transfer/shipment/fields.py
```diff
@@ -1,13 +1,36 @@
...
 from ....models.definitions.shipment import Shipment
 from ....models.definitions.shipment_detail import ShipmentDetail
 
 shipment = ObjectType("Shipment")
 shipment_detail = ObjectType("ShipmentDetail")
 
 
+@shipment.field("labelIdentifier")
+def resolve_shipment_label_identifier(shipment_obj, _):
+    def first_letters_of_base_name(base_id):
```
HaGuesto, line 16: does it make a performance difference if the function is in the resolver and not in a helper file? In the FE there are cases where you not try to do that since the function definition would be applied each time something is re-rendered.

HaGuesto (approved):
## Functional review
@pylipp generally the review passes  

However, @vahidbazzaz @aerinsol There are some points we should address (not in this PR):
1. How should the options in the BoxView assignToShipment feature look like @aerinsol ?
image

2. Same question @aerinsol for QrReader MultiBox Action.
image

3. In the shipmentCard there title is now generally overflowing. @vahidbazzaz can you fix this in a new PR? @aerinsol is there input from you how it should look like?
image

aerinsol:
I have quick answers for all of these, plus we also need to add a column for ID to the "Manage Shipments" table. @vahidbazzaz will scope the difficulty of creating dropdown with smaller subtext under the main one. I can do mockup if needed.

back/boxtribute_server/business_logic/box_transfer/shipment/fields.py
pylipp: Fair point, the function would be defined everytime the resolver is called which can be multiple times per request.

back/boxtribute_server/business_logic/box_transfer/shipment/fields.py
```diff
@@ -1,13 +1,36 @@
...
+        base = Base.get_by_id(base_id)
+        return base.name[:2].upper()
```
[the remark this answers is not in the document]
pylipp: Python3 strings are utf-8 by default, so I don't see issues with umlauts. I've added unit tests though and found an edge case

pylipp:
to be superseded by https://trello.com/c/U9mhtkZz/1298-20-implement-new-shipmentid-in-fe

PR #1060, opened 2023-10-09
Emit warning to Sentry when multi-base agreement created

pylipp:
Tested with BE running locally, logged [here](https://boxwise.sentry.io/issues/4535394176/?project=6584296&query=is%3Aunresolved&referrer=issue-stream&stream_index=0) (I had to manually add the DSN when initializing the SDK).

This will not be effective during local testing.


pylipp:
[2.0 BE: notify when multi-base agreement created](https://trello.com/c/HBzOhLf6/1254-20-be-notify-when-multi-base-agreement-created)

PR #1061, opened 2023-10-09
[1st viz] Fix beneficiary demographics for production data

pylipp:
Fixing issues I found after testing with production data.

### DIFF
--- a/back/boxtribute_server/business_logic/statistics/crud.py
+++ b/back/boxtribute_server/business_logic/statistics/crud.py
@@ -66,7 +66,9 @@ def compute_beneficiary_demographics(base_ids=None):
     """
     gender = fn.IF(Beneficiary.gender == "", "D", Beneficiary.gender)
     created_on = db.database.truncate_date("day", Beneficiary.created_on)
-    age = compute_age(Beneficiary.date_of_birth)
+    age = fn.IF(
+        Beneficiary.date_of_birth > 0, compute_age(Beneficiary.date_of_birth), None
+    )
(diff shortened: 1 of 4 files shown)


back/boxtribute_server/business_logic/statistics/crud.py
```diff
@@ -59,13 +59,14 @@ def compute_beneficiary_demographics(base_ids=None):
     results as list.
     The 'age' dimensions actually represents a range of ages (e.g. 0-5, 5-10, etc.)
     """
-    bin_width = 5
     gender = fn.IF(Beneficiary.gender == "", "D", Beneficiary.gender)
     created_on = db.database.truncate_date("day", Beneficiary.created_on)
-    age = fn.FLOOR((date.today().year - Beneficiary.date_of_birth.year) / bin_width)
+    # Age calculation is an approximation (if the current time is June, anyone born in a
+    # later month will have an age a year older than they actually are)
+    age = date.today().year - Beneficiary.date_of_birth.year
```
HaGuesto, line 66: What is the implication of not making it an approximation? I don't see that a lot more code is needed to make this exact. So, I don't understand why you are introducing an inaccuracy here.

HaGuesto (changes requested):
@pylipp Are you returning the data of people who do not have a date_of_birth set in the data? Maybe a good question for tomorrow's SU with @MaikNeubert and @aerinsol how to treat that.

Ohterwise, I have one more question.

back/boxtribute_server/business_logic/statistics/crud.py
pylipp: fixed it :+1: used similar code in another place

PR #1062, opened 2023-10-10
Avoid returning duplicate boxes when filtering for tag IDs

pylipp:
Update test data to verify


pylipp:
[2.0 Avoid returning duplicate boxes when filtering for tag IDs](https://trello.com/c/c6vXrkKK/1296-20-avoid-returning-duplicate-boxes-when-filtering-for-tag-ids)

HaGuesto (approved):
works. Thanks for the quick update @pylipp

PR #1102, opened 2023-11-04
[4th viz] Extend movedBoxes data with info about shipped/lost/scrap boxes

pylipp:
This PR brings two new target types ("Shipment" and "BoxState") and collect info about shipped boxes, and lost/scrap boxes, resp.) into the result of the `movedBoxes` query.

Test by connecting with prod DB and running the FE for Moved Boxes.

### Explanation

Target type | movedOn | targetId
--- | --- | ---
Shipment | timestamp of when the shipment was sent | name of target base
BoxState | timestamp of when the box was set to Lost/Scrap state | name of box state

### DIFF
--- a/back/test/endpoint_tests/test_statistics.py
+++ b/back/test/endpoint_tests/test_statistics.py
@@ -1,3 +1,5 @@
+from datetime import date
+
 from boxtribute_server.enums import ProductGender, TargetType
 from boxtribute_server.models.utils import compute_age
 from utils import assert_successful_request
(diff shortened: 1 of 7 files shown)


back/test/endpoint_tests/test_statistics.py
```diff
@@ -159,29 +161,41 @@ def test_query_top_products(
...
                 "movedOn": "2022-12-05",
             },
+            {
+                "boxesCount": 5,
+                "categoryId": 1,
+                "targetId": base_name,
+                "movedOn": date.today().isoformat(),
+            },
         ],
         "dimensions": {
```
pylipp, line 187: todo: I should add test data such that a lost/scrap box appears here

MaikNeubert (approved):
I tested it on a few locations and it looks good to me :)

PR #1104, opened 2023-11-07
Enable DB replica

pylipp:
I implemented a custom class derived from peewee's `FlaskDB` to glue together Flask and the databases (the primary one and, for select-queries, the replica).

Further changes are about cleaning up env variables related to configure the DB connection locally and in CircleCI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -248,6 +248,7 @@ jobs:
               MYSQL_DB=${DB_NAME}
               MYSQL_PORT=3306
               MYSQL_SOCKET=${DB_SOCKET}
+              MYSQL_REPLICA_SOCKET=${DB_REPLICA_SOCKET}
               SENTRY_DSN=${SENTRY_BE_DSN}
               SENTRY_ENVIRONMENT=${ENVIRONMENT}
               SENTRY_RELEASE=${CIRCLE_SHA1}
@@ -320,6 +321,7 @@ jobs:
               MYSQL_DB=${DB_NAME}
               MYSQL_PORT=3306
               MYSQL_SOCKET=${DB_SOCKET}
+              MYSQL_REPLICA_SOCKET=${DB_REPLICA_SOCKET}
               SENTRY_DSN=${SENTRY_BE_DSN}
               SENTRY_ENVIRONMENT=${ENVIRONMENT}
               SENTRY_RELEASE=${CIRCLE_SHA1}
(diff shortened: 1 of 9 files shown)


HaGuesto:
@HaGuesto priority

.circleci/config.yml
```diff
@@ -193,7 +193,8 @@ jobs:
           name: Run pytest
           command: |
             source $(pyenv root)/versions/3.10.4/envs/env/bin/activate
-            ENVIRONMENT=test pytest --cov-report xml:test-results/coverage.xml --cov=boxtribute_server
+            # overwrite environment variable because it must not depend on the CircleCI context
+            ENVIRONMENT=development MYSQL_PORT=3306 pytest --cov-report xml:test-results/coverage.xml --cov=boxtribute_server
```
HaGuesto, line 197: weird that the socket is needed here if you are using a socket

back/boxtribute_server/app.py
```diff
@@ -65,11 +78,15 @@ def before_sentry_send(event, hint):
...
+        # always used
         user=os.environ["MYSQL_USER"],
         password=os.environ["MYSQL_PASSWORD"],
         database=os.environ["MYSQL_DB"],
+        # used for connecting to development / CI testing DB
+        host=os.getenv("MYSQL_HOST"),
+        port=int(os.getenv("MYSQL_PORT", 0)),
+        # used for connecting to Google Cloud from GAE
         unix_socket=os.getenv("MYSQL_SOCKET"),
+        replica_socket=os.getenv("MYSQL_REPLICA_SOCKET"),
```
HaGuesto, line 90: We need to remember to set this variable in the contexts in circleci

back/boxtribute_server/db.py
```diff
@@ -1,15 +1,61 @@
...
+    def get_model_class(self):
+        """Whenever a database model (representing a table) is defined, it must derive
+        from `db.Model` which calls this method.
+        """
+        if self.database is None:
+            raise RuntimeError("Database must be initialized.")
+
+        class BaseModel(self.base_model_class):
+            @classmethod
+            def select(cls, *args, **kwargs):
```
HaGuesto, line 38: elegant

HaGuesto (approved):
I like your solution, feels elegant. @pylipp 

Just one comment: we started actually that we do not want the statviz queries to have an effect or block behavior in Boxtribute. Since now the replica is used for ALL querys, we just ensured that the statviz queries will not have an effect on the mutations in Boxtribute.

However, we also started at the point that we want to make these queries public and that is not happening at least for now. Therefore, I'm fine with this PR and it will be a nice test.

back/boxtribute_server/app.py
pylipp: I already set it as a project-wide variable

.circleci/config.yml
pylipp: What do you mean? weird that the _port_ is needed? It's because the connection to the DB for integration tests is via TCP. Socket will only be effective in deployed environments.

PR #1107, opened 2023-11-10
Sankey moved boxes

MaikNeubert:
Sankey base diagramm is finished, aligned with other graphs, filtered by the selected timerange and exportable as jpg/svg  

!To be reviewed after the image export PR

chart(27)


aerinsol:
@MaikNeubert can you make the diagram list number and/or percentages of each branch vs. the total?

PR #1115, opened 2023-11-16
Resolve qr hash forwarded from dropapp

HaGuesto:
@pylipp and @aerinsol:

## What has changed?
Since we disabled the [mobile part of dropapp](https://github.com/boxwise/dropapp/pull/566), there is a bit of routing that moved to v2 for the case that a user scans a QR-code with an external QR-Reader.
In this case, the external QR-Reader requests the mobile.php url from dropapp which now just forwards the qr hash to v2. The business logic if BoxView should be shown or BoxCreate or an Error lies no in v2.

## Test cases
Generally, I would like to ask you both to test most of the test cases of 3.4.8 of the FE functional tests of the test plan.
You can use the following hashes from the staging seed for the tests:

- QR-Code without Box: '168df347206ed26cd2213c0e4767ccb'
- QR-Code with Box in base 1: 'db7d22cebdc57632bfc0a23a504424b'
- QR-Code with Box in base 2: '0a85f8d3bc751ce1469bd357e56e922'
- QR-Code with Box in base 3: 'dc89000023e5fe7148bfd77bc9ddbfb'
- QR-Code connected to soft-deleted Box in base 1: You have to delete a box first through dropapp or set the deleted value in stock for a box
- legacy QR-Code connected to Box in base 1: I do not know yet how to recreate this, but I checked production and there are only 5 InStock Boxes left in Nea Kavala that are still relevant for this case.

## How to test
@pylipp: If you don't wanna run dropapp in parallel, I'd suggest you always start with requesting `/qrreader/<hash>` and use a hash that is true for the test cases in 3.4.8.

@aerinsol: Once it is merged and deployed to master, I would like you to always start from dropapp and also check if the redirecting works properly. --> start your tests with requesting `staging.boxtribute.org/mobile.php?barcode=<hash>`
Of course, you can also just scan then qr-codes generated and used on staging.

### DIFF
--- a/front/src/App.tsx
+++ b/front/src/App.tsx
@@ -27,6 +27,7 @@ import ShipmentView from "views/Transfers/ShipmentView/ShipmentView";
 import QrReaderView from "views/QrReader/QrReaderView";
 import NotFoundView from "views/NotFoundView/NotFoundView";
 import { useAuthorization } from "hooks/useAuthorization";
+import ResolveHash from "views/QrReader/components/ResolveHash";
 
 interface IProtectedRouteProps {
   minBeta: number;
@@ -65,20 +66,13 @@ function App() {
   return (
     <Routes>
       <Route index />
-      <Route path="bases" element={<Layout />}>
+      <Route path="bases">
         <Route index />
-        <Route path=":baseId">
+        <Route path=":baseId" element={<Layout />}>
           <Route index element={<BaseDashboardView />} />
-          <Route path="transfers" element={<ProtectedRoute minBeta={2} redirectPath="/qrreader" />}>
-            <Route path="agreements">
-              <Route index element={<TransferAgreementOverviewView />} />
-              <Route path="create" element={<CreateTransferAgreementView />} />
-            </Route>
-            <Route path="shipments">
-              <Route index element={<ShipmentsOverviewView />} />
-              <Route path="create" element={<CreateShipmentView />} />
-              <Route path=":id" element={<ShipmentView />} />
-            </Route>
+          <Route path="qrreader">
+            <Route index element={<QrReaderView />} />
+            <Route path=":hash" element={<ResolveHash />} />
           </Route>
           <Route path="boxes">
             <Route index element={<Boxes />} />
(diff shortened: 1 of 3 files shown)


pylipp:
I logged in as coordinator@samos.org (base 3), and used the codes below with the `/qrreader/<hash>` 

> * QR-Code without Box: '168df347206ed26cd2213c0e4767ccb'

I'm redirected to CreateBox page as expected.

> * QR-Code with Box in base 1: 'db7d22cebdc57632bfc0a23a504424b'

Staying on /qrreader with the error message about not being permitted to view the box as expected.

> * QR-Code with Box in base 2: '0a85f8d3bc751ce1469bd357e56e922'

Staying on /qrreader with the error message about not being permitted to view the box as expected.

> * QR-Code with Box in base 3: 'dc89000023e5fe7148bfd77bc9ddbfb'

Directed to BoxView as expected (http://localhost:3000/bases/3/boxes/4884878)

> * QR-Code connected to soft-deleted Box in base 1: You have to delete a box first through dropapp or set the deleted value in stock for a box

I set the deleted-value for the box with QR code `dc89000023e5fe7148bfd77bc9ddbfb` and I'm directed to BoxView. Not sure what's supposed to happen since nothing's written in the test plan either.

> * legacy QR-Code connected to Box in base 1: I do not know yet how to recreate this, but I checked production and there are only 5 InStock Boxes left in Nea Kavala that are still relevant for this case.

Did not test this.

I also tested "User scans QR code where hash is not found in db" and "User scans QR code and server returns unexpected error" and they work as expected.

Approved :+1:

pylipp (approved):

HaGuesto:
[2.0 ResolveHash forwarded from dropapp (use case of scannning with external QR-Reader)](https://trello.com/c/yQtKJUko/1311-20-resolvehash-forwarded-from-dropapp-use-case-of-scannning-with-external-qr-reader)

aerinsol:
OK. I just tested this by generating a few new QR codes after creating a new base (Boxaid - Malta). When trying to scan those, I got an unspecified error message "error - something went wrong!". This is for God Admin.

I am going to guess the staging seed will reset soon so I will let you try and replicate it on your own in staging. @HaGuesto

aerinsol:
@HaGuesto I'm a bit confused. When a box is deleted in Dropapp, I'll get a 403 on the desktop app: (Sentry #2d06c7f2ee0e49a6b6d9540287516c08). However, if I scan in on an external scanner it will load. Not sure if that is what you intended to do.

PR #1116, opened 2023-11-17
BoxesActionsAndTable: Implement Un-/Assign to Shipment Action

vahidbazzaz:
- add assign and unassign action buttons and wire up hook
- add transformer for shipment options
- add options to dropdown and finalise assign/unassign handlers
- add subtitle to action button
- fix display of incorrect total of unassigned boxes
- finalise un-/assign boxes to a shipment


front/src/views/Boxes/components/BoxesActionsAndTable.tsx
```diff
@@ -38,54 +54,251 @@ function BoxesActionsAndTable({
     navigate(`/bases/${baseId}/boxes/${labelIdentifier}`);
 
   // --- Actions on selected Boxes
```
HaGuesto, line 56: @aerinsol from here until line 260 you have the texts that are showing for the alerts.

pylipp (approved):
## Functional review
Overall, no blockers from my side that cannot be fixed in a follow-up PR :stuck_out_tongue: biggest one is probably the size-filtering one. But let's bring this into master

- I have a 800x1300 notebook screen that does not fit the entire table width. I can horizontal-scroll but it's a bit weird that the horizontal scroll bar is hidden at the bottom of the table but the vertical one is visible
- first column is called "Box number" but in dropapp it's "Box ID". Personally I find "Box label" okay but it depends on what's used by users in practice
- filter by box state works though the state names for the filters could be in PascalCase instead of UPPERCASE
- filter by size does not work as expected: Filtering by size S returns all boxes with sizes that contain the letter S
- filter by product: dropdown filter list could be alphabetically sorted
- showing/hiding columns works. But it's not being remember after I reload the page (not sure if that's intended. I think it's quite useful especially when going back and forth between BoxesView and individual BoxViews)
- column names are center aligned. When I hide a lot of columns, it looks a bit off (especially the select-all check-box moves, and is not aligned anymore with the rows checkboxes)
- when selecting a row and clicking "Move to Location", the move is executed. When I check the corresponding BoxView, the location is updated but there is no history entry displayed

Tested and working as expected
- assign/unassign to shipment works including error handling
- when clicking a box row, I'm directed to BoxView
- filter by gender works

HaGuesto:
@pylipp very nice review. Thank you. @aerinsol. I'll try to address some in the follow-up PR. @aerinsol, please check if my priorization is ok for you

Here, some quick comments:

> I have a 800x1300 notebook screen that does not fit the entire table width. I can horizontal-scroll but it's a bit weird that the horizontal scroll bar is hidden at the bottom of the table but the vertical one is visible
I'll try to always show the horizontal bar. nice to have?

> first column is called "Box number" but in dropapp it's "Box ID". Personally I find "Box label" okay but it depends on what's used by users in practice
@aerinsol for you to decide

> filter by box state works though the state names for the filters could be in PascalCase instead of UPPERCASE
noted down as nice to have

> filter by size does not work as expected: Filtering by size S returns all boxes with sizes that contain the letter S
I'll take a look. rather must have.

> filter by product: dropdown filter list could be alphabetically sorted.
nice to have.

>showing/hiding columns works. But it's not being remembered after I reload the page (not sure if that's intended. I think it's quite useful especially when going back and forth between BoxesView and individual BoxViews)
@pylipp @aerinsol I will not be able to implement a persistent cache when reloading (like pressing the reload button in the Browser), because the cache is flushed and we have not implemented a persisitent cache yet. However, going back and forth between Boxview and BoxesView should remember the columns (I will test this again).

> column names are center aligned. When I hide a lot of columns, it looks a bit off (especially the select-all check-box moves, and is not aligned anymore with the rows checkboxes)
I already have that on my list.

> when selecting a row and clicking "Move to Location", the move is executed. When I check the corresponding BoxView, the location is updated but there is no history entry displayed (this only happens if I had been to this BoxView before)
seems like history entries are not passed correctly in the cache then. Will take look. Nice to Have?

aerinsol:
> > I have a 800x1300 notebook screen that does not fit the entire table width. I can horizontal-scroll but it's a bit weird that the horizontal scroll bar is hidden at the bottom of the table but the vertical one is visible
> 
> I'll try to always show the horizontal bar. nice to have?

Must have along with the number of columns being shown. 

> > first column is called "Box number" but in dropapp it's "Box ID". Personally I find "Box label" okay but it depends on what's used by users in practice

@HaGuesto better make both Dropapp and V2 "Box #" then. Both should be the same. If it's easier we can just change V2 right now. 

> 
> > filter by box state works though the state names for the filters could be in PascalCase instead of UPPERCASE
> 

Not a priority. 

> > filter by size does not work as expected: Filtering by size S returns all boxes with sizes that contain the letter S
> 
> I'll take a look. rather must have.
> 
> > filter by product: dropdown filter list could be alphabetically sorted.
> 

No. Don't do this, since we don't have categories right now in the table. We should decide later if we should sort Categories inside the same filter or not. 

> However, going back and forth between Boxview and BoxesView should remember the columns (I will test this again).

This doesn't work @HaGuesto 

> > when selecting a row and clicking "Move to Location", the move is executed. When I check the corresponding BoxView, the location is updated but there is no history entry displayed (this only happens if I had been to this BoxView before)
> 
> seems like history entries are not passed correctly in the cache then. Will take look. Nice to Have?

Must have, but not necessarily right now (not required for launch)

front/src/views/Boxes/components/BoxesActionsAndTable.tsx
```diff
@@ -38,54 +54,251 @@ function BoxesActionsAndTable({
...
+        false,
+      );
+      if (
+        assignBoxesToShipmentResult.notInStockBoxes &&
+        assignBoxesToShipmentResult.notInStockBoxes.length > 0
+      ) {
+        newAlerts.push({
+          id: alerts.length,
+          status: "info",
+          message: `Cannot assign ${
```
aerinsol, line 137: what happens if 10 of those boxes are in stock, and 10 of those are not? Does the instock move succeed? @HaGuesto

front/src/views/Boxes/components/BoxesActionsAndTable.tsx
```diff
@@ -38,54 +54,251 @@ function BoxesActionsAndTable({
...
+            ![BoxState.Receiving, BoxState.MarkedForShipment, BoxState.InTransit].includes(
+              box.values.state,
+            ),
+        )
+        .map((box) => box.values.labelIdentifier);
+
+      const boxCountInShipmentStates = selectedBoxes.length - movableLabelIdentifiers.length;
+      if (boxCountInShipmentStates > 0) {
+        newAlerts.push({
+          id: alerts.length,
```
aerinsol, line 84: @HaGuesto I need to check this more carefully, but I feel that in general all of these should be toasts and not alerts. The general rule of thumb is: toasts for dynamic responses to user actions; alerts for persistent states OR confirmation/denial that has detailed information the user needs to write down.

aerinsol: I have some comments on the texts for the messages too. Would you like a new Trello card for this and the above feedback? @HaGuesto

PR #1128, opened 2023-11-23
BoxView: Hide Reassignment of Shipment, Change Alerts and Style

vahidbazzaz:
As part of this PR, the reassignment of shipments is hidden and the alerts and styles are also changed


vahidbazzaz:
[2.0 BoxView: hide reassignment of shipment](https://trello.com/c/4w2ZeeDY/1314-20-boxview-hide-reassignment-of-shipment)

HaGuesto (approved):
Functionally, everything works.

For the Alerts on top of the BoxCard, I was not too happy with the responsive behavior and refactored it a bit. Now, we do not have a overflow in the BoxCard anymore and the Alert on top is in the same length.

PR #1129, opened 2023-11-23
Prepare Menu and add action based permission checking for routing

HaGuesto:
@aerinsol I will tag you in two places in the code:

1. where the MenuItems are defined and their titles and what action based permissions you will need to have to see the menu.
2. where the routes are defined and what action based permissions you will need to have to see the menu.

Could you check for both if I have put the correct permissions? 

This is how the menu looks now in mobile:
image

This is how the menu looks now in desktop:
image
image


front/src/App.tsx
```diff
@@ -75,27 +81,77 @@ function App() {
             <Route path=":hash" element={<ResolveHash />} />
           </Route>
           <Route path="boxes">
-            <Route index element={<Boxes />} />
+            <Route
+              index
+              element={<Protected component={<Boxes />} requiredAbp={["view_inventory"]} />}
```
HaGuesto, line 86: @aerinsol can you check from here to line 151 if the required action based permissions are correct?

front/src/components/HeaderMenu/HeaderMenuContainer.tsx
```diff
@@ -1,136 +1,93 @@
...
 
-  const menuItems: MenuItemsGroupData[] = useMemo(
-    () => [
+  // TODO: do this at route definition
+  const menuItems: IMenuItemsGroupData[] = useMemo(() => {
+    const oldAppUrlWithBase = `${process.env.REACT_APP_OLD_APP_BASE_URL}/?camp=${baseId}`;
+    return [
       {
-        text: "Classic View",
+        text: "Inventory",
```
HaGuesto, line 20: @aerinsol can you check from here to line 64 if the titles and the required action based permissions are correct?

HaGuesto:
[2.0 Prepare Menu and add action based permission checking for routing](https://trello.com/c/QybtGq4Y/1315-20-prepare-menu-and-add-action-based-permission-checking-for-routing)

front/src/App.tsx
aerinsol: @HaGuesto this is "Manage Boxes" right? if so then it should be `manage_inventory` permission, `view_inventory` permission is only for Stock Overview

aerinsol:
@HaGuesto looks ok, but I didn't see in the route protection area anything related to V1 in the area you tagged me in. Is that handled elsewhere?

HaGuesto:
The v1 routes are defined in dropapp repo. I cannot implement a route protection in this repo for routes that are outside the react app. The only thing I did and can do is to show the links to v1 only if the user has the correct action based permissions.

front/src/components/HeaderMenu/HeaderMenuContainer.tsx
HaGuesto: @aerinsol here is the permissions defined that you need to see a menu.

front/src/components/HeaderMenu/HeaderMenuContainer.tsx
```diff
@@ -1,136 +1,93 @@
...
+            requiredAbp: ["create_label"],
+            external: true,
           },
           {
-            link: `${process.env.REACT_APP_OLD_APP_BASE_URL}/`,
-            name: "Go to classic desktop",
+            link: "/boxes",
+            name: "Manage Boxes v2",
+            beta: true,
+            requiredAbp: ["create_transfer_agreement"],
```
aerinsol: Should this be `manage_inventory` permission for Manage Boxes v2? @HaGuesto

aerinsol (changes requested):
Looks good, with some correction in both menu view permssions and app permissions around Manage Boxes. Details in comments.

HaGuesto:
@aerinsol I updated the PR according to your request. Could you approve the PR and merge it.

Then you can test it. I especially, would like you to test the transfer menus as warehouse.volunteer@lesvos.org who does not have permissions for all routes.

aerinsol:
@HaGuesto I want to log that I'm force merging as requested, although codecov is complaining about lack of test coverage.

PR #1150, opened 2023-12-06
[statviz] Update GraphQL types

pylipp:
As per request of @maikneubert, for avoiding type-casting in FE.

All statviz-specific queries are affected, if you have to make changes to FE code you can push it here I suggest.


HaGuesto:
@MaikNeubert you have to decide what to do with this PR

MaikNeubert (approved):

PR #1154, opened 2023-12-08
Avoid triggering 'select * from stock'

pylipp:
See 4f22adc2d or https://github.com/coleifer/peewee/issues/2586 for
details.
This hopefully fixes excessive memory usage when the BE tries to fetch
and cache 100k of rows from the stock table.


pylipp:
[2.0 Performance bug: Avoid triggering 'select * from stock'](https://trello.com/c/x9MXG1ZE/1333-20-performance-bug-avoid-triggering-select-from-stock)

PR #1156, opened 2023-12-08
Fix assigning tags when creating box

pylipp:
Previously assign_tag() was used here which tried to select the Box
object which was only just about to be created.
Strangely this does not fail in the tests, and was only reproduced and
reported as a bug this week although the functionality is in use for at
least half a year. The issue is caused by the DB replica installed in
production: for DB reads, the replica is used but it does not contain
the write-change of the created box yet which has only gone into the
primary DB instance.


pylipp:
[2.0 Bug: assigning tag at box creation fails](https://trello.com/c/kMWAYK05/1334-20-bug-assigning-tag-at-box-creation-fails)

PR #1157, opened 2023-12-09
Improve BoxesView FE performance PART 1

HaGuesto:
https://www.loom.com/share/82741ae1290840f7b8935eca03c10498?sid=a7909b55-b3c1-45d0-a6c6-66a8ceb3c205

PR #1160, opened 2023-12-11
Inspect boxes-query performance

pylipp:
I inspected SQL generated by peewee, and profiled the stack of the app when executing the `boxes` GraphQL query while using a production DB dump.

### Findings

- I could improve filtering of the select-query (filter on join instead of where)
- I added a filter for non-deleted boxes
- when a large result is queried, there will be overhead with both peewee and ariadne
    - peewee as the ORM will have significant overhead transforming the raw SQL rows into Python objects (e.g. datetime parsing is quite costly)
    - ariadne will issue an enormous amount of function calls to resolve the query fields (e.g. for each box the location and product), and since these make use of batch-loading, the eventual resolving of awaitables has an overhead, too

### Possible solutions braindump

- smaller size of requested data (i.e. true pagination, only instock boxes, see #1157)
- hand-craft a single SQL query in the `boxes` resolver (this somewhat defeats the purpose of GraphQL)
- experiment with a new, synchronous [batch-loading library](https://github.com/jkimbo/graphql-sync-dataloaders): in order to use the `aiodataloader` library, currently an asyncio routine is patched into the (sync) Flask app with the (sync) ORM peewee. Maybe this generates significant overhead.
- in combination with the previous, try to speed up Python execution time using [gevent](https://docs.peewee-orm.com/en/latest/peewee/database.html#async-with-gevent) async execution

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,6 +7,8 @@ yarn-error.log*
 node_modules
 **/node_modules
 
+**/dump*.sql
+
 # environments
 .env.*
 .env
--- a/back/boxtribute_server/business_logic/warehouse/box/queries.py
+++ b/back/boxtribute_server/business_logic/warehouse/box/queries.py
@@ -25,13 +25,18 @@ def resolve_box(*_, label_identifier):
 def resolve_boxes(*_, base_id, pagination_input=None, filter_input=None):
     authorize(permission="stock:read", base_id=base_id)
 
-    # Join with Location model to filter for Location base ID below
-    selection = Box.select().join(Location)
+    selection = Box.select().join(
+        Location,
+        on=(
+            (Box.location == Location.id)
+            & (Location.base == base_id)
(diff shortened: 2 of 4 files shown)


pylipp:
[2.0 BE Performance: excessive loading times of BE request for BoxesView](https://trello.com/c/SsA0hd3O/1337-20-be-performance-excessive-loading-times-of-be-request-for-boxesview)

HaGuesto:
@pylipp code review is passing. I will merge this now, nur check again if delete Boxes are not part of the result.

HaGuesto:
Functional review is also passing. @pylipp  I was a bit confused since you filtered for deleted Boxes in the ON clause of the join function.

Is this join function than an inner join and not a left join?

pylipp:
> Functional review is also passing. @pylipp I was a bit confused since you filtered for deleted Boxes in the ON clause of the join function.
> 
> Is this join function than an inner join and not a left join?

It's an inner join.

PR #1165, opened 2023-12-15
Unify tsconfig and eslint

HaGuesto:
Learnings: 

- current setup works locally but not in docker
- on local yarn install creates a symlink to ui-components in node_modules, in docker it does not
- assumption is that we do not copy the right files in the root Dockerfile


HaGuesto:
[Refactor Repo FE structure and align ts, eslint and prettier](https://trello.com/c/4e3fEKMa/1348-refactor-repo-fe-structure-and-align-ts-eslint-and-prettier)

PR #1172, opened 2023-12-18
Use DB replica only for GraphQL queries

pylipp:
Follow-up to #1104.

When a GraphQL request contains a mutation, the database operations will be executed on the primary instance.
See commit messages for details.

This PR includes small improvements of the DatabaseManager, too.

### DIFF
--- a/back/boxtribute_server/business_logic/statistics/crud.py
+++ b/back/boxtribute_server/business_logic/statistics/crud.py
@@ -1,3 +1,5 @@
+from functools import wraps
+
 from peewee import JOIN, SQL, fn
 
 from ...db import db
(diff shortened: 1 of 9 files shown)


pylipp:
[2.0 Use DB replica only for GraphQL queries](https://trello.com/c/r5jICDlf/1347-20-use-db-replica-only-for-graphql-queries)

jamescrowley:
@pylipp quick question - totally makes sense that with the consistency model in GCP (ie eventual consistency), if we’re making decisions during a mutation, we should definitely be reading from primary. 

However, my understanding was the intention of the read-only replica was only to offload readonly OLAP style reporting, and all other operations would naturally run off primary. 

However, if I’m reading this PR correctly, is the intention that *all* read-only queries will run off the replica?

HaGuesto:
@jamescrowley yes, you are correct. The initial intention of the read-replica got somehow a bit lost in the many back and forth around the Statviz queries.
I think we should align here again @pylipp . I do not fully remember when we made the decision to use it for all select queries or if we just wanted to quickly test the read replicas with the least effort.

pylipp:
Thanks for bringing this up, James.
I don't fully remember either, and in #1104 we had similar confusion 😬 
The situation now again is different since the statviz part will be integrated as fancygraphs into boxtribute v2.

If you think there's major (dis)advantages to either approach (proposed as here vs off-loading statviz-related requests only) I'll change it @HaGuesto @jamescrowley

jamescrowley:
@pylipp I think given the original motivation (as I understood it) was to de-risk running expensive queries on the replica to avoid it affecting the operation of the primary functions in the app, I would say we should only be off-loading statviz related requests at this stage - otherwise, we risk the main app still being negatively affected by this load. @HaGuesto obviously shout if you disagree

back/boxtribute_server/business_logic/statistics/crud.py
```diff
@@ -19,6 +21,36 @@
...
 
 
+def use_db_replica(f):
+    """Decorator for a resolver that should use the DB replica for database selects."""
+
+    @wraps(f)
+    def decorated(*args, **kwargs):
+        if db.replica is not None:
+            with db.replica.bind_ctx(
+                [
```
HaGuesto, line 31: if you want you can change this to include all models

HaGuesto (approved):

PR #1184, opened 2024-01-08
Fix age calculation

pylipp:
With the beginning of the new year, all test-back jobs fail. This PR provides the fix.


pylipp:
[2.0 Fix age calculation](https://trello.com/c/q95oiq8p/1349-20-fix-age-calculation)

PR #1191, opened 2024-01-18
[Draft] add statviz dashboard to v2

MaikNeubert:


HaGuesto (commented):
# Functional review
@MaikNeubert @aerinsol 

I'll try to separate really functional stuff that is needed for a release or something that we wanna pass to a Designer.

## must haves for the release
- [ ] the date format should be the regional date format
image
- [ ] we need to make the error and / or no data behavior clear consistent. @aerinsol do you want a red alert as we do in other parts of v2 if a query fails or something else. Do you wanna distinguish the case of no data vs error, e.g. network error?
- [ ] the options for MovedBoxes look weird if there is no data, possibly related to the point above.
image
- [ ] In Stock Overview, it should be Category, Product, and Size instead of categoryName, productName, sizeName in the selection. In the overlay, I would also rather say: Select the next grouping
image
- [ ] Form fields should have consistant style, e.g. no rounded edges, thick border, ...


## passing to a Designer
- The grouping of graphs to the different accordeons make sense from a dev perspectiv since they share the same settings, but not from a user spertective. 
- I feel the settings take up to much space and I would probably put them in pop-ups or accordeons. I could also imagine puting some settings into the heading like Top Products by Items.
- I really like the Stock Overview, but I think we definitely also need a table view. (also maybe for the top products or csv export as a beginning.

Otherwise, functionally it works quite well.

HaGuesto:
also @aerinsol what should the name in the menu be?

front/src/views/Statviz/Dashboard/DashboardView.tsx
```diff
@@ -0,0 +1,5 @@
+import { Dashboard } from "@boxtribute/shared-components";
```
HaGuesto, line 1: not sure why this is needed

shared-components/Statviz/Dashboard/Dashboard.tsx
```diff
@@ -7,20 +7,17 @@ import StockOverview from "./StockOverview";
...
-        spacing="10px"
+        borderWidth="1"
+        spacing="10"
         direction={["column", "row"]}
-        style={{ padding: "20px", marginBottom: "40px" }}
+        style={{ padding: "15px", marginBottom: "15px" }}
         shadow="md"
       >
-        <WrapItem w="350px">
+        <WrapItem w="350">
```
HaGuesto, line 20: is 350 not the chakra ui value then?

HaGuesto (approved):
## Code review

nothing to complain

HaGuesto:
I'm planning to merge this so that @aerinsol can test on master already after figuring out why the BE tests fail.

@MaikNeubert

front/src/views/Statviz/Dashboard/DashboardView.tsx
MaikNeubert: removed

MaikNeubert:
@HaGuesto 
i am using the Data-Picker that is also used in other parts of v2. I think something like a datepicker should be consistent across the application. The datepicker is based on the default browser input "date" and it doesn't support local date formats.

Should i search for a good Datepicker that meets our requirement and replace it also in all other places where it is used in v2?

PR #1234, opened 2024-02-13
ADR for sync-async backend

pylipp:
https://github.com/boxwise/boxtribute/blob/adr/sync-async-backend/docs/adr/adr_sync-async-back-end.md

### DIFF
--- a/docs/adr/adr_sync-async-back-end.md
+++ b/docs/adr/adr_sync-async-back-end.md
@@ -82,6 +82,16 @@ From the [ariadne GitHub forum](https://github.com/mirumee/ariadne/discussions/1
 
 </details>
 
+<details>
+  <summary>From hackernews:</summary>
+
+1. [Discussion of Python asyncio](https://news.ycombinator.com/item?id=33547323), as introduced in [this blog post](https://superfastpython.com/python-asyncio/)
+    - emphasis on using Python async for I/O bound tasks only
+    - [gevent mentioned as alternative](https://news.ycombinator.com/item?id=33553316), also [here](https://news.ycombinator.com/item?id=33551049)
+    - [trap of miss-defining sync/async endpoints with FastAPI](https://github.com/tiangolo/fastapi/issues/260#issuecomment-495945630)
+
+</details>
+
 ## Consequences
 
 What becomes easier or more difficult to do because of this change?
@@ -102,6 +112,7 @@ What becomes easier or more difficult to do because of this change?
 - (+) subscriptions are supported
 - (+) calls to external APIs are non-blocking
 - (o) little team experience with development of async Python
+- (o) uncertainty about the hosting of the BE service because ASGI might not be supported in Google App Engine
 - (-) risk and expense of replacing web-framework (Flask) and ORM (peewee) with async counterparts
 
 ### Details: Move to fully async BE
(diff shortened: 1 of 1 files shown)


jamescrowley:
Adding comments for the record from my CTO list

Pro sync:
On async vs sync python: I kinda feel like you need a good reason to go async right now, it doesn’t feel mature. In a very I/O heavy environment sure, but in middleware code I just haven’t seen that need (generally the DB is doing the I/O lifting). There seems to be a push to go that route - uvicorn or whatever in front, and then Dj channels, Starlette, et al as you said - but there’s a lot of novelty and not much payback.
---
Second the hesitance on using async python. I tried to dive into using async falcon for a little LLM thing the other day as a way to use python libs while not blocking threads forever waiting for the models. Felt like that scene from the simpsons with Sideshow Bob and the rakes. Edgecases, wrapped values etc everywhere. Libraries that claimed to be async but had just forgotten to wrap parts of themselves leading to hours long debugging. I’m sure it can be nice, but if you need something async use node or Elixir or something else where the ideas are more mature (I assume C# and Java these days as well, just really not my scene so can’t comment).
---
Pro async:
-----
I’ve been happily using async python in production for years. I’m happy to talk details I’d you want to migrate from Flask to Quart (which should be the easiest path). I would agree that either an entirely sync or async codebase is best, mixing the two is difficult to reason about.
As a disclaimer I’m a maintainer of Flask and Quart.
-----
“I would agree that either an entirely sync or async codebase is best, mixing the two is difficult to reason about.
Came here to say this. Straddling the middle ground feels like a recipe for divided teams, knowledge, and understanding.
I’m a big fan of Ariadne’s schema first approach. It feels right to me, in that you have a pre-agreed contract for the transport layer, in the same way that you would expect to agree a relational database table structure. Any teams that depend on it can work to it, and have confidence that it’ll be adhered to by providers/consumers. If not, you have bigger problems.
-----
Refactoring sync to async is clearly a bit of a pain, but I would argue it’s relatively mechanical and you can call out to async code pretty safely while you’re transitioning. Calling sync within async, on the other hand, is a world of potential mystery and intrigue that I would prefer to avoid - so e.g. the falcon uses are all sync, and falcon.sync_to_async() and friends all have a lot of caveats and therefore we decided not to use them.

jamescrowley:
And also adding the discussion https://github.com/mirumee/ariadne/discussions/1149 that @pylipp opened up

jamescrowley:
I've also pinged David for his thoughts - will keep you posted.

pylipp:
@aerinsol @HaGuesto 
I've updated this ADR with a bit of my latest research involving potentially hosting an ASGI server on Google App Engine.

I suggest to merge it in its current "undecided" form because the overall issue of the mixed sync-async state does not affect us too much and there's still some uncertainties of going either direction, and to remove these uncertainties some prototyping and experimenting would be necessary.

> As a disclaimer I’m a maintainer of Flask and Quart.

@jamescrowley it's handy to have this person in your contacts! :) 

> I've also pinged David for his thoughts - will keep you posted.

@jamescrowley did you hear back from him?

pylipp:
> @jamescrowley did you hear back from him?

NVM, found [this on Slack](https://boxwise.slack.com/archives/CB1UYDGLU/p1708928741687409)

aerinsol (approved):
Comprehensive and detailed coverage of the main points. Excellent.

HaGuesto (approved):

PR #1248, opened 2024-02-19
Update invoke_yarn script

pylipp:
This PR fixes the `invoke_yarn` script. Previously an array definition was missing (in the process of #1165).

Now yarn from the `statviz` docker-compose service is used, instead of requiring a local installation.

Test it via
```sh
pre-commit run --all-files eslint
pre-commit run --all-files prettier
pre-commit run --all-files tsc
pre-commit run --all-files graphql-types
```

Note: I'm not sure why `docker-compose exec -w /app statviz yarn --cwd front run generate-graphql-ts-types` produces so much output, can you check @haguesto?

Details:
- update comments
- replace unknown file_collections array
- add alias to invoke yarn in docker-compose service

### DIFF
--- a/invoke_yarn
+++ b/invoke_yarn
@@ -39,9 +39,9 @@ if [[ "$mode" = "lint" ]]; then
 elif [[ "$mode" = "format" ]]; then
     yarn statviz run format:write "$@"
 elif [[ "$mode" = "tsc" ]]; then
-    yarn front --cwd front run tsc:precommit --noEmit "${front_files[@]}"
-    yarn statviz --cwd statviz run tsc:precommit --noEmit "${statviz_files[@]}"
-    yarn statviz --cwd shared-components run tsc:precommit --noEmit "${shared_components_files[@]}"
+    yarn front --cwd front run tsc:precommit "${front_files[@]}"
+    yarn statviz --cwd statviz run tsc:precommit "${statviz_files[@]}"
+    yarn statviz --cwd shared-components run tsc:precommit "${shared_components_files[@]}"
 elif [[ "$mode" = "graphql-types" ]]; then
     yarn front --cwd front run generate-graphql-ts-types
     yarn statviz --cwd statviz run generate-graphql-ts-types
--- a/invoke_yarn
+++ b/invoke_yarn
@@ -21,17 +21,11 @@ for file in "$@"; do
 done
 
 yarn() {
-    local service container
+    local service
     service="$1"
-    container=boxtribute-"$service"-1
     shift
 
-    if ! [ "$(docker inspect -f '{{.State.Running}}' "$container")" = "true" ]; then
-        echo "$service not running"
-        docker-compose up -d "$service"
-    fi
-
-    docker exec -w /app "$container" yarn "$@"
+    docker-compose run --rm -w /app "$service" yarn "$@"
 }
 
 if [[ "$mode" = "lint" ]]; then
--- a/invoke_yarn
+++ b/invoke_yarn
@@ -1,36 +1,41 @@
 #!/usr/bin/env bash
-# Utility script to invoke front-end style checks in the front directory, or other
+# Utility script to invoke style checks in front-end directories, or other
 # tools in the repository using yarn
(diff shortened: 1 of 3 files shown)


invoke_yarn
```diff
@@ -1,36 +1,41 @@
...
-        file_collections["statviz"]+=("${file#statviz/}")
+        statviz_files+=("${file#statviz/}")
     elif [[ $file == shared-components/* ]]; then
         shared_components_files+=("${file#shared-components/}")
     fi
 done
 
-# TODO: move this to the pre-commit config. Otherwise people need to install yarn
+yarn() {
+    docker-compose exec -w /app statviz yarn "$@"
```
HaGuesto, line 24: use docker and not docker-compose

invoke_yarn
```diff
@@ -1,36 +1,41 @@
...
     elif [[ $file == statviz/* ]]; then
-        file_collections["statviz"]+=("${file#statviz/}")
+        statviz_files+=("${file#statviz/}")
     elif [[ $file == shared-components/* ]]; then
         shared_components_files+=("${file#shared-components/}")
     fi
 done
 
-# TODO: move this to the pre-commit config. Otherwise people need to install yarn
+yarn() {
```
HaGuesto, line 23: create two yarn functions - one running in the statviz container and one running in the front container

invoke_yarn
```diff
@@ -1,36 +1,41 @@
...
+    docker-compose exec -w /app statviz yarn "$@"
+}
```
HaGuesto, line 25: if no docker container exists with the correct name, then try to run yarn locally.

invoke_yarn
```diff
@@ -1,36 +1,41 @@
...
+
 if [[ "$mode" = "lint" ]]; then
     yarn run lint:fix "$@"
 elif [[ "$mode" = "format" ]]; then
     yarn run format:write "$@"
 elif [[ "$mode" = "tsc" ]]; then
-    echo "${file_collections["shared-components"]}"
-    yarn --cwd front run tsc:precommit --noEmit "${file_collections["front"]}"
-    yarn --cwd statviz run tsc:precommit --noEmit "${file_collections["statviz"]}"
+    yarn --cwd front run tsc:precommit --noEmit "${front_files[@]}"
```
HaGuesto, line 32: the cwd flag should not be needed anymore here

HaGuesto (changes requested):
additionally, if you want you can try to ignore this warning which makes pre-commit fail:

  0:0  warning  File ignored by default.  Use a negated ignore pattern (like "--ignore-pattern '!<relative/path/to/filename>'") to override

invoke_yarn
pylipp: does not work, I get `error Command "tsc:precommit" not found.`
It's only defined in the statviz and front package.json but not in the root one.

pylipp:
> additionally, if you want you can try to ignore this warning which makes pre-commit fail:
> 
> 0:0 warning File ignored by default. Use a negated ignore pattern (like "--ignore-pattern '!<relative/path/to/filename>'") to override

We should use the new option `--no-warn-ignored` but it only works with FlatEslint. [SO link](https://stackoverflow.com/a/77758970/3865876). I've updated the pre-commit config to ignore some files from .eslintignore in the eslint hook as a work-around.

invoke_yarn
```diff
@@ -1,36 +1,51 @@
...
-# TODO: move this to the pre-commit config. Otherwise people need to install yarn
+yarn() {
+    local service container
+    service="$1"
+    container=boxtribute-"$service"-1
+    shift
+
+    if ! [ "$(docker inspect -f '{{.State.Running}}' "$container")" = "true" ]; then
+        echo "$service not running"
+        docker-compose up -d "$service"
```
HaGuesto, line 31: I would like it here to run the job locally instead of starting up docker

pylipp: You mean, use `docker-compose run --rm ...` to run everything in a new container?

HaGuesto:
closed due to 

https://trello.com/c/gtV5lYon

FYI: @fhenrich33

PR #1250, opened 2024-02-20
manuel time group control over createdboxes

MaikNeubert:
Add manuel time group control over createdBoxes


HaGuesto (approved):
## code review passed

What do you do if there is an overflow? E.g. Days selected for a time span of a year?

MaikNeubert:
> ## code review passed
> 
> What do you do if there is an overflow? E.g. Days selected for a time span of a year?

Nothing, everything what happens is, that the graph will just look ugly. We can consider setting reasonable boundaries, but what does a user expect when selecting "days" for grouping while having a time range of more then 6 month selected?

Since we want to offer a lot of options i think the effort to prevent configurations (filters, download options etc.) that lead to unpractical results is too high and not really worth it in general.

PR #1233, opened 2024-02-13
Use batch-loader for Box.history

pylipp:
When formatting the messages of box history, batch-load for all requested boxes, and format using SQL joins (i.e. insert product, size, box state, location name), instead of iterating over each raw change and fetching property names.

The following query

```graphql
{
  boxes(baseId: 27, paginationInput: {first: 500}) {
    elements {
      history {
        id
        user { id }
        changes
        changeDate
      }
    }
  }
}
```

previously took ~8s and caused about 1800 DB calls.

Now it's <0.5s with 5 DB calls (3 due to pagination, 2 for the history and user batch-loaders).

Also found and fixed the bug that the `Box.history.user` field was missing a batch-loader call.

The generated SQL is somewhat like this (didnt replace all placeholders)
```sql
select
	GROUP_CONCAT(`t1`.`id`) as `ids`,
	GROUP_CONCAT(`t1`.`changedate`) as `change_dates`,
	GROUP_CONCAT(IFNULL(`t1`.`user_id`, "NULL")) as `user_ids`,
	GROUP_CONCAT(case when (`t1`.`changes` = %s) then CONCAT(%s, `t2`.`label`, %s, `t3`.`label`) when (`t1`.`changes` = %s) then CONCAT(%s, `t4`.`name`, %s, `t5`.`name`) when (`t1`.`changes` = %s) then CONCAT(%s, `t6`.`label`, %s, `t7`.`label`) when (`t1`.`changes` = %s) then CONCAT(%s, `t8`.`label`, %s, `t9`.`label`) when (`t1`.`changes` = %s) then CONCAT(%s, `t1`.`from_int`, %s, `t1`.`to_int`) when (`t1`.`changes` like %s) then replace(`t1`.`changes`, %s, %s) when (`t1`.`changes` like %s) then replace(`t1`.`changes`, %s, %s) else `t1`.`changes` end) as `all_changes`,
	`t1`.`record_id`
from
	`history` as `t1`
left outer join `products` as `t4` on
	((`t4`.`id` = `t1`.`from_int`)
		and (`t1`.`changes` = "product_id"))
left outer join `products` as `t5` on
	((`t5`.`id` = `t1`.`to_int`)
		and (`t1`.`changes` = "product_id"))
left outer join `location
s` as `t2` on
	((`t2`.`id` = `t1`.`from_int`)
		and (`t1`.`changes` = "location_id"))
left outer join `locations` as `t3` on
	((`t3`.`id` = `t1`.`to_int`)
[...]

PR #1259, opened 2024-02-26
Increase group_concat_max_len for BoxHistory loader

pylipp:
e.g. for Hermine's box with ID 26663 (and 65 history entries), the batch loader fails, causing all `history` fields to be returned as null.

Follow-up to #1233.

@MaikNeubert you could test it locally by running the webapp service connected to the prod DB via cloud SQL proxy, not sure if you've done it before.


pylipp:
[2.0 Increase group_concat_max_len for BoxHistory loader](https://trello.com/c/8rbLfz3z/1378-20-increase-groupconcatmaxlen-for-boxhistory-loader)

HaGuesto (approved):

PR #1265, opened 2024-03-07
ADR for more reliable GraphQL error response structure

pylipp:
[Rendered version](https://github.com/boxwise/boxtribute/blob/adr/graphql-error-responses/docs/adr/adr_graphql-error-responses.md)

### DIFF
--- a/docs/adr/adr_graphql-error-responses.md
+++ b/docs/adr/adr_graphql-error-responses.md
@@ -1,0 +1,160 @@
+# ADR: GraphQL error response structure
+
+[Trello card](https://trello.com/c/2fJoBEf8/1380-20-assort-adr-on-graphql-error-responses)
+
+Decision Deadline: 2024-03-15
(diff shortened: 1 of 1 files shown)


pylipp:
[2.0 ASSORT: ADR on GraphQL error responses](https://trello.com/c/2fJoBEf8/1380-20-assort-adr-on-graphql-error-responses)

aerinsol (approved):

HaGuesto (commented):
Generally, sounds like a nice approach. @jamescrowley this might be interesting to have a read.

Just one thing I want to highlight here. It will have an impact how we assert the result in the front-end. Especially, asserting the correct type might be a bit annoying in the beginning, e.g. typescript will complain that variables could also be undefined. 
However, I also think that implementing this this structure then automatically enforces us in the front-end to handle and assert the responses correctly.

PR #1285, opened 2024-03-20
Add docs for Auth0 dependency

jamescrowley:
At the request of DPGA, adding docs regarding our Auth0 dependency.


jamescrowley:
Self-merge for expediency as DPGA chased me for this - but feel free to propose changes and/or review post-merge.

Issue #1275, opened 2024-03-17
2nd iteration bugfixes/improvements


MaikNeubert:
## Feature Refinements Needed Before Launch for Basic Usability
Beneficiary Title should be "Beneficiaries Registered"

- [x]  Top Products Title should be "Top Created Products"
- [x] Sankey - exclude all orange and red-tinted colors from positive flows

~~Sankey - negative flows should have a negative number and be in the format "<locationname> (incoming)", e.g. "Stockroom (incoming) -40"~~ -> currently not possible

- [x] Sankey - "Negative Box Flow" Legend at the bottom of the graph should instead say "Red targets indicate an incoming/reverse flow, i.e. a removal of items/boxes from that location to instock storage."
- [x] Sankey - outgoing boxes should be renamed EITHER "outgoing <boxes/items>" or "outgoing stock"

-  [x] Sankey - Shipment flows should include the orgname, so <basename> | <orgname> e.g. " Samos | Samos Volunteers", "Lakadikia | IHA" -> @pylipp is it possible to add this to the target dimension of the moved boxes query?
Probably next iteration -> update already merged to master

- [x] Stock Chart - Rename it to "Drilldown Chart of Instock Items/Boxes"
- [x] Stock Chart-  Change the grouping Control Name to "Grouped by", and the text next to it to "Current Grouping: <Group>" instead of the current "<Group>"
- [x] Add Text in the Donut Center that summarizes the current drilldown level and split (e.g., "All Items by Category", "Tops by Size", "Women's Items/Boxes by Category" "S Sizes by Product") - this should be visible in the download.
- [x] Beneficiary title: "Beneficiaries Registered"

## Small Feature Refinements for Better Impact
- [ ] Tag, Gender, and Product Gender filters should be viewable in the Graph Title ("filtered by..." between viz title and download options)
- [ ] Sankey - default order flows by size, largest to smallest
- [x] make filter panel sticky
- [ ] include applied filters in downloads as an option

## Larger Feature Refinements for Better Impact
- [ ] Thousand Separator Radio Control (None | Comma | Period)
- [ ] Sorting Control for Sankey (biggest -> smallest, Smallest -> Biggest, A->Z, Z->A)

pylipp:
Hej @MaikNeubert adding org dimension should be easy. Is the organisation name as single field alright, or do you want the organisation ID in the facts and then look up the name in the dimensions info?

MaikNeubert:
@pylipp 
[...]

PR #1286, opened 2024-03-22
Statviz/2nd iteration

MaikNeubert:
Issue: https://github.com/boxwise/boxtribute/issues/1275


shared-components/statviz/components/visualizations/movedBoxes/BoxFlowSankey.tsx
```diff
@@ -129,13 +129,22 @@ export default function BoxFlowSankey({ width, height, data, boxesOrItems }: IBo
...
     outgoingNode,
-    ...movedBoxes.map((movedBox) => ({
-      id: movedBox.targetId,
-      name: movedBox.isNegative ? `${movedBox.name} removed` : movedBox.name,
-      nodeColor: movedBox.isNegative
-        ? "red"
-        : sample(["#9467bd", "#8c564b", "#e377c2", "#7f7f7f", "#bcbd22"]),
-    })),
+    ...movedBoxes.map((movedBox) => {
+      const getName = () => {
```
HaGuesto, line 133: I would rather define this utility function somewhere outside. This way it is created during each render.

HaGuesto (approved):
@aerinsol I suggest that the sticky header can be minimized in some way. Otherwise, I do not think you can use it on mobile.

@MaikNeubert review passes. There was only one small remark

PR #1288, opened 2024-03-24
Tighten back-end security

pylipp:
- Remove `/public` route from staging (not helpful since staging DB does not contain history data)
- Whitelist URL paths that need database connection (otherwise DB connection is opened for any route requested, causing errors like [here](https://boxwise.sentry.io/issues/5067454591/?environment=production&project=6584296&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=30d&stream_index=0))
- Drop test databases at end of test session (improved clean-up)
- Define blueprints in separate module (more consistent)
- Remove Flask default route `/static/<filename>` (although not reachable due to our GAE configuration, it's not needed)

### DIFF
--- a/back/app-api.yaml
+++ b/back/app-api.yaml
@@ -5,9 +5,6 @@ handlers:
 - url: /
   script: auto
   secure: always
-- url: /public
-  script: auto
-  secure: always
 - url: /docs
   secure: always
   static_files: graphql-docs/index.html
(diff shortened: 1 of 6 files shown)


pylipp:
[2.0 Tighten back-end security](https://trello.com/c/l0P3UTbm/1387-20-tighten-back-end-security)

aerinsol (approved):

PR #1303, opened 2024-04-05
Update Python version to 3.12

pylipp:


pylipp:
having issues with mypy, trying to resolve [here](https://github.com/python/typeshed/pull/11731#issuecomment-2076531944)

back/Dockerfile
vahidbazzaz, line 1: @pylipp Could you please update the path in the Dockerfile:
change it from `RUN echo "/app/back" > /usr/local/lib/python3.10/site-packages/boxtribute-server.egg-link` to `RUN echo "/app/back" > /usr/local/lib/python3.12/site-packages/boxtribute-server.egg-link`

vahidbazzaz (changes requested):
Thanks, @pylipp . I have gone through your PR and encountered an issue building the image on my MacBook M1 machine. To resolve it, I needed to add the following command to the Dockerfile to ensure the compiler and necessary libraries were available:

```Dockerfile
RUN apk add --no-cache build-base libffi-dev
```

Without this, I received the following error:

```
× Building wheel for cffi (pyproject.toml) did not run successfully.
│ exit code: 1
╰─> [48 lines of output]
    No working compiler found, or bogus compiler options passed to
    the compiler from Python's standard "distutils" module.  See
    the error messages above.  Likely, the problem is not related
    to CFFI but generic to the setup.py of any Python package that
    tries to compile C code.  (Hints: on OS/X 10.8, for errors about
    -mno-fused-madd see http://stackoverflow.com/questions/22313407/
    Otherwise, see https://wiki.python.org/moin/CompLangPython or
    the IRC channel #python on irc.libera.chat.)
    Trying to continue anyway.  If you are trying to install CFFI from
    a build done in a different context, you can ignore this warning.
```

Additionally, after a successful build, I encountered a CORS error when sending requests to the backend. Could you please investigate this issue and make the necessary changes to resolve it?

Thank you!

back/Dockerfile
pylipp: good catch :muscle:

vahidbazzaz:
> > I encountered a CORS error when sending requests to the backend. Could you please investigate this issue and make the necessary changes to resolve it?
> 
> @vahidbazzaz does this error also happen on current master?

@pylipp I have tested this, and it is working fine in the master branch.

pylipp:
> > > I encountered a CORS error when sending requests to the backend. Could you please investigate this issue and make the necessary changes to resolve it?
> > 
> > 
> > @vahidbazzaz does this error also happen on current master?
> 
> @pylipp I have tested this, and it is working fine in the master branch.

@vahidbazzaz I'm a bit puzzled because the only relevant code change is a single line, replacing a deprecated datetime function. The rest is all changes to config files, mostly for deployment.

Is there any additional error information from the console of the `webapp` service or from the browser dev tools?

pylipp:
@vahidbazzaz I wonder if we're running into something like https://stackoverflow.com/a/61045178/3865876

Anyways it would help a lot if you could copy the exact request the FE sends by copying it e.g. as a curl command.

vahidbazzaz:
> @vahidbazzaz I wonder if we're running into something like https://stackoverflow.com/a/61045178/3865876
> 
> Anyways it would help a lot if you could copy the exact request the FE sends by copying it e.g. as a curl command.

@pylipp I've just rebuilt the Docker image in this branch, and it's now working perfectly. It seems the issue was with my previous build. Thanks!

PR #1309, opened 2024-04-09
[statviz] Reference authz test cases from test plan

pylipp:
[test plan](https://docs.google.com/spreadsheets/d/1sDhSsaVwNxAhGn1VYACDPVepZpvlwTkgy9nmeTaRy2Q/edit?pli=1#gid=1709672082)

### DIFF
--- a/back/test/endpoint_tests/test_statistics.py
+++ b/back/test/endpoint_tests/test_statistics.py
@@ -362,11 +362,13 @@ def test_query_stock_overview(
 
 
 def test_authorization(read_only_client, mocker):
+    # Test case 11.1.4
     # Current user is from base 1 of organisation 1.
     # Hence the user is not allowed to access base 2 from organisation 1
     query = "query { createdBoxes(baseId: 2) { facts { productId } } }"
     assert_forbidden_request(read_only_client, query)
 
+    # Test case 11.1.2
     # An accepted agreement exists between orgs 1 and 2 for bases 1+2 and 3.
     # Hence the user is allowed to access stock-related data from base 3
     for query in [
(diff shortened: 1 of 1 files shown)


pylipp:
[2.0 StatViz BE testing](https://trello.com/c/mKNwn1wN/1391-20-statviz-be-testing)

back/test/endpoint_tests/test_statistics.py
```diff
@@ -395,10 +399,12 @@ def test_authorization(read_only_client, mocker):
     ]:
         assert_forbidden_request(read_only_client, query)
 
+    # Test case 11.1.5
     # Base 5 does not exist
     query = "query { createdBoxes(baseId: 5) { facts { productId } } }"
     assert_forbidden_request(read_only_client, query)
```
aerinsol, line 405: @pylipp shouldn't this be an invalid data response rather than a forbidden response?

aerinsol (changes requested):
Just a small clarification needed

back/test/endpoint_tests/test_statistics.py
pylipp: @aerinsol my thoughts were: if a forbidden response is returned, it isnt even revealed that the base does not exists which I find more secure.
Also for checking invalid data a DB query is required but for authz check it's not.

aerinsol: I see, it's part of the series of tests for unauthorized access to show that now more information is shown than needed. Thanks for clarifying.

aerinsol (approved):

PR #1318, opened 2024-04-17
Enable cron-job to reseed staging DB

pylipp:
Some logic taken from
https://github.com/boxwise/dropapp/blob/dc7fa2c608460115519dacd25d41dc051b559525/cron/reseed-db.php

Updated .gcloudignore to insure init.sql is present in the deployed GAE environment.

Also it seems the `reseed-db` PHP cron job does run after all: [cron logs](https://console.cloud.google.com/cloudscheduler?project=dropapp-242214) with [details](https://console.cloud.google.com/logs/query;query=protoPayload.taskName%3D%226729109412488123316%22%20protoPayload.taskQueueName%3D%22__cron%22;cursorTimestamp=2024-04-17T00:18:00.864919Z;aroundTime=2024-04-17T00:36:17.182Z;duration=PT1H?project=dropapp-242214)

### Functional testing

Extend `.env` with

```bash
MYSQL_USER=root
MYSQL_HOST=127.0.0.1
MYSQL_PORT=32000
MYSQL_DB=dropapp_dev
MYSQL_PASSWORD=dropapp_root
```

and apply the following patch

```diff
diff --git a/back/boxtribute_server/cron/reseed_db.py b/back/boxtribute_server/cron/reseed_db.py
index 6fd73591..73ba55bf 100644
--- a/back/boxtribute_server/cron/reseed_db.py
+++ b/back/boxtribute_server/cron/reseed_db.py
@@ -15,11 +15,11 @@ def execute():
         f"-p{os.environ['MYSQL_PASSWORD']}",
         "-D",
         os.environ["MYSQL_DB"],
-        "-S",
-        os.environ["MYSQL_SOCKET"],
+        # "-S", os.environ["MYSQL_SOCKET"],
         # For testing locally on dev's machine with active docker-compose db service and
         #    dotenv run flask --debug --app boxtribute_server.dev_main:app run -p 5005
-        # "-h", "127.0.0.1", "-P", "32000",
+        "-h", "127.0.0.1", "-P", "32000",
         "-e",
         f"source {DATABASE_DUMP_FILEPATH}",
     ]
```

Then run
```bash
docker-compose up -d db
dotenv run flask --debug --app boxtribute_server.dev_main:app run -p 5005

# missing header
curl 'http://localhost:5005/cron/reseed-db'

# unknown url-path
curl 'http://localhost:5005/cron/foo'
```

Now to verify it actually works:
1. go to `http://localhost:5005/graphql` 
[...]

PR #1334, opened 2024-04-30
Fix cron-job to reseed staging DB

pylipp:
Instead of trying to use the mysql CLI which is not installed in the
Google App Engine environment. 
Follow-up to #1318.


pylipp:
[2.0 Enable cron-job to reseed staging DB](https://trello.com/c/ESgDE4QM/1395-20-enable-cron-job-to-reseed-staging-db)

PR #1336, opened 2024-05-02
make base removal script work with prod env

pylipp:
- Avoid importing peewee models
- Pass Auth0 auth parameters via CLI instead of .env file


pylipp:

PR #1338, opened 2024-05-03
Fix public key decoding error in v2-staging

pylipp:
- Revert "Debug public key"
- Update auth0-python package
- Extend gcloudignore


vahidbazzaz (approved):
Thank for a quick fix! 

I have tested this locally and it working fine.

PR #1342, opened 2024-05-07
fix that all single base usergroups are soft-deleted

HaGuesto:


pylipp:

pylipp (approved):

PR #1348, opened 2024-05-16
Fix readme and Auth0 integration tests

pylipp:
Likely caused by an improper teardown of a previous test run, e.g. due
to exceeding the Auth0 API rate limit which prevented proper deletion
of Auth0 test users


pylipp:

PR #1349, opened 2024-05-17
Fix: Disabling Install PWA on Client

vahidbazzaz:
This PR temporarily disables the installation of PWAs by commenting out `manifest.js` since it hasn't been tested yet.


vahidbazzaz:
[2.0 BUG: Install to Desktop Option should be Disabled (as it's untested)](https://trello.com/c/9zRby4kq/1303-20-bug-install-to-desktop-option-should-be-disabled-as-its-untested)

PR #1368, opened 2024-06-06
Use init.sql as staging DB seed

pylipp:
- Seed staging DB with init.sql dump
- Revert "Ping database server when generating lots of data"


pylipp:
[2.0 Enable cron-job to reseed staging DB](https://trello.com/c/ESgDE4QM/1395-20-enable-cron-job-to-reseed-staging-db)

PR #1384, opened 2024-06-23
Update Auth0 role IDs for staging when reseeding DB

pylipp:


pylipp:
[BUG: new user invitations do not work on staging](https://trello.com/c/SbJ6MDmQ/1424-bug-new-user-invitations-do-not-work-on-staging)

PR #1399, opened 2024-07-01
Update seeds with new XXL size

pylipp:
cf. boxwise/dropapp#599


pylipp:
[Add XXL size to sizegroup](https://trello.com/c/Qdwfqwn1/1437-add-xxl-size-to-sizegroup)

PR #1410, opened 2024-07-12
Self ref passing for downloading graphs on Statviz

fhenrich33:
My WIP/Attempt to fix the Stativiz download issue. I need a larger sample of data, however.

@aerinsol @HaGuesto @pylipp


fhenrich33:
[2.0 BUG: StatViz download graphs are broken](https://trello.com/c/u4kM6qTk/1419-20-bug-statviz-download-graphs-are-broken)

HaGuesto:
# functional review 

Generally, the download works. However, I encountered cases where the download button stayed in endless loading state without a timeout. I was not able to figure out on what it depended, but something we should check out again together @fhenrich33 before merging.
image

fhenrich33:
> # functional review
> 
> Generally, the download works. However, I encountered cases where the download button stayed in endless loading state without a timeout. I was not able to figure out on what it depended, but something we should check out again together @fhenrich33 before merging. image

Bummer, could you share an example of a case where this happened? We might want to take a look at other libraries and options.

HaGuesto:
reproduction:

- create a box with a QR-code
- download the graph for Created Boxes in statviz
- download the same graph again.

Assumption from Hans: It might happen when graphs are downloaded a second time.

We will create a follow up ticket since it is rather edge-case like. We will merge this ticket and test in production, too.

PR #1437, opened 2024-07-29
Enable filter for deleted bases

pylipp:
Filtering is now possible with these two queries:
```graphql
organisations { bases(filterInput: {includeDeleted: false}) { id }}
bases(filterInput: {includeDeleted: false}) { id }
```
Default for the filter parameter is to _include_ deleted bases. I did it this way because I didn't want a `bases` query to have breaking-change behavior. There might be API users who rely on the `bases` query returning deleted bases, too; and there is our own FE that might issue the `bases` query in different places, and I didn't want to break it.

I adjusted the FE already.

Furthermore the `Base.deletedOn` field was added.

### DIFF
--- a/back/boxtribute_server/business_logic/box_transfer/agreement/crud.py
+++ b/back/boxtribute_server/business_logic/box_transfer/agreement/crud.py
@@ -21,16 +21,19 @@ from ....models.utils import convert_ids, utcnow
 
 
 def _base_ids_of_organisation(organisation_id):
-    """Return IDs of that bases that belong to the organisation with given ID."""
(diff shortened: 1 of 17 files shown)


pylipp:
[2.0 BUG: Disabled bases are still visible for shipment agreement selection](https://trello.com/c/I8IMDIDX/1442-20-bug-disabled-bases-are-still-visible-for-shipment-agreement-selection)

HaGuesto (commented):
# code review passed

ACTION: please make `includeDeleted`:false the default.

HaGuesto:
# feature review passed

HaGuesto (approved):

PR #1442, opened 2024-08-01
Fix flakey Auth0 CLI test

jamescrowley:
Prevents this test having issues if run concurrently against auth0

### DIFF
--- a/back/test/integration_tests/test_cli.py
+++ b/back/test/integration_tests/test_cli.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import random
 import time
 from datetime import date
 from unittest.mock import patch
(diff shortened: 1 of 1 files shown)


back/test/integration_tests/test_cli.py
```diff
@@ -18,17 +19,20 @@
...
 WAIT = 5
 
+test_id_static_suffix = "-TEST"
+test_id_suffix = f"{date.today().isoformat()}-{random.randint(0, 1000000)}{test_id_static_suffix}"
 
 @pytest.fixture
 def auth0_roles(auth0_management_api_client):
     # Set up test roles in Auth0
     interface = auth0_management_api_client._interface
+    #generate test id suffix with datetime and a uuid guaranteed to be increasing
```
pylipp, line 29: can you move this comment to line 23 above? Also not sure how correct it still is since no UUID is used but instead a random int.
Can you maybe add that we do this to guarantee unique role names in Auth0?

pylipp:
[2.0 Fix flaky Auth0 integration test](https://trello.com/c/vU7Qywfl/1474-20-fix-flaky-auth0-integration-test)

back/test/integration_tests/test_cli.py
jamescrowley: Sorry, that was a bad rejig. I'll fix up with a clearer comment.

pylipp (approved):

fhenrich33:
[2.0 flaky FE test 4.8.2.1](https://trello.com/c/RnWDDglX/1463-20-flaky-fe-test-4821)

PR #1445, opened 2024-08-02
Add CodeCov configuration

jamescrowley:
Attempt to get codecov working as expected even when builds fail

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -4,7 +4,14 @@ codecov:
   notify:
     after_n_builds: 2
     wait_for_ci: yes
-
+coverage:
+  status:
+    # project doesn't work reliably, so show the result
+    # but don't fail anything
+    project:
+      default:
+        threshold: 100%
+    patch: true
 # https://docs.codecov.com/docs/flags#two-approaches-to-flag-management
 flag_management:
   default_rules:
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,11 @@
+# Docs: https://docs.codecov.com/docs/codecovyml-reference
+codecov:
+  require_ci_to_pass: yes
+  notify:
+    after_n_builds: 2
+    wait_for_ci: yes
+
+# https://docs.codecov.com/docs/flags#two-approaches-to-flag-management
+flag_management:
+  default_rules:
+    carryforward: true


pylipp:
[2.0 Force codecov run even with failing CI](https://trello.com/c/kYyInIua/1479-20-force-codecov-run-even-with-failing-ci)

.circleci/config.yml
```diff
@@ -148,6 +148,7 @@ jobs:
       - codecov/upload:
           flags: frontend
           file: coverage/coverage-final.json
+          when: always
```
pylipp, line 151: that's the [default](https://circleci.com/developer/orbs/orb/codecov/codecov#commands) but you wanted to make sure?

jamescrowley: Oops. I have to admit I hadn't tested - I took from the Trello discussion around codecov that reports weren't being uploaded when tests failed, and hence the flip flopping in coverage? I clearly misunderstood. Can you clarify when you have a second what you think the issue has been? Thanks 🙏

codecov.yml
```diff
@@ -0,0 +1,10 @@
+codecov:
```
pylipp, line 1: can you link https://docs.codecov.com/docs/codecovyml-reference here?

.circleci/config.yml
pylipp: @jamescrowley I suspect the following issue: e.g. when running the BE tests with pytest, the python-coverage package will not generate any report if tests fail but I have to double check this.

If no report has been generated, codecov has no data for upload, and then coverage numbers will drop.

jamescrowley: at least on GitHub I believe codecov's default behaviour is to fail if there is no report - [handle_no_reports_found](https://github.com/codecov/codecov-action). I can't see any reference to what the behaviour is on circleci.

I was hoping my other config change for requiring 2 reports - `after_n_builds: 2` - may tackle this? Do you happen to remember a PR where you've seen this happen recently?

codecov.yml
jamescrowley: Done

.circleci/config.yml
jamescrowley: I've removed the attributes from the PR given they do nothing!

jamescrowley:
@pylipp just had a failure here: https://github.com/boxwise/boxtribute/pull/1444/checks?check_run_id=28347330393 which I don't fully understand - possibly because it's not rebased against the latest commit on main?

jamescrowley:
@pylipp - found this: https://github.com/codecov/codecov-bash/issues/83 it's an old one, but it feels like it could be in this area (code coverage at head vs the PR base)

.circleci/config.yml
pylipp: in #1446 test-front fails and [no reports are uploaded to codecov](https://app.circleci.com/pipelines/github/boxwise/boxtribute/5137/workflows/5021dcd5-ac3a-43d7-a44e-0ae88a726d60/jobs/28042)

jamescrowley:
@pylipp as mentioned in Slack, suspect this could be tied to issues with front-end tests given all the failures are from indirect changes. Given this is only happening in codecov/project and not patch, and patch is most valuable (given it just looks at coverage in the specific PR), I've pushed a change that drops the codecov/project leniency to 100% until we can track this down further.

pylipp:
> @pylipp as mentioned in Slack, suspect this could be tied to non-determinism issues with front-end tests given all the failures are from indirect changes. Given this is only happening in codecov/project and not patch, and patch is most valuable (given it just looks at coverage in the specific PR), I've pushed a change that drops the codecov/project leniency to 100% until we can track this down further.

sounds reasonable, let's give it a try. :boom:

pylipp (approved):

PR #1446, opened 2024-08-02
Check only for not allowed or found MediaStream errors

fhenrich33:
Chrome (mobile only) was catching the [MediaStream.NotReadableError](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia#notreadableerror) because the camera wasn't loaded yet when the permission was evaluated. Thus showing an error toast.


HaGuesto (approved):
smart. good solution. Let's see if it works

PR #1447, opened 2024-08-02
Eagerly run location and shipment query to cache result

fhenrich33:


fhenrich33:
[2.0 Reduce initial loading time by removing auth waterfall in FE](https://trello.com/c/eq3ZB7Ao/1352-20-reduce-initial-loading-time-by-removing-auth-waterfall-in-fe)

aerinsol:
Can you explain what is eagerly in this case @fhenrich33 ?

fhenrich33:
> Can you explain what is eagerly in this case @fhenrich33 ?

Certainly @aerinsol 

Currently, we do:

`Auth > ORGANISATION_AND_BASES_QUERY > store base id in context if matches with URL param > MULTI_BOX_ACTION_OPTIONS_FOR_LOCATIONS_TAGS_AND_SHIPMENTS_QUERY > render`

I'm condensing it to:

`Auth > ORGANISATION_AND_BASES_QUERY + MULTI_BOX_ACTION_OPTIONS_FOR_LOCATIONS_TAGS_AND_SHIPMENTS_QUERY  > render`

Because Apollo will cache the query result and it will be handled accordingly. With the change, it won't run synchronously, but in parallel, avoiding the blocking, cascading effect.

PR #1471, opened 2024-08-14
Update GraphQL schema documentation

pylipp:
I built the schema documentation that we host e.g. on api-staging.boxtribute.org/docs ([instructions](https://github.com/boxwise/boxtribute/blob/master/docs/graphql-api/README.md#usage)), and zipped it:
[docs.zip](https://github.com/user-attachments/files/16713998/docs.zip)

Extract and view `public/index.html`.

### DIFF
--- a/back/boxtribute_server/business_logic/box_transfer/agreement/crud.py
+++ b/back/boxtribute_server/business_logic/box_transfer/agreement/crud.py
@@ -140,9 +140,10 @@ def create_transfer_agreement(
 ):
     """Insert information for a new TransferAgreement in the database. Update
     TransferAgreementDetail model with given source/target base information. By default,
-    the agreement is established between all bases of both organisations (indicated by
-    NULL for the Detail.source/target_base field). As a result, any base that added to
-    an organisation in the future would be part of such an agreement.
+    the agreement is established with all non-deleted bases of the partner organisation.
+    As a result, any base added to the partner organisation afterwards will NOT be part
+    of the agreement. A new agreement needs to be created then to transfer goods with
+    the new base.
     Convert optional local dates into UTC datetimes using user timezone information.
     Raise an InvalidTransferAgreementOrganisation exception if the current user's
     organisation is identical to the target organisation.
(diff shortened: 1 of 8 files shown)


pylipp:
[2.0 Improve documentation of graphQL schema](https://trello.com/c/9mky1CzF/1488-20-improve-documentation-of-graphql-schema)

back/boxtribute_server/business_logic/box_transfer/agreement/crud.py
```diff
@@ -140,9 +140,9 @@ def create_transfer_agreement(
 ):
     """Insert information for a new TransferAgreement in the database. Update
     TransferAgreementDetail model with given source/target base information. By default,
-    the agreement is established between all bases of both organisations (indicated by
-    NULL for the Detail.source/target_base field). As a result, any base that added to
-    an organisation in the future would be part of such an agreement.
+    the agreement is established with all non-deleted bases of the partner organisation.
+    As a result, any base that added to the partner organisation will NOT be part of the
```
aerinsol, line 144: We should probably say and will need to establish an additional agreement to transfer goods

back/boxtribute_server/business_logic/user/fields.py
```diff
@@ -8,7 +8,17 @@
 
 
 @user.field("bases")
-def resolve_bases(*_):
+def resolve_bases(user_obj, _):
+    if user_obj.id != g.user.id:
+        # If the queried user is different from the current user, we don't have a way
+        # yet to fetch secure information about that user's bases; it would require
```
aerinsol, line 14: So we will just throw an error or hit logins or...?

aerinsol (changes requested):
Nice, just some small clarifications

back/boxtribute_server/business_logic/user/fields.py
pylipp: we return `null` for the list of a user's bases

aerinsol: So basically it needs to be handled by the front end right? @HaGuesto or @fhenrich33 can you confirm it's been handled?

pylipp: yes but it's a field that the FE does not use (yet). Only user ID and name are used

PR #1472, opened 2024-08-14
Feature: Tags history

pylipp:
- Add temporal columns to TagsRelation
- Update TagsRelation business logic to set temporal fields when assigning or removing tags
- Update DB seed


pylipp:
[2.0/dropapp: implement temporal awareness about adding/removing tags](https://trello.com/c/0QOt3EcA/1489-20-dropapp-implement-temporal-awareness-about-adding-removing-tags)

HaGuesto (approved):
# code reviewed passed

# feature review
We move the feature review to when INserts and Deletions are also covered, because we need an updated init.sql here first.

back/boxtribute_server/graph_ql/definitions/protected/types.graphql
```diff
@@ -46,6 +46,7 @@ type Box implements ItemsCollection {
 }
 
 type HistoryEntry {
+  " IDs are not guaranteed to be unique due to history data being pulled from multiple sources "
```
HaGuesto, line 49: please make a consolidation task out of changing this comment: Rather add a prfix to the ID

pylipp: see [653b0ad](https://github.com/boxwise/boxtribute/pull/1472/commits/653b0ad871ec0fb64765f7983b0b5d85ee51ccd5)

HaGuesto:
# functional review
There is only one case where I'd probably change something. I would not show Box history items about assigning tags on the timestamp when the box is created. We also do not show the selected product at the moment of creation in the box history and this would end up confusing when tags are assigned before the box is created.
image

Therefore, I'd remove the assign tags entries from the box history when the box is created.

pylipp:
I agree that it might be confusing to see the created-record entries as more recent in the history. I thought this might be solved by grouping history entries with identical timestamp (tagged you in [trello](https://trello.com/c/Yur2g6CU)).

If you think this is not sufficient, I'll try to make the assigned-tag entries that belong to creation disappear. I think the easiest way is to not set the created-fields of TagsRelations when creating a box. Then these fields would only be set if an existing box is updated with tags.

PR #1494, opened 2024-08-29
Count only active shipments in Agreements Overview

pylipp:


pylipp:
[2.0 BUG: Manage Agreements Page Should Only List Active Shipments Per Base](https://trello.com/c/PrR1zdXa/1504-20-bug-manage-agreements-page-should-only-list-active-shipments-per-base)

aerinsol (approved):

PR #1496, opened 2024-08-30
Fix race condition by replacing globalContext with baseId hook

fhenrich33:


fhenrich33:
[Bug 2.0 fix race condition for currentBaseId retrival](https://trello.com/c/aecHq1vv/1506-bug-20-fix-race-condition-for-currentbaseid-retrival)

front/src/views/Box/BoxView.tsx
```diff
@@ -62,6 +62,7 @@ import { formatDateKey, prepareBoxHistoryEntryText } from "utils/helpers";
 import BoxDetails from "./components/BoxDetails";
 import TakeItemsFromBoxOverlay from "./components/TakeItemsFromBoxOverlay";
 import AddItemsToBoxOverlay from "./components/AddItemsToBoxOverlay";
+import { useBaseIdParam } from "hooks/useBaseIdParam";
```
HaGuesto, line 66: @fhenrich33 you can also exchange the usage in l. 254

front/src/views/Transfers/CreateShipment/CreateShipmentView.tsx
```diff
@@ -64,7 +65,7 @@ function CreateShipmentView() {
   const { globalPreferences } = useContext(GlobalPreferencesContext);
 
   // variables in URL
-  const baseId = globalPreferences.selectedBase?.id!;
+  const { baseId } = useBaseIdParam();
```
HaGuesto, line 69: @fhenrich33 in l. 109 pls replace `globalPreferences.organisation?.id;` with `currentBase?.organisation?.id` because then you can remove the globalPreferences context fully

front/src/hooks/useLoadAndSetGlobalPreferences.ts
```diff
@@ -21,11 +21,8 @@ export const useLoadAndSetGlobalPreferences = () => {
```
HaGuesto, line 21: Why is there no semicolon in l. 17?

front/src/hooks/useLoadAndSetGlobalPreferences.ts
```diff
@@ -21,11 +21,8 @@ export const useLoadAndSetGlobalPreferences = () => {
 
   useEffect(() => {
     // run query only if the access token is in the request header from the apollo client and the base is not set
-    if (user && !globalPreferences.selectedBase?.id) {
-      runOrganisationAndBasesQuery();
-    }
-  }, [runOrganisationAndBasesQuery,
-    user, globalPreferences.selectedBase?.id]);
+    if (user && baseId === "0") runOrganisationAndBasesQuery();
```
HaGuesto, line 24: This would mean that the query is only run if the url does not start with `/bases/<base_id>` and , thus the baseId === "0", correct. However, we want the globalPreferences filled also for the cases that a baseId is passed in the url. Therefore, I would put this back where it was.

front/src/hooks/useLoadAndSetGlobalPreferences.ts
HaGuesto, line 21: could you rename data in l.19 to something more descriptive, e.g. organisationAndBasesData

front/src/hooks/useLoadAndSetGlobalPreferences.ts
```diff
@@ -59,17 +56,17 @@ export const useLoadAndSetGlobalPreferences = () => {
         } else {
```
HaGuesto, line 56: This case and the `NotFoundView` are both triggering redirects. Could you remove this case since we actually do not need it anymore.

HaGuesto (changes requested):
Hey @fhenrich33 thanks for the PR. I put some inline change requests.

front/src/hooks/useLoadAndSetGlobalPreferences.ts
fhenrich33: Prettier weirded out here somehow 🤔

front/src/views/Transfers/CreateShipment/CreateShipmentView.tsx
fhenrich33: Notice that I'm only renaming `baseId` from `useBaseIdParam()` hook. So I can't really do that. 

However, there's a task to replace the global context to address that (or I can do it right here as this is related).

PR #1503, opened 2024-09-03
Soft-delete tags relations

pylipp:
- based on #1472


pylipp:
[2.0/dropapp: implement temporal awareness about adding/removing tags](https://trello.com/c/0QOt3EcA/1489-20-dropapp-implement-temporal-awareness-about-adding-removing-tags)

back/boxtribute_server/business_logic/tag/crud.py
```diff
@@ -80,19 +80,23 @@ def update_tag(
 
     with db.database.atomic():
         if object_type_for_deletion is not None:
-            TagsRelation.delete().where(
-                (TagsRelation.object_type == object_type_for_deletion)
-                & (TagsRelation.tag == id)
+            TagsRelation.update(deleted_on=utcnow(), deleted_by=user_id).where(
```
HaGuesto, line 83: pls use timestamp from decorator. 
Pls change also for boxes.
Pls check if timestamp in history is the same during creation as the created_on field

back/boxtribute_server/business_logic/warehouse/box/crud.py
```diff
@@ -130,6 +131,7 @@ def update_box(
     Insert timestamp for modification and return the box.
     """
     box = Box.get(Box.label_identifier == label_identifier)
+    now = utcnow()
```
HaGuesto, line 134: pls use timestamp from decorator

PR #1504, opened 2024-09-04
Show loading on global preferences loading

fhenrich33:
Branched out from https://github.com/boxwise/boxtribute/pull/1496


fhenrich33:
[Bug 2.0 show loading state if necessary globalPreferences are not loaded yet](https://trello.com/c/TonqZh8u/1508-bug-20-show-loading-state-if-necessary-globalpreferences-are-not-loaded-yet)

front/src/views/Transfers/CreateTransferAgreement/components/CreateTransferAgreement.tsx
```diff
@@ -174,6 +177,8 @@ function CreateTransferAgreement({
     validFrom,
   ]);
 
+  if (isGlobalStateLoading) return <APILoadingIndicator />;
```
HaGuesto, line 180: is there a reason why you put this in CreateTransferAgreement and not in the container component CreateTransferAgreementView? Otherwise, I'd move it up.

front/src/views/Transfers/CreateShipment/CreateShipmentView.tsx
```diff
@@ -62,9 +64,10 @@ function CreateShipmentView() {
   const { triggerError } = useErrorHandling();
   const { createToast } = useNotification();
   const { globalPreferences } = useContext(GlobalPreferencesContext);
```
HaGuesto, line 66: as mentioned https://github.com/boxwise/boxtribute/pull/1496#discussion_r1740609085

we could remove the usage of globalPreferences all together so that the loading state is not even necessary. Would do  so now.

front/src/components/BreadcrumbNavigation.tsx
```diff
@@ -35,6 +37,10 @@ export function BreadcrumbNavigation({ items }: IBreadcrumbNavigationProps) {
   const { globalPreferences } = useContext(GlobalPreferencesContext);
   const orgName = globalPreferences.organisation?.name;
   const baseName = globalPreferences.selectedBase?.name;
+  const { isLoading: isGlobalStateLoading } = useLoadAndSetGlobalPreferences();
+
+  if (isGlobalStateLoading) return <APILoadingIndicator />;
```
HaGuesto, line 42: could you replace this with a simple skeleton. It looks weird when the global loading indicator is shown instead of just one line in the view. Maybe ther is already a fitting skeleton in /components/Skeletons.tsx

HaGuesto (changes requested):
Thanks @fhenrich33! 95% there. I left some small inline comments and I am missing edits to ShipmentsOverviewView

front/src/views/Transfers/CreateShipment/CreateShipmentView.tsx
fhenrich33: The organization id is used throughout this view, so we still need to keep it there. However, as I mentioned in the comment following yours in https://github.com/boxwise/boxtribute/pull/1496#discussion_r1742971510, there's a task to address the global context refactor.

HaGuesto: ok, then let's leave it for later. thought it is just in one occasion

PR #1505, opened 2024-09-04
Handle droapp redirects not starting with `/bases/:baseId`

fhenrich33:


fhenrich33:
[2.0 forwarding from dropapp mobile.php to v2](https://trello.com/c/LXJ0Hxqo/1509-20-forwarding-from-dropapp-mobilephp-to-v2)

front/src/App.tsx
```diff
@@ -24,23 +24,31 @@ import { AlertWithoutAction } from "components/Alerts";
 import { ErrorBoundary } from "@sentry/react";
 import Dashboard from "@boxtribute/shared-components/statviz/dashboard/Dashboard";
 
-interface IProtectedRouteProps {
+type ProtectedRouteProps = {
```
HaGuesto, line 27: is there a reason you changed this styling format? Hope you will update it then in the README

front/src/App.tsx
```diff
@@ -24,23 +24,31 @@ import { AlertWithoutAction } from "components/Alerts";
...
   component: ReactElement;
   redirectPath: string | undefined;
   requiredAbp?: string[];
   minBeta?: number;
-}
+};
+
+type DropappRedirectProps = {
+  path: "/boxes/:boxId" | "/boxes/create/:qrCodeHash" | "/qrreader" | "/qrreader:qrCodeHash";
```
HaGuesto, line 35: I think you miss a slash here `/qrreader:qrCodeHash` : `/qrreader/:qrCodeHash`

front/src/App.tsx
```diff
@@ -54,10 +62,42 @@ function Protected({ component, redirectPath, requiredAbp, minBeta }: IProtected
...
-};
+
+/**
+ * Handle Dropapp (Boxtribute V1 app) redirects whose paths don't start with `/bases/:baseId`.
+ *
+ * Fetch first available base id from user JWT token from Auth0 to prepend `/bases/:baseId` with that id, if available.
+ */
+function DropappRedirect({ path }: DropappRedirectProps) {
+  const { user } = useAuth0();
+  let pathToRedirect = "/error";
```
HaGuesto, line 73: I do not think that we have this path

front/src/App.tsx
```diff
@@ -223,6 +261,14 @@ function App() {
           </Route>
         </Route>
       </Route>
+      <Route path="boxes">
+        <Route path=":boxId" element={<DropappRedirect path="/boxes/:boxId" />} />
+        <Route path="create" element={<DropappRedirect path="/boxes/create/:qrCodeHash" />} />
+      </Route>
+      <Route path="qrreader">
+        <Route index element={<DropappRedirect path="/qrreader" />} />
+        <Route path=":qrCodeHash" element={<DropappRedirect path="/qrreader:qrCodeHash" />} />
```
HaGuesto, line 270: I think you missed the slash here again

HaGuesto (approved):
great work. really like the PR @fhenrich33.

I think there is just one or two typos. I already approved it so you can merge the PR directly, when u fixed them

front/src/App.tsx
fhenrich33: The reason for this is to hit the `<NotFoundView />` Route. I'll leave a comment to make this more explicit.

front/src/App.tsx
fhenrich33: May I push for this style from now onwards? I'll change the README if I have your blessing. 😁

PR #1507, opened 2024-09-04
Downgrade `pnpm` so Dependabot can read the lockfile

fhenrich33:
Dependabot isn't issuing alerts with `pnpm` v9 lockfile format. See https://github.com/dependabot/dependabot-core/issues/10534

Downgrading to latest pre v9 lockfile `pnpm release until v9 is supported: https://github.com/pnpm/pnpm/releases/tag/v8.15.9

Next steps:

- Make a task to track revisiting v9 compatibility.
- Run `Check for updates` after this PR is merged to check for Dependabot security alerts that we missed due to the aforementioned issue.

image

This also address the following Dependabot PRs:

https://github.com/boxwise/boxtribute/pull/1506
https://github.com/boxwise/boxtribute/pull/1502
https://github.com/boxwise/boxtribute/pull/1488
https://github.com/boxwise/boxtribute/pull/1485


fhenrich33:
[2.0 FE downgrade pnpm due dependabot alert bug](https://trello.com/c/MIvhTpHm/1514-20-fe-downgrade-pnpm-due-dependabot-alert-bug)

HaGuesto (approved):
Thanks @fhenrich33

HaGuesto:
Should we include this in the production deploy on Monday or not @fhenrich33 ?

fhenrich33:
> Should we include this in the production deploy on Monday or not @fhenrich33 ?

Let's hold off and test it a bit more before we pull the trigger to be on the safe side.

jamescrowley:
@fhenrich33 looks good, thanks! Just a note that if folks don't downgrade pnpm (using `corepack use pnpm@8.15.9`), they will accidentally upgrade the lock file again, so we're going to have to keep an eye out for that.

I had hoped that

package-manager-strict-version=true

in .npmrc would at least limit fall out if pnpm 9 is installed (supported since then to enforce the packageManager version), but I couldn't get it working

jamescrowley (approved):

fhenrich33:
> @fhenrich33 looks good (and running locally fine for me), thanks! Just a note that if folks don't downgrade pnpm (using `corepack use pnpm@8.15.9`), they will accidentally upgrade the lock file again, so we're going to have to keep an eye out for that.
> 
> I had hoped that
> 
> ```
> package-manager-strict-version=true
> ```
> 
> in .npmrc would at least limit fall out if pnpm 9 is installed ([supported since v9.2](https://pnpm.io/npmrc#package-manager-strict-version) to enforce the packageManager version), but I couldn't get it working

Let's keep a close look at the following PRs to the frontend, and revisit the pnpm issue in the Dependabot tracker. I think it's the best move for now, IMO. @jamescrowley @HaGuesto @pylipp

PR #1511, opened 2024-09-09
Remove transactions from history in init.sql

pylipp:
- Avoid saving created transaction in history
- Update data in init.sql
- Update phinxlog


pylipp:
[dropapp: Free Shop transactions are saved in the history table](https://trello.com/c/eYI74wIy/1470-dropapp-free-shop-transactions-are-saved-in-the-history-table)

PR #1529, opened 2024-09-13
Enable eslint and type checking for Distribution feature

jamescrowley:
* Enabled eslint and type checking for Distribution feature
* Removed some disabled eslint rules that were not necessary

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -63,8 +63,6 @@ module.exports = {
     // we do not need to import React in every file
     "react/react-in-jsx-scope": "off",
     "react/jsx-props-no-spreading": "off",
-    // TODO: turn this on
-    "react/prop-types": "off",
     // matter of preference (allows to use props.propName instead of deconstructing props first)
     "react/destructuring-assignment": "off",
     // allow other than jsx extensions
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -63,8 +63,6 @@ module.exports = {
     // we do not need to import React in every file
     "react/react-in-jsx-scope": "off",
     "react/jsx-props-no-spreading": "off",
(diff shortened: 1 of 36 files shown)


.eslintrc.cjs
```diff
@@ -56,31 +57,19 @@ module.exports = {
       {
         selector: "interface",
         format: ["PascalCase"],
-        custom: {
-          regex: "^I[A-Z]",
-          match: true,
-        },
```
jamescrowley, line 62: Distribution doesn't obey this rule, but I think it's probably ok to relax this one for now?

fhenrich33: I intend to suggest defaulting to `type` instead of `interface` whenever possible so that's a nudge in that direction.

front/src/components/HeaderMenu/HeaderMenuDesktop.tsx
```diff
@@ -107,7 +107,7 @@ function HeaderMenuDesktopContent({ ...props }: IHeaderMenuProps) {
   );
 }
 
-function HeaderMenuDesktopContainer({ children }) {
+function HeaderMenuDesktopContainer({ children }: { children: React.ReactNode }) {
```
fhenrich33, line 110: I think children should be inferred correctly from all but most obscure Component types (or those self closing HTML tags). Is that not the case somehow?

jamescrowley: It's because they are React component props, that the linter requires explicit types even if they can be inferred, I believe.

jamescrowley:
[2.0: FE test/linting improvements from James](https://trello.com/c/cGPi62FA/1545-20-fe-test-linting-improvements-from-james)

jamescrowley:
Merging this, as I need to make conflicting changes to update Storybook dependency which will resolve webpack security issues

PR #1531, opened 2024-09-13
Fix test warnings about deprecated defaultProps

jamescrowley:
React deprecated defaultProps back in 2019, and we're getting test warnings about it - https://github.com/facebook/react/pull/16210. This switches them to use JS default props instead.


HaGuesto:
[2.0: FE improvements from James](https://trello.com/c/GR7CCcc7/1530-20-fe-improvements-from-james)

PR #1535, opened 2024-09-13
Stop logging expected GraphQL errors in test output

jamescrowley:
The current test set up intercepts GraphQL errors to print them as output. This is useful, as they warn about missing mock data and other errors. However, these messages are mixed up in 'expected' failures that we've set up in the tests themselves. 

This PR switches to separate classes for our fake errors so we can exclude them from the console output.


jamescrowley:
[2.0: FE test/linting improvements from James](https://trello.com/c/cGPi62FA/1545-20-fe-test-linting-improvements-from-james)

PR #1537, opened 2024-09-13
back-end: Measure products

pylipp:


pylipp:
[EPIC: Products with measure (continuous size)](https://trello.com/c/FtY3Uz4X/1517-epic-products-with-measure-continuous-size)

back/boxtribute_server/business_logic/statistics/crud.py
```diff
@@ -429,6 +432,9 @@ def compute_stock_overview(base_id):
             boxes.c.box_state,
             Product.category.alias("category_id"),
             fn.TRIM(fn.LOWER(Product.name)).alias("product_name"),
+            fn.CONCAT(
+                fn.ROUND(boxes.c.measure_value * Unit.conversion_factor, 2), Unit.symbol
```
HaGuesto, line 436: please round and group by absolute measure value and not relative to the display unit.

back/boxtribute_server/business_logic/statistics/sql.py
```diff
@@ -217,8 +232,9 @@
 FROM DeletedBoxes t
 JOIN products p ON p.id = t.product
 JOIN locations loc ON loc.id = t.location_id
+LEFT OUTER JOIN units u ON u.id = t.stock_display_unit_id
 LEFT OUTER JOIN tags_relations tr ON tr.object_id = t.box_id AND tr.object_type = "Stock" AND tr.deleted_on IS NULL
-GROUP BY moved_on, p.category_id, p.name, p.gender_id, t.size_id, loc.label
+GROUP BY moved_on, p.category_id, p.name, p.gender_id, t.size_id, loc.label, measure_name
```
HaGuesto, line 237: put measur before location in the grouping

HaGuesto (approved):
# code review passed 

# functional review
will be coming when we have the mutations for boxes updated.

FYI:
FE: since sizeId can be null now we need to check validation in form handlers.

pylipp:
> FE: since sizeId can be null now we need to check validation in form handlers.

@HaGuesto possible related: [build-front](https://app.circleci.com/pipelines/github/boxwise/boxtribute/5516/workflows/1014d127-0c63-44c5-8455-9b31b57b532b/jobs/30735) breaks

back/boxtribute_server/business_logic/statistics/crud.py
```diff
@@ -429,6 +432,10 @@ def compute_stock_overview(base_id):
             boxes.c.box_state,
             Product.category.alias("category_id"),
             fn.TRIM(fn.LOWER(Product.name)).alias("product_name"),
+            fn.CONCAT(
+                fn.ROUND(boxes.c.measure_value, 3),
+                Case(Unit.dimension_id, ((28, "kg"), (29, "l")), ""),
```
HaGuesto, line 437: grouping over a conditional in a column is one of the most expensive actions one can do in SQL. I'd suggest another solution. 
Rather query both columns (measure value and dimension) and group by both columns and then do the concat.

back/boxtribute_server/business_logic/statistics/sql.py
```diff
@@ -306,6 +336,10 @@
     TRIM(LOWER(p.name)) AS product_name,
     p.gender_id AS gender,
     d.source_size_id AS size_id,
+    -- neglect possible history of box's measure_value
+    CONCAT(ROUND(b.measure_value, 3),
```
HaGuesto, line 340: move the round function up to the cached function

back/boxtribute_server/graph_ql/definitions/public/types.graphql
```diff
@@ -99,7 +99,8 @@ type MovedBoxesResult {
   categoryId: Int!
   productName: String!
   gender: ProductGender!
-  sizeId: Int!
+  sizeId: Int
+  measureName: String
```
HaGuesto, line 103: please rename to absoluteMeasure

back/boxtribute_server/business_logic/statistics/sql.py
HaGuesto: SELECT ROUND(your_decimal_field, 3 - FLOOR(LOG10(ABS(your_decimal_field)) + 1)) AS rounded_value
FROM your_table_name;

HaGuesto: round to 3 significant non-zero digits

HaGuesto: AGREED: BE returns absoluteMeasure (just the number rounded to 3 non-zero significant digits) and dimension_id in facts. Dimension_id is then put into dimensions too

HaGuesto (approved):
looks good @pylipp. Is there a way that you could check the performance on the statistics queries before and after?

pylipp:
[2.0: BE implementation for measure-products](https://trello.com/c/ob2cwjyz/1565-20-be-implementation-for-measure-products)

pylipp:
> looks good @pylipp. Is there a way that you could check the performance on the statistics queries before and after?

Checked with the test DB (base 1 because it contains the most boxes; however it's just 500) and didn't observe any difference when validating against current master.

I can't run the branch on the current production DB because the branch expects certain new columns (stock.measure_value, stock.display_unit_id, etc). If the newest migrations in dropapp are deployed to prod, I can try again.

PR #1542, opened 2024-09-16
Return products of All type by default in Base.products fields

pylipp:


pylipp:
[Return products of All type by default in Base.products fields](https://trello.com/c/buCWZEaM/1522-return-products-of-all-type-by-default-in-baseproducts-fields)

PR #1560, opened 2024-09-24
Use gunicorn preload

pylipp:
cf. https://www.joelsleppy.com/blog/gunicorn-application-preloading


pylipp:
[2.0: use gunicorn preload](https://trello.com/c/jhGzwOQV/1529-20-use-gunicorn-preload)

PR #1578, opened 2024-10-05
Fix: Re-render video element on camera permission change for iOS

vahidbazzaz:
This PR addresses an issue where the video element does not properly re-render when the camera permission state changes on iOS. The `isCameraNotPermitted` state is now updated to `false` when permissions are granted, and this state change is passed to the `QrReaderScanner` component to trigger a re-render of video element.


vahidbazzaz:
[BUG: No iPhone camera access in QR scanner even after giving permissions](https://trello.com/c/HjvYpFRC/1528-bug-no-iphone-camera-access-in-qr-scanner-even-after-giving-permissions)

HaGuesto (approved):
thanks Vahid! nice PR

# code review
@vahidbazzaz I only changed one line because a similar if statement as you introduced in QRReaderScanner already existed.

# functional review
I'm not really able to test it since I do not have iOS. @vahidbazzaz could you confirm on staging if it is working?

PR #1595, opened 2024-10-08
Bump vitest to v2

jamescrowley:


jamescrowley:
[2.0: FE dev dependency bumps](https://trello.com/c/OpanAdUL/1546-20-fe-dev-dependency-bumps)

PR #1618, opened 2024-10-09
Upgrade to Storybook 8

jamescrowley:
Switch to SWC and Vite, given we are already using Vite elsewhere and not using Babel


jamescrowley:
[2.0: FE dev dependency bumps](https://trello.com/c/OpanAdUL/1546-20-fe-dev-dependency-bumps)

PR #1639, opened 2024-10-11
Align traceparent with Sentry's sentry-trace header

jamescrowley:
The span id within transparent will not match, but the trace id itself will. Can review if/when Sentry support for traceparent https://github.com/getsentry/sentry-javascript/issues/11171#issuecomment-2408107582 progresses.


jamescrowley:
@pylipp @HaGuesto confirmed this works in staging. traceid aligned between sentry/gcp. The span id does not, as sentry still does it's own thing wrapping requests, but I think this will do us for now until Sentry supports direct integration.

pylipp:
[2.0: Align traceparent with Sentry's sentry-trace header](https://trello.com/c/DV4NX6Fp/1563-20-align-traceparent-with-sentrys-sentry-trace-header)

pylipp (approved):

PR #1648, opened 2024-10-18
BE: support intra-org shipments

pylipp:
- Support creating intra-org shipments
- Add validation and verify authz


pylipp:
[2.0 Intra-org shipments between bases of the same org](https://trello.com/c/2rl8WblK/1550-20-intra-org-shipments-between-bases-of-the-same-org)

HaGuesto (commented):
# functional and code review passed

PR #1657, opened 2024-10-23
Validate matching product/location base when creating box

pylipp:


pylipp:
[2.0: Validate matching product/location base when creating/updating box](https://trello.com/c/YyVl7jWU/1569-20-validate-matching-product-location-base-when-creating-updating-box)

back/boxtribute_server/business_logic/warehouse/box/crud.py
```diff
@@ -182,6 +190,10 @@ def update_box(
     box_contains_measure_product = box.size_id is None
     new_product = Product.get_by_id(product_id or box.product_id)
     new_product_is_measure_product = is_measure_product(new_product)
+    new_location = Location.get_by_id(location_id or box.location_id)
```
HaGuesto, line 193: if both the new location and the new product are in another base, this would not fail.
Don't we also need a validation that the new_location_id and the old_location_id are in the same base?

pylipp: yes, I had thought of it too; will add it to the validation

HaGuesto (approved):
# code and fundtional review passing

PR #1673, opened 2024-11-04
Only allow deletion of boxes in warehouse-states

pylipp:


pylipp:
[2.0: Only allow deletion of boxes in warehouse-states](https://trello.com/c/Hrymzuua/1583-20-only-allow-deletion-of-boxes-in-warehouse-states)

HaGuesto (approved):
# functional and code review passed

PR #1675, opened 2024-11-04
Make all mutations available via beta-levels

pylipp:
- all mutations can be controlled via beta-levels
- when a new mutation is added to the BE, it must be added to one beta-level
- when in development, the FE requests a new mutation using a test user, the mutation needs to moved to a smaller beta-level


pylipp:
[2.0: Beta-level logic and test enhancements](https://trello.com/c/Zri1PeqM/1585-20-beta-level-logic-and-test-enhancements)

HaGuesto (approved):

PR #1685, opened 2024-11-09
Intra-org shipments between bases of the same org

fhenrich33:
firefox_CMKmjabMOM


fhenrich33:
[2.0 FE: Intra-org shipments between bases of the same org](https://trello.com/c/2rl8WblK/1550-20-fe-intra-org-shipments-between-bases-of-the-same-org)

front/src/views/Transfers/CreateShipment/CreateShipmentView.tsx
```diff
@@ -109,6 +109,7 @@ function CreateShipmentView() {
   const currentBase = allAcceptedTransferAgreements?.data?.base;
   const currentOrganisationLabel = `${currentBase?.organisation?.name} - ${currentBase?.name}`;
   const currentOrganisationId = globalPreferences.organisation?.id;
+  const currentOrganisationBases = globalPreferences.availableBases;
```
HaGuesto, line 112: available bases only shows the other bases in your organisation if you are a Headof Ops user. We will need to query the bases to create shipments with differently. @pylipp any suggestion how best to query all bases in your organisations even the ones you don not have access to?
Maybe:
```
organisation(id:2) {
      id
      name
      bases {
        id
        name
      }
    }
    ```

File: front/src/views/Transfers/CreateShipment/CreateShipmentView.tsx [hunk 2]
```diff
@@ -136,6 +137,7 @@ function CreateShipmentView() {
           name: agreement.targetOrganisation.name,
           bases: agreement.targetBases,
           agreementId: agreement.id,
+          agreementComment: agreement.comment,
```
HaGuesto, line 140: pls. add the same line in l. 132

front/src/views/Transfers/CreateShipment/CreateShipmentView.tsx
pylipp: @HaGuesto that's exactly the way to go.

HaGuesto (approved):
## code review passed
just on comment in the code

## functional review passed
please hide intra-org tab if ther is no option available for intra-org shipment

front/src/views/Transfers/CreateShipment/CreateShipmentView.tsx
HaGuesto: @pylipp I have only the name and id of other bases is available through this query. It is the same query used to create transferAgreements, right?

aerinsol:
@HaGuesto @fhenrich33 
Some requested UI changes:
1. The other tab should only show the name of the organization, NOT the name of the base. Here the name of the org is "BoxCare", where Thessaloniki is the current base and CANNOT be selected as one of the options in the dropdown (i.e., you can't send a transfer to yourself). 
2. I'd like to keep the format of the original Figma design, ie.., <semibold>`baseName`</semibold> `orgName`

PR #1722, opened 2024-11-15
Use a new query to get all available bases for current org

fhenrich33:
Follow-up for https://github.com/boxwise/boxtribute/pull/1685#pullrequestreview-2436147595


fhenrich33:
[2.0 FE: Intra-org shipments between bases of the same org](https://trello.com/c/2rl8WblK/1550-20-fe-intra-org-shipments-between-bases-of-the-same-org)

front/src/views/Transfers/CreateShipment/CreateShipmentView.tsx
```diff
@@ -39,6 +40,19 @@ export const ALL_ACCEPTED_TRANSFER_AGREEMENTS_QUERY = gql`
   }
 `;
 
+export const ALL_AVAILABLE_BASES_QUERY = gql`
+  query AllAvailableBases($orgId: ID!) {
+    organisation(id: $orgId) {
+      id
+      name
```
pylipp, line 47: Do you even need to fetch id and name of the organisation?

fhenrich33: I pass it as a prop to `CreateShipment.tsx`

HaGuesto (commented):
missing:
- add loading state of new query in line 256 
- add error case of new query in line 281

change request:
- move noAcceptedAgreements alert in line 279 into sub component via props and show in the intER-org shipment tab instead of the corresponding form.

front/src/views/Transfers/CreateShipment/CreateShipmentView.tsx
```diff
@@ -110,7 +124,15 @@ function CreateShipmentView() {
   const currentOrganisationName = currentBase?.organisation?.name;
   const currentOrganisationBase = currentBase?.name;
   const currentOrganisationId = globalPreferences.organisation?.id;
-  const currentOrganisationBases = globalPreferences.availableBases;
+
+  const AllBasesOfCurrentOrg = useQuery<AllAvailableBasesQuery>(ALL_BASES_OF_CURRENT_ORG_QUERY, {
+    variables: {
+      orgId: currentOrganisationId,
```
HaGuesto, line 130: in the functional test an error toast appears on a reload since the currentOrganisationId is not set while useLoadAndSetGlobalPreferences is still loading.

is there a way quick way to only execute the query one useSetAndLoadGlobalPreferences is done?

HaGuesto (approved):
## functionally and code review generally passed.

There are two small bits that you could fix before merging, but I do not need to review again.

HaGuesto:
@fhenrich33 can you fix the build error, pls.

PR #1742, opened 2024-11-17
Base Switcher

fhenrich33:
firefox_tqELfpimMk

firefox_VDl4kIa2xi


fhenrich33:
[2.0 FE Menu: Add Base Switcher](https://trello.com/c/8tc3SIzg/1597-20-fe-menu-add-base-switcher)

pylipp (changes requested):
**Functional review**
1. On my desktop (1366x768), the settings section of the menu becomes unreachable when I have a menu with more than two subitems opened: 
2024-11-25-111409_279x631_scrot
2. When in base 2, I open the base switcher and click "Switch" without having selected anything from the list, I'm directed to `/bases/1` that I'm not authorized for (as dev_headofops@boxcare.org user). Should the Switch-button be disabled until a selection is made?
3. Should the current base be excluded from the selection? Seems logical to me because the user is already there.

front/src/components/HeaderMenu/BaseSwitcher.tsx
```diff
@@ -0,0 +1,64 @@
...
+    navigate(`/bases/${value}${currentPath}`);
+    onClose();
+  };
+
+  return (
+    <>
+      <Modal isOpen={isOpen} onClose={onClose}>
+        <ModalOverlay />
+        <ModalContent>
+          <ModalHeader>Switch Bases to</ModalHeader>
```
pylipp, line 37: I don't have access to the designs yet, but because a user can only use the app in one base at a time, it might make sense to change this to "Switch Base to"

fhenrich33: @aerinsol thoughts?

aerinsol: Agree! Typo on my end, will update in designs.

fhenrich33: @pylipp I didn't manage to reproduce 1. on either Chrome or Firefox:

msedge_K9csTlEhf5

fhenrich33:
@pylipp I didn't manage to reproduce the issue with a 1366x768 on either Chrome or Firefox:

msedge_K9csTlEhf5

aerinsol:
Looks good @fhenrich33 , however don't forget to merge/rebase in the updates in master that remove the beta tag from Manage Shipments and Manage Agreements.

fhenrich33:
@aerinsol Done, and updated GIFs!

front/src/components/HeaderMenu/MenuDesktop.tsx
```diff
@@ -18,12 +18,16 @@ import MenuIcon, { Icon } from "./MenuIcons";
...
+import { useBaseIdParam } from "hooks/useBaseIdParam";
 
 function MenuDesktop({ menuItemsGroups }: IHeaderMenuProps) {
   const { isOpen, onOpen, onClose } = useDisclosure();
   const { handleLogout } = useHandleLogout();
+  const { baseId: currentBaseId } = useBaseIdParam();
   const { globalPreferences } = useContext(GlobalPreferencesContext);
   const baseName = globalPreferences.selectedBase?.name;
+  const currentOrganisationHasMoreThanOneBaseAvailable =
+    (globalPreferences.availableBases?.filter((base) => base.id !== currentBaseId).length || 0) > 1;
```
pylipp, line 30: @fhenrich33 why do you filter the current base? For a 2-base organisation this results in `currentOrganisationHasMoreThanOneBaseAvailable` being false, and then the base switcher button is unclickable.

I think this is sufficient:    `(globalPreferences.availableBases?.length || 0) > 1`

Same in MenuMobile.tsx

pylipp:
thanks @fhenrich33, good work implementing the points 2.-4. [mentioned before](https://github.com/boxwise/boxtribute/pull/1742#pullrequestreview-2457981573). Just one final adjustment requested :)  
Weird that the issue about the cut-off menu bottom is not reproducible, I have it in both Firefox and Chrome but maybe it's due my Linux window-manager (i3) :nerd_face:

front/src/components/HeaderMenu/MenuDesktop.tsx
fhenrich33: Whoops, missed a `= `for the `>= 1` condition. Thanks!

(Also this is a way to tell codewise that we don't want to show the current base as an option.)

fhenrich33:
> thanks @fhenrich33, good work implementing the points 2.-4. [mentioned before](https://github.com/boxwise/boxtribute/pull/1742#pullrequestreview-2457981573). Just one final adjustment requested :) Weird that the issue about the cut-off menu bottom is not reproducible, I have it in both Firefox and Chrome but maybe it's due my Linux window-manager (i3) 🤓 so let's not worry

Could a window manager mess with this? Tried on Windows 11, macOS (forgot which but it's recent), and KDE Plasma, and that glitch didn't show up 😮

pylipp (approved):

front/src/components/HeaderMenu/MenuDesktop.tsx
pylipp: > Also this is a way to tell codewise that we don't want to show the current base as an option.

thx for the fix; I just thought that the options are constructed in the [BaseSwitchter](https://github.com/boxwise/boxtribute/pull/1742/files#diff-4f046224a7d284e2ab9daecc2f6be08e04b0bc9632e93a2ae613274fee0e35c4R25)

front/src/components/HeaderMenu/MenuDesktop.tsx
```diff
@@ -5,54 +5,38 @@ import {
...
+function MenuDesktop({ menuItemsGroups }: IHeaderMenuProps) {
+  const { isOpen, onOpen, onClose } = useDisclosure();
+  const { handleLogout } = useHandleLogout();
+  const { baseId: currentBaseId } = useBaseIdParam();
+  const { globalPreferences } = useContext(GlobalPreferencesContext);
+  const baseName = globalPreferences.selectedBase?.name;
+  const currentOrganisationHasMoreThanOneBaseAvailable =
+    (globalPreferences.availableBases?.filter((base) => base.id !== currentBaseId).length || 0) >=
+    1;
+  const allowMultipleAccordionsOpen = window.screen.availHeight > 1080;
```
HaGuesto, line 32: please switch to useMediaQuery from chakra.

front/src/components/HeaderMenu/BaseSwitcher.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const firstAvailableBaseId = currentOrganisationBases?.find((base) => base)?.id;
+  const [value, setValue] = useState(firstAvailableBaseId);
+
+  // Need to set this as soon as we have this value available to set the default radio selection.
+  useEffect(() => {
+    setValue(firstAvailableBaseId);
+  }, [firstAvailableBaseId, currentBaseId]);
+
+  const switchBase = () => {
+    const currentPath = window.location.pathname.split("/bases/")[1].substring(1);
```
HaGuesto, line 36: please switch to location variable from react-router-dom

HaGuesto (approved):
## code and functional review passed

All points from @pylipp were adressed. There is only a small change to remove the use of window variable, but I do not need to review this again.

PR #1752, opened 2024-11-19
fix default timerange filter in statviz

HaGuesto:
This is a fix for the bug mentioned [here](https://trello.com/c/XUrjWcSI).

Reason for the error: The state was used and edited before passing the state to the state update function. This is not best practice and did somehow not work on the initial load.

@pylipp you wanted to create a bug ticket for it, could you assign this PR to it?


pylipp:
[2.0: statviz graphs initially empty even though valid data available](https://trello.com/c/s0melRyu/1610-20-createdboxes-graph-initially-empty-even-though-valid-data-available)

pylipp (approved):

PR #1753, opened 2024-11-19
Experiment with react-intl

pylipp:
see also #1794 

- Add dependencies
- Simple first version
- Use formatjs extraction-distribution workflow


pylipp:
Closing since associated ADR has been merged. This PR shall be revisited when implementation is due (however agreement is to use react-i18next, not react-intl).

PR #1794, opened 2024-11-26
Add i18n ADR

pylipp:
[Doc as rendered markdown](https://github.com/boxwise/boxtribute/blob/adr/i18n/docs/adr/adr_i18n.md).

Superceding [this doc](https://docs.google.com/document/d/1dVGQIl_8fO5_JDpVCg9g_GVG7Zl0sbx5VqUmAcV5VO0/edit?pli=1&tab=t.0#heading=h.mfo0lhwbjnsb).

### DIFF
--- a/docs/adr/adr_i18n.md
+++ b/docs/adr/adr_i18n.md
@@ -123,11 +123,17 @@ Using a TMS might not be worth the additional cost and maintenance effort, given
 
 ### Initial steps
 
-- Find all text components in FE code, and wrap them such that they can be extracted and replaced
+**Infrastructure** (20h)
 - Install i18n package
-- Extract messages and actually translate them (into the desired languages)
 - Set up workflow such that newly added text is auto-extracted (warn about missing translation)
 - Ensure that translation files are shipped with the deployed FE bundle
+- OPTIONAL (extra 20h): integrate TMS (cloud-repo and/or cloud-CI integration)
+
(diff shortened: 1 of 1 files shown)


aerinsol (commented):
Looks good. Might be worth revisiting TMS from a non-developer POV. Would integrating with a TMS add time to the dev estimate @pylipp ?

pylipp:
> Might be worth revisiting TMS from a non-developer POV. Would integrating with a TMS add time to the dev estimate?

@aerinsol yes, integrating a TMS would add time to the initial dev estimate I think. The suggested TMSs are cloud-based, so there needs to be some integration into with the repo and/or CI.

> I vouch for react-i18next, as it's the most widely supported by the React community overall. A plethora of plugins and code are built on top of it, and it's my personal preference because it's decoupled from React implementation-wise (compared to react-intl).

Thanks for your assessment @fhenrich33 it's valid point in favor of react-i18next to be framework-agnostic (even though we are very unlikely to move away from react, as Roanna said during SU). @HaGuesto also named react-i18next as his favorite.

pylipp:
@aerinsol I just updated the [section about the effort estimation](https://github.com/boxwise/boxtribute/blob/adr/i18n/docs/adr/adr_i18n.md#consequences--effort-estimation) because I noticed a confusing missing newline in there, too

aerinsol (approved):

PR #1823, opened 2024-12-04
Use jotai to store global preferences

fhenrich33:
Rendered ADR: https://github.com/boxwise/boxtribute/blob/9ead086f0b0a36df2adbadf951a2393150e6b81f/docs/adr/adr-global-state-management.md

Note: Playwright tests will cover these changes.


fhenrich33:
[2.0 FE: Replace GlobalPreferencesContext](https://trello.com/c/CK6TtXPS/1495-20-fe-replace-globalpreferencescontext)

HaGuesto:
please update this [ADR](https://github.com/boxwise/boxtribute/blob/master/docs/adr/adr_frontend_global_state.md) with this PR

docs/adr/adr-global-state-management.md
```diff
@@ -0,0 +1,167 @@
+# ADR: Use Jotai for Global State Management
```
HaGuesto, line 1: @fhenrich33 could you add something on the reactivevariable of Apollo too in here. We tried it out for a few small feature and would love to know how it compares with jotai.

aerinsol:
Here I think the comparative analysis could be improved in a way similar to what was done with Playwright. The current format makes it hard to assess the cost/benefit of the various solutions.

fhenrich33:
> Here I think the comparative analysis could be improved in a way similar to what was done with Playwright. The current format makes it hard to assess the cost/benefit of the various solutions.

@aerinsol Legend just updated so I have to make some changes taking that into account, along with the inclusion of Apollo's Reactive Variables.

PR #1824, opened 2024-12-04
Receiving and Sorting tabs for Manage Shipments, default sort to last updated

fhenrich33:
firefox_wEgMAYMgi1


pylipp:
One quick remark, telling from the GIF: the shipments currently listed in Receiving should be in Sending and vice versa (you can tell from the shipment identifier; the last part is `<first two letters of sending base>x<first two letters of receiving base>`, so since the user is in Thessaloniki, when in the Receiving tab, the `TH` should be second in the identifier). I didn't find the sorting logic in the code but basically the Receiving tab should contain all shipments with `shipment.targetBase === currentBase`

fhenrich33:
> One quick remark, telling from the GIF: the shipments currently listed in Receiving should be in Sending and vice versa (you can tell from the shipment identifier; the last part is `<first two letters of sending base>x<first two letters of receiving base>`, so since the user is in Thessaloniki, when in the Receiving tab, the `TH` should be second in the identifier). I didn't find the sorting logic in the code but basically the Receiving tab should contain all shipments with `shipment.targetBase === currentBase`

Thanks @pylipp. That also was an opportunity to revisit the overall direction calculation for the shipments now that we support intra-org shipments. I think this should do the trick.

HaGuesto (approved):
well done. 

# code review and functional review passed

aerinsol:
@fhenrich33 @HaGuesto  the count of shipments at the top tab should exclude canceled and completed shipments.

pylipp:
> @fhenrich33 @HaGuesto the count of shipments at the top tab should exclude canceled and completed shipments.

addressed in #1834

PR #1828, opened 2024-12-05
re-write ALL_SHIPMENTS query

HaGuesto:
since the ALL_SHIPMENTS_QUERY was just used in the shipments overview and does not need the recursive query of shipment detail, I only had to separate out the query from the existing fragment.

ALL FE tests passed locally


HaGuesto:
https://trello.com/c/NLpDI8gO

pylipp (approved):

pylipp:
### Functional test passes

Thx for looking into this!

PR #1831, opened 2024-12-06
Add first concept for link-sharing

pylipp:
[Rendered document](https://github.com/boxwise/boxtribute/blob/draft-for-link-sharing/docs/auth/public_sharing_of_statistics.md#draft-for-prototypical-implementation).

[PasswordPusher ADR](https://github.com/boxwise/boxtribute/blob/draft-for-link-sharing/docs/adr/adr_link-sharing-backend.md)

It got more detailed in the BE part. For the UI it's rather some questions/ideas.

LMK if parts are not clear.


pylipp:
[2.0: Draft concept for public link sharing of statviz](https://trello.com/c/TVTCaM47/1628-20-draft-concept-for-public-link-sharing-of-statviz)

docs/auth/public_sharing_of_statistics.md
```diff
@@ -100,3 +100,108 @@ From [Slack Canvas](https://boxwise.slack.com/docs/T99PBKNTU/F06QQS70XQT):
...
+```
+
+#### Database
+
+- new `link_sharing` table to store meta-data (base ID, URL query parameters, expiration date, creation date, creator)
+- corresponding peewee model
+
+### Open questions
+
+- how do avoid misuse of the link, or the exposed public GraphQL endpoint (DDoS mitigation)?
```
jamescrowley, line 205: Could look at rate limiting server side based on the link hash being passed in. Could also consider https://cloud.google.com/armor/docs

docs/auth/public_sharing_of_statistics.md
```diff
@@ -100,3 +100,108 @@ From [Slack Canvas](https://boxwise.slack.com/docs/T99PBKNTU/F06QQS70XQT):
...
+1. After one week, the link expires. When the link is opened, a short message is displayed.
+
+## Architecture
+
+### Fundamental assumptions
+
+- the shared data will be live (as opposed to "freezing" the data to the time of link creation)
+- link format is `.../<code> (can't use `.../bases/X/statviz`: insecure because it's easy to navigate to other bases and view their data; also not unique (can't expire))
+- the expiration time of the link is one week. Later we can make is customizable, and/or add an action to invalidate a created link
+- we expose the public app in the `api` GAE service. Hence it won't interfer with the main `app` service
```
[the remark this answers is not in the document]
pylipp: No, we're not doing this yet. It's a suggestion for the link-sharing use case.

docs/auth/public_sharing_of_statistics.md
```diff
@@ -100,3 +100,108 @@ From [Slack Canvas](https://boxwise.slack.com/docs/T99PBKNTU/F06QQS70XQT):
...
+
+### Effects on full-stack
+
+- there has to be a new action-based permission to allow link-creation for certain usergroups
+- there has to be a public FE and a public GraphQL endpoint to fetch data
+- the public FE is hosted under `api.boxtribute.org/shared` and deployed with the `deploy-api` CircleCI job (depends a new `build-statviz` job)
```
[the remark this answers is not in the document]
pylipp: I suggested this for simplicity (less devops effort because we don't need to establish a new subdomain nor any new GAE services). The endpoints we currently have for the `api` subdomains are the GraphQL one at `/` and the documentation at `/docs`.
For a separation of concerns it probably makes sense to have a dedicated subdomain for link-sharing.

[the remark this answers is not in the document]
jamescrowley: @pylipp got it. Is there any other way we can achieve that? If the back end link sharing is going to be part of the standard API on the api subdomain, then any reason the link sharing can't just operate as part of the same FE code base? Or am I misunderstanding here?

docs/auth/public_sharing_of_statistics.md
```diff
@@ -100,3 +100,108 @@ From [Slack Canvas](https://boxwise.slack.com/docs/T99PBKNTU/F06QQS70XQT):
...
+
+#### Database
+
+- new `link_sharing` table to store meta-data (base ID, URL query parameters, expiration date, creation date, creator)
+- corresponding peewee model
+
+### Open questions
+
+- how do avoid misuse of the link, or the exposed public GraphQL endpoint (DDoS mitigation)?
+- can FE resolve a URL like `.../<code>` into a route like `.../<code>/bases/X/statviz` on the FE **and** process the returned data as if it was a GraphQL response?
```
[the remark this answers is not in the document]
pylipp: We (Hans, Felipe, me) talked about this yesterday; it seems to be feasible but requires adjustments to existing queries.

I'll try to explain it like this: in v2, when the user visits e.g. `/bases/2/statviz`, it will load the page and send a GraphQL request to the BE. A GraphQL response is returned and used to build the page.
For the scenario of link-sharing I was thinking whether the following is possible:
1. user visits link at `.../<code>`
2. FE sends request incl the code
3. BE checks the code. Since it has saved the type of requested data at the time of link creation, it can now look up the data and return it in the response, incl the corresponding route for FE
4. FE goes to the returned route **and** plugs in the returned data (instead of issuing another GraphQL request)
You can see this as a mind experiment of mine rather :smile: does it make sense to you?

The approach we agreed on is different though:
1. FE sends request with code
2. BE validates code and sends metadata route
3. FE goes to route and runs GraphQL queries to fetch data

aerinsol (approved):
LGTM, would like to hear more technical commentary on @jamescrowley from the answer, and both @jamescrowley and @HaGuesto on PP analysis.

docs/auth/public_sharing_of_statistics.md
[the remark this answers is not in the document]
pylipp: Sorry, this comment slipped through somehow. @jamescrowley 

> any reason the link sharing can't just operate as part of the same FE code base? 
Do you want the link-sharing back-end to be part of the FE code base? Or do you want the public FE for shared links to be in the same code base as the standard FE?

One more thing re the new public BE: When we introduce a new `/shared` route on the api subdomain, we can (and will) use a new schema there, different from the standard API schema (hosted on `/`).

docs/auth/public_sharing_of_statistics.md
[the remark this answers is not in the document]
HaGuesto: @pylipp sorry, but did we really agree on that? Performance-wise this is not a nice solution since we are creating a waterfall, e.g. you have to wait for two requests instead of just one. This is rather how I would do it:

1. FE sends request with code
2. BE validates code and sends back information which view is behind that code and the corresponding requested data needed to show the view, e.g.
```
{
    resolveSharableLink {
        view: "statviz",
        pageData: {
```
3. FE should either render the returned view under the same route or redirect (not sure what is easier). 
4. Since the data is already cached in the FE from the first request, no additional request to the BE should happen.

docs/auth/public_sharing_of_statistics.md
```diff
@@ -100,3 +100,153 @@ From [Slack Canvas](https://boxwise.slack.com/docs/T99PBKNTU/F06QQS70XQT):
...
+
+#### PP Alternatives
+
+According to [this list](https://alternativeto.net/software/passwordpusher/), [onetimesecret](https://docs.onetimesecret.com/docs/rest-api) can be considered an alternative but it has the immense drawback that a secret can only be viewed once.
+
+Other services in the list like vanish.so are either fairly unpopular and/or dont' provide any API.
+
+The following uses PP as exemplatory service.
+
+#### Usage as link-sharing back-end
```
HaGuesto, line 227: From my POV, PP should only replace DB sturcture and business logic in the BE. It should not change the flow of link sharing. I have two reasons for that:
1. If we ever need to implement it ourselves it is easier if we do not have to change flows.
2. in your process FE takes care of creating the secure link and thereby of authorization. However, you should not trust the FE to this. (If you want I can name examples for the future where you could easily fake data for the code generation).

Therefore the flows should look like from my POV like this:

### link generation
1. FE sends request to BE to generate link for public sharing (including the view to be shared, at some point the expiration time, maybe even some parameter for the view, e.g. filter parameter states)
2. BE generates code 
     a. either in its own way
     b. by sending a request to the PP REST API. We would need an admin user + password for this and possibly a passphrase if we think we have to secure the data stored in the PP. 
3. The BE returns the generated code.

FYI. This way we also have access to the stats of PP since all links would be generated under our admin account.
 
### link retrieval
1. FE sends request with code to BE.
2. BE validates code
    a. in its own way
    b. by retrieving the information from PP
3. BE sends back information which view is behind that code and the corresponding requested data needed to show the view, e.g.
{
    resolveSharableLink {
        view: "statviz",
        pageData: {
4. FE should either render the returned view under the same route or redirect (not sure what is easier).
5. Since the data is already cached in the FE from the first request, additional request to the BE only happen if some data was missing.

docs/auth/public_sharing_of_statistics.md
```diff
@@ -100,3 +100,153 @@ From [Slack Canvas](https://boxwise.slack.com/docs/T99PBKNTU/F06QQS70XQT):
...
+```graphql
+type Mutation {
+  createShareableLink(creationInput: LinkCreationInput): ShareableLinkCreationResult
+}
+
+input LinkCreationInput {
+  baseId: Int!
+  # everything after /
+  urlTail: String
+  page: ShareablePage!
```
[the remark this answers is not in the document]
HaGuesto: also can you rename this from page to view in general?

docs/auth/public_sharing_of_statistics.md
```diff
@@ -100,3 +100,153 @@ From [Slack Canvas](https://boxwise.slack.com/docs/T99PBKNTU/F06QQS70XQT):
...
+union ShareableLinkCreationResult = ShareableLink | InsufficientPermissionError
+
+type ShareableLink {
+  id: ID!
+  code: String!
+  # to be inserted after /shared/<code>/
+  urlTail: String!
+  validUntil: Datetime
+  createdOn: Datetime!
+  createdBy: User
```
HaGuesto, line 180: probably view shoiuld be in here too.

docs/auth/public_sharing_of_statistics.md
```diff
@@ -100,3 +100,153 @@ From [Slack Canvas](https://boxwise.slack.com/docs/T99PBKNTU/F06QQS70XQT):
...
+}
+```
+
+For the public GraphQL schema:
+
+```graphql
+type Query {
+  resolveShareableLink(code: String!): ShareableLinkResult
```
HaGuesto, line 188: If you read my other comments, the SharableLinkResult should include SharablePage but also the page data necessary to show the page.

docs/auth/public_sharing_of_statistics.md
```diff
@@ -100,3 +100,153 @@ From [Slack Canvas](https://boxwise.slack.com/docs/T99PBKNTU/F06QQS70XQT):
...
+
+  " All statistics queries must have a `code` parameter for validation in the public BE "
+  createdBoxes(baseId: Int!, code: String): CreatedBoxesData
```
HaGuesto, line 191: agreed.

docs/auth/public_sharing_of_statistics.md
```diff
@@ -100,3 +100,153 @@ From [Slack Canvas](https://boxwise.slack.com/docs/T99PBKNTU/F06QQS70XQT):
...
+
+#### Advantages
+
+- we can save the entire BE business and database logic around creating and storing shared links (incl. not having to take care of link expiration; plus having the passphrase as extra security mechanism (albeit it can be inspected easily when the public FE sends a request to the public BE))
+- PP provides unlimited creation of links, however with rate-limit
+- PP is open-source and [trusted by many organisations and considered secure](https://docs.pwpush.com/docs/faq/#trust-is-a-concern--why-should-i-trust-and-use-password-pusher)
+
+#### Disadvantages
+
+- less data insight: we don't have control over created links (no information about amount or which pages are being shared; at least not without further tracking)
```
HaGuesto, line 243: not true if we are creating the links from the BE which we should.

docs/auth/public_sharing_of_statistics.md
```diff
@@ -100,3 +100,153 @@ From [Slack Canvas](https://boxwise.slack.com/docs/T99PBKNTU/F06QQS70XQT):
...
+- increased complexity: we add a second interface that FE has to communicate with; and we have to define a data structure to store information on PP
```
HaGuesto, line 244: not true if only BE is communicating with it.

HaGuesto:
I took a look at the PP REST API. Here my thoughts:

+ it does what we need and is not much more complex than this.
+ we can integrate it in a modular way so that we can easily replace it by another provider or our own structure if we really run into problems at some point.
+ the integration is very simple (PP account creation and then API token creation, implementing two kind of requests from the BE, adding secrets to CircleCI) and will save us time to get first prototypes out.
- I didn't see what the rate limit is but from the documentation it sounds like we should not run into problems in the near future. 
- repo is active, but mainly maintained by one person.
- we need to be careful what data we want to store in the links. There shouldn't be user information in there, maybe a user id but only if we really need it. 

I'd go for it. @pylipp feel free to book me if you want to talk about it. I also pinged @jamescrowley on it.

docs/auth/public_sharing_of_statistics.md
pylipp: okay, I agree to still have "our" GraphQL API to generate and retrieve links; less of a hazzle in the case of moving away from PP

pylipp: > in your process FE takes care of creating the secure link and thereby of authorization. However, you should not trust the FE to this.

I don't fully understand. Is it connected to my sentence about "the passphrase can be inspected easily when the public FE sends a request to the public BE"? Someone could fake a request with arbitrary parameters to PP, and then use the generated link on our BE to fetch any data.

docs/auth/public_sharing_of_statistics.md
pylipp: If we decide to have a single `resolveShareableLink` query which includes the view data in its response, we don't need further public queries at all.

docs/auth/public_sharing_of_statistics.md
pylipp: We should reconsider how useful and sensible storing of URL details is. A URL might be specified to a route in one FE but actually have a different in a second FE. So I'd rather store the type of view only, and have a small mapping in each FE to find the correct route for a view type. This also helps avoid the case that a FE route is changed but still part of the metadata of a long-time valid shared link.

docs/auth/public_sharing_of_statistics.md
pylipp: still to evaluate: one PP user for all envs, or multiple (at least two for prod, and demo/staging/dev)

pylipp:
@HaGuesto Thanks for the feedback and the new idea of integrating PP, I quite like it too. I directly updated the doc acc. a couple of comments.

docs/auth/public_sharing_of_statistics.md
pylipp: Clarified with @HaGuesto : URL parameters should be stored but not route details

pylipp:
Related [follow-up discussion on Slack](https://boxwise.slack.com/archives/CB1UYDGLU/p1737366509715589?thread_ts=1736852594.802029&cid=CB1UYDGLU)

pylipp:
[2.0: Draft concept for public link sharing of statviz](https://trello.com/c/TVTCaM47/1628-20-draft-concept-for-public-link-sharing-of-statviz)

PR #1832, opened 2024-12-06
Playwright tests

fhenrich33:


aerinsol:
@fhenrich33 can you post a link in the description with a fully rendered file, like what @pylipp does with his?

aerinsol:
Good ADR @fhenrich33. There are some missing sentences in the context section that should be cleaned up, and I'd like that section on MSW to clarify that you're referring to the current production or development environment (or a specific part of it) rather than just saying "we". 

Otherwise, nice work, please request reviews whenever you're ready.

shared-components/statviz/components/visualizations/stock/StockDataFilter.test.tsx
jlin95, line 1: Are we removing the old tests and expecting to rewrite them again using Playwright? @fhenrich33

front/src/views/Boxes/components/transformers.ts
```diff
@@ -10,13 +10,13 @@ export const boxesRawDataToTableDataTransformer = (boxesQueryResult: BoxesForBox
       (element) =>
         ({
           labelIdentifier: element.labelIdentifier,
-          product: element.product!.name,
-          productCategory: element.product!.category.name,
-          gender: element.product!.gender,
+          product: element.product?.name,
```
jlin95, line 13: quick qn: Why are we removing the non-null assertion operator here? Is it possible that product/location can be optional?

fhenrich33: Good that you brought that up!

`!` aka [non-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator) is a TypeScript compiler flag of sorts, telling TS to exclude `null` and `undefined` from the type of this property, or in a way telling it's okay if they end up being those types. 

But since this piece of code blew up for me using mock data, the right thing to do is to use `?` or the optional chaining operator to avoid exceptions on trying to access properties inside an `undefined` or `null` object/property.

shared-components/statviz/components/visualizations/stock/StockDataFilter.test.tsx
fhenrich33: Yes, for rewriting tests with Playwright; no, for rewriting tests that already exist. I'm almost finished with that right now!

docs/adr/adr_frontend_tests.md
```diff
@@ -0,0 +1,79 @@
...
+     - Good community support and documentation.
+     - Easy to set up and use for isolated testing.
+   - Cons:
+     - Limited to testing individual components and their interactions.
+     - Does not provide comprehensive application testing capabilities (in comparison to Playwright).
+     - Runs against a mocked DOM structure (jsdom) instead of a real browser, which leads to flakiness and unreliability.
+     - No built-in support for cross-browser testing.
+
+2. **Playwright**
+   - Pros:
```
HaGuesto, line 40: I thought the main Pro argument is that you can easily generate mock data @fhenrich33 ?

front/README.md
```diff
@@ -82,35 +84,104 @@ Afterwards:
 
 ## Testing
 
-Testing is done with React Testing Library and Jest.
+Testing is done with [Playwright](https://playwright.dev/) for application, integration, and component testing.
```
HaGuesto, line 87: feels like you have pasted the same section twice here. Also please ensure that the old section on react testing library is still in, but that it is clear that we are slowly deprecating it.

fhenrich33: Did you read from the README in the PR description section? I didn't find any dupes there.

Also, I'm replicating the entire test suite where we used React Testing Library with Playwright, so I don't think it makes sense to keep it in that case.

aerinsol:
Good clear updates to the comparative evaluation. Nice work.

front/README.md
fhenrich33: Apologies @HaGuesto , it seems that the description didn't updated properly.

docs/adr/adr_frontend_tests.md
fhenrich33: This would be the benefit of [gql.tada](https://gql-tada.0no.co/), used with MSW request handlers. Playwright does have a very similar mocking capability as MSW, but we use the latter to allow us to mock yet-to-be-implemented schemas in the future.

aerinsol (approved):

front/README.md
```diff
@@ -84,35 +84,104 @@ Afterwards:
 
 ## Testing
```
HaGuesto, line 85: Please update this section to mention that we are slowly deprecating React Testing Library and move slowly to playwright. So, keep the old instructions in.
Also mention that all new test should be written in playwright.

front/README.md
```diff
@@ -84,35 +84,104 @@ Afterwards:
...
+We use [Mock Service Worker](https://mswjs.io/) (MSW for short) to mock data in a way that faithfully mimics the real server. It's also very useful to mock GraphQL Schema bits that aren't implemented in the production backend yet to build new features on top of it.
 
-To eliminate repetitive code, a custom renderer was built in `front/src/tests/test-utils.js`. It allows developers to render a component in a test environment where chakra, Apollo and Routes are wrapped around it. The utility also exports the entire react testing library, so you should import from this utility instead of `@testing-library/react`. See `EditBox.test.js` for examples of the custom renderer's use.
+Application test specs are inside the `/tests` folder. Other tests are colocated with the files that are the subject of what is being tested.
 
-Tests and test coverage can be run with the following command:
+See Playwright's docs to see how to make the most of its API to help you build test cases e.g. [recording tests by using the app](https://playwright.dev/docs/codegen-intro).
```
HaGuesto, line 96: please add two steps how to install it locally. To me pnpm install was not enough. It also asked me to install the browsers. Since downloading all browsers were quite large I only installed chromium with:

`pnpm exec playwright install chromium --with-deps`

front/README.md
```diff
@@ -84,35 +84,104 @@ Afterwards:
...
 
-```sh
-# run tests
-docker compose exec front pnpm test
+### Fixtures and mock data
+
+Mock data is located inside `tests/fixtures.ts`, and is usually retrieved from the development backend calls to speed up testing and development since fixtures for that data were already written. As long as both the front-end and backend match the GraphQL Schema, we can be sure that the mocked data will match real data consumed in the app. You can add more fixtures by hand or by using the real API results that you can get from the Network tab inside Devtools in the browser.
+
```
HaGuesto, line 100: Please add more detailed steps or a short video how you send the query to the BE, retrieve the test data and then add it as a fixture.

front/README.md
```diff
@@ -84,35 +84,104 @@ Afterwards:
...
 pnpm test
 
+# Run tests in UI mode (opens a Chromium browser to visually run and debug tests)
+pnpm test:ui
+
 # test coverage
+pnpm test:coverage
+
+# run tests inside Docker
+docker compose exec front pnpm test
```
HaGuesto, line 125: This did not work:
1. no Browsers were installed in the docker container -> add chromium Browser installation to Dockerfile
2. the front docker container is located at /front, but `pnpm test`  only works if you call it from the repo root with `docker compose exec front pnpm -w run test`

.circleci/config.yml
```diff
@@ -138,37 +138,47 @@ jobs:
           paths:
             - front/dist
 
+  lint-front:
```
HaGuesto, line 141: revert the circleci config so that the tests with react-testing library still run.

front/README.md
```diff
@@ -84,35 +84,104 @@ Afterwards:
...
+### Fixtures and mock data
+
+Mock data is located inside `tests/fixtures.ts`, and is usually retrieved from the development backend calls to speed up testing and development since fixtures for that data were already written. As long as both the front-end and backend match the GraphQL Schema, we can be sure that the mocked data will match real data consumed in the app. You can add more fixtures by hand or by using the real API results that you can get from the Network tab inside Devtools in the browser.
+
+To label the fixture/mock data, follow the convention of placing it as `userName: { graphqlOperationName: { modifierEGbaseId: data } }`. Then use it inside the MSW handler at `mswHandlers.ts` (also to export the handler at the bottom of the file).
+
+> [!IMPORTANT]  
+> You must place this entry on your `.env` file for the tests to run against MSW and mock data in `tests/fixtures.ts`.
+>
+> `USE_MSW=true`
```
HaGuesto, line 106: It would be nice to know a bit more how we can use MSW better btw. What else it might be useful for.

HaGuesto (commented):
@fhenrich33 I would like you to do the following to get this PR merged asap:

1. add the react-testing library tests back in so that we do not have to migrate all tests at once now.
2. revert the circleCI config that it still reports on the react-testing library tests. we will add in playwright once all of us have written some first tests with it.
3. please make pnpm work in the docker dev set-up and update the corresponding README section
4. could you move the tests to the components they belong.
5. rename the tests folder at repo root to playwright or something else less generic

PR #1841, opened 2024-12-09
Use when:true in CircleCI jobs

pylipp:
Acc. to warning in CircleCI web UI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -174,7 +174,7 @@ jobs:
       - run:
           name: Show memory usage of FE tests in Bytes
           command: cat /sys/fs/cgroup/memory/memory.max_usage_in_bytes
-          when: always
+          when: true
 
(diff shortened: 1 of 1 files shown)


HaGuesto (approved):

HaGuesto (commented):
@pylipp the circleci pipeline throws an error, probably due to invalid yaml file. There is a way to check if the yaml file works online

https://support.circleci.com/hc/en-us/articles/360006735753-How-to-validate-your-CircleCI-configuration

pylipp:
Thanks for looking into this, and providing the validation link. Actually there's nothing to do from our side.
CircleCI phrased the notification in a confusing way but [clarified now](https://discuss.circleci.com/t/scope-of-when-always-config-syntax-change/52338) that the change is about `when`s in `workflows`, not in `steps`.

PR #1845, opened 2024-12-10
Standarize tabs styling + uppercase section headers in sidebar

fhenrich33:
Screen Shot 2024-12-11 at 12 05 49

Screen Shot 2024-12-11 at 12 05 57

Screen Shot 2024-12-11 at 12 06 16

### DIFF
--- a/front/src/components/HeaderMenu/MenuDesktop.tsx
+++ b/front/src/components/HeaderMenu/MenuDesktop.tsx
@@ -72,7 +72,7 @@ function MenuDesktop({ menuItemsGroups }: IHeaderMenuProps) {
           ))}
         </Accordion>
         <Accordion marginTop={"auto"}>
-          <strong style={{ marginLeft: "1rem" }}>Settings</strong>
+          <strong style={{ marginLeft: "1rem", textTransform: "uppercase" }}>Settings</strong>
           <AccordionItem>
             <AccordionButton
               gap={3}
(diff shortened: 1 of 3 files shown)


fhenrich33:
[2.0 FE Bug: Strange Tab Behavior Connecting Manage Shipments tab style with the tab style inside one of BoxView components](https://trello.com/c/gThsgaxa/1636-20-fe-bug-strange-tab-behavior-connecting-manage-shipments-tab-style-with-the-tab-style-inside-one-of-boxview-components)

fhenrich33:
> @fhenrich33 , after looking at this, I'd like to change the Move/Transfer tabs back to Pascal case. That does keep the text case different from QR Scanner but that is OK. I think those internal tabs will end up being on a case-by-case basis.
> 
> If you're planning/already putting the large page tabs (e.g., SENDING/RECEIVING) into a global component I'm OK with having them all caps and setting that as default.

I don't think it's worth having this as a component just for that since we use wildly varying props and conditionals throughout the app with the tabs components. We would be abstracting just for the sake of abstraction and losing flexibility.

fhenrich33:
[2.0 Section Headers in the Sidebar Should be Capitalized #tiny](https://trello.com/c/IySItAmy/1639-20-section-headers-in-the-sidebar-should-be-capitalized-tiny)

HaGuesto (approved):
## code review passed

pylipp (approved):

PR #1852, opened 2024-12-12
Standardize Desktop Behavior of QR Scanner - should NOT show it as home page. Show StatViz Dashboard instead

HaGuesto:


HaGuesto:
https://trello.com/c/Dk3IRAva

front/src/views/ErrorView/ErrorView.tsx
```diff
@@ -0,0 +1,24 @@
+import { useAuth0 } from "@auth0/auth0-react";
+import { Alert, AlertIcon, Button, Container } from "@chakra-ui/react";
+
+interface ErrorViewProps {
+  error: string | undefined;
```
fhenrich33, line 5: Nit: this could be `error?: string;`

pylipp (commented):
**Functional testing**
1. went to `/bases/2/foo` -> correctly redirected to `/bases/2/statviz`
2. went to `/foo` -> correctly redirected to `/bases/2/statviz`
3. went to `/bases/5/boxes` -> correctly redirected to error view with logout button. however two error toast are shown. Same with `/bases/1/boxes`

So far so good :+1: will check again when mobile-only redirection to /qrreader is implemented.

front/src/App.tsx
```diff
@@ -116,21 +112,11 @@ function App() {
...
-    return (
-      <>
-        <Alert status="error">
-          <AlertIcon />
-          {error}
-        </Alert>
-        <Button onClick={() => logout()}>Logout</Button>
-      </>
-    );
+    return <ErrorView error={error} />;
```
pylipp, line 115: Thanks for the subcomponent :+1:

fhenrich33 (commented):
Will this comment be addressed in this PR? https://trello.com/c/Dk3IRAva#comment-6746d22bbe73f4ae229adc8c

Double toast in the error page aside, LGTM! 👍

Waiting for the mobile redirect to `/qrreader` for another go.

HaGuesto:
@pylipp The reason why it throws two error toasts is that the number of error toasts depend on the failing graphql requests.
When you request `/bases/5/boxes` than two graphQL requests are sent to the backend - one to query the boxes and one to query the options for the action buttons. 
Fixing it would maybe be to join these two queries into one, but then it only fixes it for the manage Boxes view.

I do not believe this should be part of this ticket.

front/src/views/NotFoundView/NotFoundView.tsx
```diff
@@ -1,9 +1,20 @@
...
-  return <Navigate to="/" replace state={{ origin: location.pathname }} />;
+  const { triggerError } = useErrorHandling();
+
+  if (!user || !user["https://www.boxtribute.com/base_ids"]) {
+    triggerError({ message: "This requested page does not exist." });
+    return <div>Not Found</div>;
+  }
+
+  // If the requested route was not found redirect to statviz
+  const baseId = user["https://www.boxtribute.com/base_ids"][0];
```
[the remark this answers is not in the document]
HaGuesto: I added a commit for this, but in most cases it will not make a difference. If you do a Browser reload (only way to test this atm through the UI all states are flushed and it does not remember the selected base from before.

@fhenrich33 will jotai change this?

front/src/App.tsx
```diff
@@ -75,7 +77,8 @@ function DropappRedirect({ path }: DropappRedirectProps) {
   if (!user || !user["https://www.boxtribute.com/base_ids"])
     return <Navigate to={pathToRedirect} replace />;
 
-  const baseId = user["https://www.boxtribute.com/base_ids"][0];
+  const baseId =
+    globalPreferences.selectedBase?.id || user["https://www.boxtribute.com/base_ids"][0];
```
pylipp, line 81: If line 77 is true, this line would never be reached, right?

front/src/views/NotFoundView/NotFoundView.tsx
```diff
@@ -18,7 +21,9 @@ function NotFoundView() {
   }
 
   // If the requested route was not found redirect to statviz
-  const baseId = user["https://www.boxtribute.com/base_ids"][0];
+
+  const baseId =
+    globalPreferences.selectedBase?.id || user["https://www.boxtribute.com/base_ids"][0];
```
pylipp, line 26: maybe using an entry from `src/utils/constants.ts` would be nice here and in App.tsx

pylipp:
> @pylipp The reason why it throws two error toasts is that the number of error toasts depend on the failing graphql requests. When you request `/bases/5/boxes` than two graphQL requests are sent to the backend - one to query the boxes and one to query the options for the action buttons. Fixing it would maybe be to join these two queries into one, but then it only fixes it for the manage Boxes view.
> 
> I do not believe this should be part of this ticket.

ok no worries then, it's expected behaviour kind of. Defs not part of this ticket but not really a problem in general I suppose

pylipp:
> Will this comment be addressed in this PR? https://trello.com/c/Dk3IRAva#comment-6746d22bbe73f4ae229adc8c

Seconding the comment about e.g. label@athens.org not being able to log in (Error toast: "Access to this page is not permitted for your user group."; redirected to /error right away).

front/src/App.tsx
fhenrich33: For this to be reached, the `if` statement in line 77 should be negative, so this makes sense. @pylipp

front/src/views/NotFoundView/NotFoundView.tsx
HaGuesto: it is used two times in the code. not yet ready for abstraction in my view since we are touching and refactoring globalPreferences probably soon again.

HaGuesto:
not sure, but I do not think there is anything left to do for me, right?

pylipp:
> not sure, but I do not think there is anything left to do for me, right?

@HaGuesto maybe label creators can be shown an empty page? At least they can see the menu then (Aid Inventory -> Print Labels)

HaGuesto:
> > not sure, but I do not think there is anything left to do for me, right?
> 
> @HaGuesto maybe label creators can be shown an empty page? At least they can see the menu then (Aid Inventory -> Print Labels)

mmh. wouldn't do that now. 
I'd have to define a second error view that shows the menu and only is shown if there is a authenticated user, that has access to the base, but not enough permissions for statviz. That is quite a bit of work on checking the right permissions only for the edge case of a label creator. Not even sure if we have one real user of this kind.

pylipp:
I see @HaGuesto, let's consider it out of scope. I'm merging the PR then and noting to myself to check about active label creation users and maybe log the issue in trello.

pylipp (approved):

PR #1855, opened 2024-12-12
Fix/statviz beneficiary text

HaGuesto:
if all DOBs and genders are specified:
image (4)

if not:
image (5)

### DIFF
--- a/shared-components/statviz/components/visualizations/demographic/DemographicPyramid.tsx
+++ b/shared-components/statviz/components/visualizations/demographic/DemographicPyramid.tsx
@@ -50,14 +50,14 @@ export default function DemographicPyramid({
       demographics?.facts as BeneficiaryDemographicsResult[],
       filter((value) => value.gender === "Male" && value.age !== null),
       groupBy("age", [summarize({ count: sum("count") })]),
-      map((value) => ({ x: value.count, y: value.age ?? 0 })),
+      map((value) => ({ x: value.count, y: value.age! })),
     );
 
     const dataXl = tidy(
       demographics?.facts as BeneficiaryDemographicsResult[],
       filter((value) => value.gender === "Female" && value.age !== null),
       groupBy("age", [summarize({ count: sum("count") })]),
-      map((value) => ({ x: value.count, y: value.age ?? 0 })),
+      map((value) => ({ x: value.count, y: value.age! })),
     );
 
     return [dataXr, dataXl];
@@ -77,7 +77,7 @@ export default function DemographicPyramid({
 
     const femaleCount = tidy(
       demographics?.facts as BeneficiaryDemographicsResult[],
-      filter((value) => value.gender === "Male"),
+      filter((value) => value.gender === "Female"),
       summarize({ total: sum("count") }),
     )[0].total;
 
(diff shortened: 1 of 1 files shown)


HaGuesto:
https://trello.com/c/XUrjWcSI

shared-components/statviz/components/visualizations/demographic/DemographicPyramid.tsx
```diff
@@ -36,7 +36,7 @@ interface IDemographicChartProps {
   height: number;
 }
 
-const heading = "Beneficiaries Registered";
+const heading = "Beneficiaries Registrations";
```
aerinsol, line 39: Beneficiary Registrations

Should be the title

shared-components/statviz/components/visualizations/demographic/DemographicPyramid.tsx
```diff
@@ -45,28 +45,99 @@ export default function DemographicPyramid({
...
+    useMemo(prepareFactsForText, [demographics?.facts]);
+
+  const beneficiariesRegistrationsText = (
+    <Text>
+      There were <chakra.span as="b">{totalCount}</chakra.span> beneficiaries registered in the
+      selected time period, <chakra.span as="b">{maleCount}</chakra.span> were male and{" "}
+      <chakra.span as="b">{femaleCount}</chakra.span> were female.
+      {ageNullOrDiverseCount ? (
+        <chakra.span>
+          <br />
```
aerinsol, line 117: Needs an additional `<br />` or if we have the the right CSS paragraph formatting, wrap the first line with `<p> </p>`, you tell me which it is

aerinsol (changes requested):
Line spacing and wording changes needed as mentioned in the Trello def

shared-components/statviz/components/visualizations/demographic/DemographicPyramid.tsx
```diff
@@ -45,28 +45,100 @@ export default function DemographicPyramid({
...
   const onExport = getOnExport(BarChartCenterAxis);
 
-  const prepareFacts = () => {
+  const prepareFactsForGraph = () => {
     const dataXr = tidy(
       demographics?.facts as BeneficiaryDemographicsResult[],
-      filter((value) => value.gender === "Male"),
+      filter((value) => value.gender === "Male" && value.age !== null),
       groupBy("age", [summarize({ count: sum("count") })]),
       map((value) => ({ x: value.count, y: value.age ?? 0 })),
```
pylipp, line 53: I suppose the `?? 0` can be removed since null age is filtered out

shared-components/statviz/components/visualizations/demographic/DemographicPyramid.tsx
```diff
@@ -45,28 +45,100 @@ export default function DemographicPyramid({
...
+
+    const maleCount = tidy(
+      demographics?.facts as BeneficiaryDemographicsResult[],
+      filter((value) => value.gender === "Male"),
+      summarize({ total: sum("count") }),
+    )[0].total;
+
+    const femaleCount = tidy(
+      demographics?.facts as BeneficiaryDemographicsResult[],
+      filter((value) => value.gender === "Male"),
```
pylipp, line 80: has to be "Female" here

pylipp:
@aerinsol if the scenario is like the following, would you like the line with "0 people..." to be hidden?
```
...
10 of these beneficiaries are not shown in the graph:
* 10 people are missing a date of birth.
* 0 people have an unknown gender.
```


Also mind that, if there are people without DoB _and_ with unknown gender, the counts of the two last lines won't add up to the line before, e.g.
```
...
10 of these beneficiaries are not shown in the graph:
* 10 people are missing a date of birth.
* 5 people have an unknown gender.
```
Would it be better to swap the first line to "Some beneficiaries are not shown in the graph:" ?

pylipp:
**Functional review passes** when adjusting the female-count filter. thx @HaGuesto :rocket:

aerinsol:
> @aerinsol if the scenario is like the following, would you like the line with "0 people..." to be hidden?
> ```

Yes, you're right that would be better. 

> 
> Also mind that, if there are people without DoB _and_ with unknown gender, the counts of the two last lines won't add up to the line before, e.g.
> ```
> ...
> 10 of these beneficiaries are not shown in the graph:
> * 10 people are missing a date of birth.
> * 5 people have an unknown gender.
> ```
> Would it be better to swap the first line to "Some beneficiaries are not shown in the graph:" ?

No, let's leave it, it is actually more clear there is an overlap between the two types of missing data that way.

shared-components/statviz/components/visualizations/demographic/DemographicPyramid.tsx
HaGuesto: great catch

shared-components/statviz/components/visualizations/demographic/DemographicPyramid.tsx
HaGuesto: seems that it is still needed for typecasting. value.age according to its type could also be undefined. So, we need the ?? still that the type changes from number | undefined to number

HaGuesto: I replaced now with the ! operator.

PR #1890, opened 2025-01-03
Back-end for assign-tags mutations

pylipp:
https://trello.com/c/zyiOzn2C/1647-20-be-implement-assigntagstoboxes-and-unassigntagsfromboxes-mutations

New mutations:
- assignTagsToBoxes
- unassignTagsFromBoxes

Deleted mutations:
- assignTagToBoxes
- unassignTagFromBoxes

### Usage
FYI for @jlin95 as heads-up for the FE implementation. I'll try to cover possible success and (partial) error scenarios:
1. all given boxes and tags are valid. The `BoxesTagsOperationResult.updatedBoxes` contains the updated boxes, the other fields are empty
2. all tags are valid but some boxes are invalid (e.g. they already have all of the requested tags assigned). In addition to (1.) the `invalidBoxLabelIdentifiers` field contains a list of box labels that were not updated
3. all boxes are valid but some tags are invalid (e.g. they're the wrong type, or they're soft-deleted; basically scenarios that should not happen if tags are correctly prepared for selection in the FE): info about errors will be listed in `tagErrorInfo`. Valid tags will still be assigned to valid boxes, and the `updatedBoxes` is filled
4. some boxes are invalid and some tags are invalid: mixture of (2.) and (3.) applies
5. all tags are invalid (e.g. when they are from different bases; again this should actually not happen in the FE): `updatedBoxes` and `invalidBoxLabelIdentifiers` are empty, `tagErrorInfo` contains `TagBaseMismatchError` for every tag
6. all boxes are invalid: `updatedBoxes` is empty

For example queries using the new mutations, please see the BE tests, e.g. [here](https://github.com/boxwise/boxtribute/blob/backend-for-assign-tags-mutations/back/test/endpoint_tests/test_box.py#L706)


back/boxtribute_server/business_logic/warehouse/box/sql.py
```diff
@@ -0,0 +1,55 @@
...
+    SELECT
+        box_ids.id as box_id,
+        tag_ids.id AS tag_id
+    FROM box_ids CROSS JOIN tag_ids
+),
+existing_relations AS (
+    -- Find existing box-to-tag assignments
+    SELECT object_id AS box_id, tag_id
+    FROM tags_relations
+    WHERE object_type = "Stock" AND deleted_on IS NULL
```
HaGuesto, line 30: AND (boxid IN box_ids OR tag_id IN tag_ids)

back/boxtribute_server/business_logic/warehouse/box/mutations.py
```diff
@@ -156,78 +164,119 @@ def resolve_move_boxes_to_location(*_, update_input):
...
+    if not valid_tag_ids:
+        return BoxesTagsOperationResult(
+            updated_boxes=[],
+            invalid_box_label_identifiers=[],
+            tag_error_info=tag_errors,
         )
-        .order_by(Box.id)
+
+    label_identifiers = set(update_input["label_identifiers"])
+    valid_boxes = execute_sql(
```
HaGuesto, line 250: somewhere here should be a check if the base of the boxes are the same as the base of the tags

back/boxtribute_server/business_logic/warehouse/box/mutations.py
```diff
@@ -156,78 +164,119 @@ def resolve_move_boxes_to_location(*_, update_input):
...
+        label_identifiers,
+        g.user.authorized_base_ids("stock:write"),
```
HaGuesto, line 252: possibly replace by base_id that all boxes and tags are in

HaGuesto (approved):
## code review

A few comments left in the code. 
Please leave assignTag and unassignTag in there.

I do not have to code review again, but will do a functional review async.

back/boxtribute_server/business_logic/warehouse/box/sql.py
pylipp: `AND (boxid IN box_ids AND tag_id IN tag_ids)` is correct because we don't want to pull in any non-requested boxes with any of the tags in `tag_ids` assigned.

back/boxtribute_server/business_logic/warehouse/box/mutations.py
pylipp: I can't replace it by the base that all boxes are in, because this _is_ the query to determine all valid boxes.

Solved it by passing the common base of all tags (guaranteed to be a single one due to prior validation), which means that any boxes in other bases are marked as invalid.

back/boxtribute_server/business_logic/warehouse/box/mutations.py
pylipp: Omitting this check in favor of filtering out any boxes in bases different from the base common to all tags, see below.

PR #1917, opened 2025-01-21
Handle undefined value in SelectColumnFilter

pylipp:
E.g. if a box has no size (but measure value + unit instead)


fhenrich33 (approved):
## Functional and Code Review Passed

PR #1935, opened 2025-01-28
GraphQL implementation update for consistency

pylipp:
- Use simpler date parsing
- Update comments
- Use only one utcnow() call when creating agreement
- Use unit loader instead of relying on join


pylipp:
[2.0: GraphQL implementation update for consistency](https://trello.com/c/B3ocgGGH/1665-20-graphql-implementation-update-for-consistency)

PR #1948, opened 2025-02-03
Avoid creating agreement without active partner bases

pylipp:


pylipp:
[2.0 BUG: Avoid creating agreement without active partner bases (BE)](https://trello.com/c/yHBosqm0/1667-20-avoid-creating-agreement-without-active-partner-bases)

PR #1973, opened 2025-02-13
Add Auth0-Slack bridge

pylipp:
Open questions:
- new app/service or re-using e.g. api-staging? Would need its own deployment step
- only for production logs or also staging?

Remember to update CI


pylipp:
[Auth0 Error Log Stream #large](https://trello.com/c/6n63oRB8/1668-auth0-error-log-stream-large)

HaGuesto (approved):
# code review passed

back/boxtribute_server/routes.py
```diff
@@ -150,3 +151,18 @@ def cron(job_name):
         return jsonify({"message": f"cleaned up {nr_addresses} email addresses"}), 200
 
     return jsonify({"message": f"unknown job '{job_name}'"}), 400
+
+
+@api_bp.post("/auth0-slack-bridge")
```
HaGuesto, line 156: I would actually not put it on a separate service and rather put it on the app_blueprint.

Reason: the logs should only be sent if there are certain errors logged in Auth0 when a user interacts with the app. This means that the app service is already running. The api servicce might have to start and we would have extra cost incurring.

pylipp: that makes sense on start-up time and extra cost :+1: 

I couldn't quite access how many log requests would be streamed to the bridge and was worried that it would negatively affect the webapp BE performance.

HaGuesto: we should be able to filter which errors are streamed.

PR #1986, opened 2025-02-20
Soft-delete organisation when removing last active base

pylipp:
- Soft-delete organisation if last active base deleted
- Print initial information about potentially removing organisation

### DIFF
--- a/back/boxtribute_server/cli/remove_base_access.py
+++ b/back/boxtribute_server/cli/remove_base_access.py
@@ -184,19 +184,20 @@ def _update_user_data_in_database(
     # Soft-delete governing organisation if no active bases left
     db.database.execute_sql(
         """\
-UPDATE organisations
-SET deleted = UTC_TIMESTAMP()
-WHERE id = (
-    SELECT id FROM (
-        SELECT o.id,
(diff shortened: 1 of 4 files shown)


pylipp:
https://trello.com/c/dnI45xkJ

back/boxtribute_server/cli/remove_base_access.py
```diff
@@ -168,6 +181,26 @@ def _update_user_data_in_database(
         (int(base_id),),
     )
 
+    # Soft-delete governing organisation if no active bases left
+    db.database.execute_sql(
+        """\
+UPDATE organisations
```
HaGuesto, line 187: UPDATE organisations o
LEFT JOIN camps c ON c.organisation_id = o.id
SET o.deleted = UTC_TIMESTAMP()
WHERE o.id = (
    SELECT organisation_id 
    FROM camps 
    WHERE id = %s
)
AND (
    SELECT COUNT(*) 
    FROM camps c2 
    WHERE c2.organisation_id = o.id 
    AND (c2.deleted IS NULL OR NOT c2.deleted)
) = 0;

HaGuesto (approved):
# functional review delayed to staging

PR #1943, opened 2025-01-31
Manage Products Feature

fhenrich33:
image

Screen Shot 2025-02-06 at 09 02 07

PR #2016, opened 2025-03-07
Fix/disable standard products

HaGuesto:
follow up to fix disable standard products in #1943


HaGuesto:
[Disable standard Product should work even if there boxes associated with this product. There should only be no InStock boxes](https://trello.com/c/3Tzm8fyN/1697-disable-standard-product-should-work-even-if-there-boxes-associated-with-this-product-there-should-only-be-no-instock-boxes)

HaGuesto:
[V2 FE: add table actions to disable ASSORT standard products #large](https://trello.com/c/HOVWOPd8/1656-v2-fe-add-table-actions-to-disable-assort-standard-products-large)

pylipp:
nice clean-up work @HaGuesto :rocket:

PR #2022, opened 2025-03-10
Only take certain box states into account in check of still being assigned to product

pylipp:
https://trello.com/c/3Tzm8fyN


HaGuesto (commented):
## code review passed @pylipp 

I assume you create a follow-up ticket to split this up between inStockItems and inTransit Items, right?

PR #2024, opened 2025-03-11
Handle god user without any base_ids in their auth token

pylipp:
https://trello.com/c/NkwbKoIt


HaGuesto (approved):
yes, makes sense. I handled god users further down similarly.

# code review passed

PR #2027, opened 2025-03-12
Make base-name skeletons load for god user

pylipp:
God user does not have base_ids set in their JWT, hence selectedBaseIdAtom keeps returning "0" which does not match the urlBaseId.

As a result, the skeletons that use selectedBaseAtom load forever (e.g. in EnableStandardProductForm).


pylipp:
https://trello.com/c/NkwbKoIt

PR #2040, opened 2025-03-21
Add Product.transferItemsCount field and update alert when trying to disable product with items in use

pylipp:
- Add Product.transferItemsCount field
- Update disable-product message
- Update value displayed in "Items in Use" column

Not totally sure about the `accessor` field in the column definition in ProductsView.
needs update after #2045 is merged


HaGuesto (approved):
## code review passed
@pylipp almost Full Stack already.

## functional review passed
I added an improvement to the warning message to show the part on shipments only if there are boxes in a shipment.

front/src/views/Products/ProductsView.tsx
```diff
@@ -304,6 +318,9 @@ function Products() {
         accessor: "instockItemsCount",
         id: "instockItemsCount",
         disableFilters: true,
+        Cell: ({ row }: CellProps<ProductRow, any>) => (
+          <>{(row.original.instockItemsCount || 0) + (row.original.transferItemsCount || 0)}</>
```
HaGuesto, line 322: @pylipp I'd actually rather do this in the data transformation. will fix it in the table view PR

PR #2062, opened 2025-03-29
Increase timeout for tests in CI

fhenrich33:


fhenrich33:
@pylipp I couldn't fix 3.4.8.4, but at least I made sure that only this test fails in CI consistently. I'm working on the Playwright PR, so sooner rather than later, this won't be an issue anymore.

pylipp (approved):

pylipp:
https://trello.com/c/eEqmJJpm

PR #2063, opened 2025-03-30
BUG FIX for unassigntags - No tags should get rendered if tagless box is selected

jlin95:
As the title describes. Currently if 2 boxes are selected and one of them has tags but the other has 0, when the box with tags gets unchecked, leaving only one box selected, the dropdown doesnt render 0 tags (which is expected behavior). 
This commit fixes that.

### DIFF
--- a/front/src/views/Boxes/components/RemoveTagsButton.tsx
+++ b/front/src/views/Boxes/components/RemoveTagsButton.tsx
@@ -28,8 +28,10 @@ const RemoveTagsButton: React.FC<RemoveTagsButtonProps> = ({
   const [selectedTagOptions, setSelectedTagOptions] = useState<IDropdownOption[]>([]);
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


jlin95:
tagging @aerinsol for functional review input!

jlin95:
[2.0 BUG: Unassign Tags does not correctly update the Manage Boxes table or unassign tags for a single tag](https://trello.com/c/bes6jD5t/1719-20-bug-unassign-tags-does-not-correctly-update-the-manage-boxes-table-or-unassign-tags-for-a-single-tag)

aerinsol:
Good inference @jlin95 , thanks for the quick response. Still needed, but I think we first need to tackle expected UI/UX flow for removing tags (as discussed in Trello). Maybe on our next face to face?

pylipp (approved):

PR #2074, opened 2025-04-01
Fix/shipmentview when receiving

HaGuesto:


HaGuesto:
[FE: adapt reconciliation View product and size is already reconciled due to both sides using ASSORT products # medium](https://trello.com/c/mBw7NjTW/1689-fe-adapt-reconciliation-view-product-and-size-is-already-reconciled-due-to-both-sides-using-assort-products-medium)

PR #2075, opened 2025-04-01
Cached products take precedence over matched products

fhenrich33:
firefox_PBdZzO05cD

### DIFF
--- a/front/src/components/BoxReconciliationOverlay/components/BoxReconciliationAccordion.tsx
+++ b/front/src/components/BoxReconciliationOverlay/components/BoxReconciliationAccordion.tsx
@@ -17,7 +17,6 @@ import { MatchProductsFormData, MatchProductsForm } from "./MatchProductsForm";
 import { IReceiveLocationFormData, ReceiveLocationForm } from "./ReceiveLocationForm";
 import { useAtom, useAtomValue, useSetAtom } from "jotai";
 import {
-  hasReconciliationMatchProductAtomCached,
   reconciliationMatchProductAtom,
   reconciliationReceiveLocationAtom,
 } from "stores/globalCacheStore";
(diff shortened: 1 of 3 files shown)


front/src/components/BoxReconciliationOverlay/components/BoxReconciliationAccordion.tsx
```diff
@@ -55,8 +56,14 @@ export function BoxReconcilationAccordion({
   onBoxUndelivered,
   onBoxDelivered,
 }: IBoxReconcilationAccordionProps) {
+  const [reconciliationMatchProductCache, setReconciliationMatchProductCache] = useAtom(
+    reconciliationMatchProductAtom,
+  );
+  const hasProductCache = useAtomValue(hasReconciliationMatchProductAtomCached);
```
HaGuesto, line 62: @fhenrich33 I would replace the new read-only atom with what you do one level below:

```
const isProductIdMatchedInCache = !!cachedReconciliationMatchProduct[matchingProductSourceId];
```

front/src/components/BoxReconciliationOverlay/components/BoxReconciliationAccordion.tsx
```diff
@@ -55,8 +56,14 @@ export function BoxReconcilationAccordion({
...
   const isProductAutoMatched = !!shipmentDetail?.autoMatchingTargetProduct;
-  const [accordionIndex, setAccordionIndex] = useState(isProductAutoMatched ? 1 : 0);
+  const [accordionIndex, setAccordionIndex] = useState(
+    hasProductCache ? 0 : isProductAutoMatched ? 1 : 0,
```
HaGuesto, line 65: and then replace the value here. Otherwise, the accordion always opens if just any product matching was cached.

HaGuesto (changes requested):

front/src/components/BoxReconciliationOverlay/components/BoxReconciliationAccordion.tsx
fhenrich33: Whoops forgot about matching product ids, nice catch!

pylipp:
@fhenrich33 how do you currently functional-test this? I'd like to see the expected behavior :sweat_smile:

fhenrich33:
> @fhenrich33 how do you currently functional-test this? I'd like to see the expected behavior 😅

Uploaded a GIF to show the behaviour cc @aerinsol

HaGuesto (approved):

PR #2076, opened 2025-04-01
Add baseId parameter to products query

pylipp:
- Revert "filter products by baseId"
- Add optional baseId input for products query
- Query only products of current base in ProductsView


HaGuesto (approved):
# code passed

# functional review passed
@pylipp I did a fix since you did not add the baseId variable to the refetchQueries

PR #2079, opened 2025-04-02
Fix default valid from using today's date in Transfer Agreement

fhenrich33:


pylipp:
@fhenrich33 was this a bug with the validFrom input in the CreateTransferAgreement form? Can you link a Trello card?

fhenrich33:
> @fhenrich33 was this a bug with the validFrom input in the CreateTransferAgreement form? Can you link a Trello card?

@pylipp This was a bug introduced by me by mistake here: https://github.com/boxwise/boxtribute/pull/2078

Because there were some type mismatches I tried to resolve, but for this specific instance, the Zod transform utility didn't work.

See this as a small rollback.

PR #2084, opened 2025-04-04
Fix/update unassign tags behavior

jlin95:
This PR fixes the unassign tags implementation by unassigning tags that were removed from the dropdown.

### DIFF
--- a/front/src/views/Boxes/components/BoxesActionsAndTable.tsx
+++ b/front/src/views/Boxes/components/BoxesActionsAndTable.tsx
@@ -248,6 +248,7 @@ function BoxesActionsAndTable({
         selectedBoxes.map((box) => box.values.labelIdentifier),
         tagIds.map((id) => parseInt(id, 10)),
       );
+      setAutoResetSelectedRows(true);
     },
     [assignTags, selectedBoxes],
   );
@@ -255,14 +256,15 @@ function BoxesActionsAndTable({
   // Unassign tags from boxes
   const { unassignTags, isLoading: isUnassignTagsLoading } = useUnassignTags();
   const onUnassignTags = useCallback(
-    async (tagIds: string[]) => {
+    (tagIds: string[]) => {
       if (tagIds.length > 0) {
         setAutoResetSelectedRows(false);
-
-        await unassignTags(
+        unassignTags(
           selectedBoxes.map((box) => box.values.labelIdentifier),
           tagIds.map((id) => parseInt(id, 10)),
-        );
+        ).then(() => {
+          setAutoResetSelectedRows(true);
+        });
       }
     },
     [unassignTags, selectedBoxes],
--- a/front/src/views/Boxes/components/BoxesActionsAndTable.tsx
+++ b/front/src/views/Boxes/components/BoxesActionsAndTable.tsx
@@ -273,8 +273,8 @@ function BoxesActionsAndTable({
     isAssignBoxesToShipmentLoading ||
     isUnassignBoxesFromShipmentsLoading ||
     isDeleteBoxesLoading ||
-    isAssignTagsLoading;
-  isUnassignTagsLoading;
+    isAssignTagsLoading ||
+    isUnassignTagsLoading;
 
   const actionButtons = useMemo(
     () => [
--- a/front/src/views/Boxes/components/BoxesActionsAndTable.tsx
+++ b/front/src/views/Boxes/components/BoxesActionsAndTable.tsx
@@ -52,6 +52,7 @@ function BoxesActionsAndTable({
   const baseId = useAtomValue(selectedBaseIdAtom);
 
   const { createToast } = useNotification();
+  const [autoResetSelectedRows, setAutoResetSelectedRows] = useState(true);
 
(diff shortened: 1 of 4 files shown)


aerinsol:
Asking Philipp to review first since I can't test until it's merged on master

front/src/views/Boxes/components/BoxesActionsAndTable.tsx
```diff
@@ -251,24 +251,24 @@ function BoxesActionsAndTable({
...
-  //       selectedBoxes.map((box) => box.values.labelIdentifier),
-  //       tagIds.map((id) => parseInt(id, 10)),
-  //     );
-  //   },
-  //   [unassignTags, selectedBoxes],
-  // );
+  const { unassignTags, isLoading: isUnassignTagsLoading } = useUnassignTags();
+  const onUnassignTags = useCallback(
+    async (tagIds: string[]) => {
+      await unassignTags(
```
pylipp, line 257: Can you add an early-exit here or better into the underlying hook in the case of `tagIds` being empty. In this case we can avoid sending a request to the BE that has no effect anyways.

pylipp:
[2.0 BUG: Unassign Tags User Feedback and Behavior is Incorrect](https://trello.com/c/nQpZW1gB/1718-20-bug-unassign-tags-user-feedback-and-behavior-is-incorrect)

pylipp (changes requested):
### Functional review
- tag-removal logic is acc. to the GMail-style
- dropdown only shows tags of the selected boxes

However, please check the expected behavior requested in the Trello card.
1. boxes should stay checked after tag-removal (@aerinsol can you specify this should _always_ happen after tag-removal, or only in the case of partial removal of tags (as in the example on the card))
2. no warning message if one of the selected boxes does not have any of the removed tags assigned (sorry about the back and forth; I gave you different directions in an earlier PR :face_with_diagonal_mouth: )

aerinsol:
> 1. boxes should stay checked after tag-removal (@aerinsol can you specify this should _always_ happen after tag-removal, or only in the case of partial removal of tags (as in the example on the card))

**Yes, boxes should always remain checked after tag removal.** This allows them to queue up multiple actions if needed. Users can always mass select/unselect using the table interaction.

jlin95:
Is this only for tags removal? Because for tags assigning we refresh the
cache and reload. Should it not be aligned for both action types?

pylipp:
> Is this only for tags removal? Because for tags assigning we refresh the cache and reload. Should it not be aligned for both action types?

@jlin95 you're right, it should be aligned. If it's simple and quick to have the behavior that boxes stay selected both after adding **and** removing tags, please add it to this PR. If it takes more than 15mins, I'd consider it feature creep and it should be handled in another PR (could you please create a Trello card then?).
FYI @aerinsol

jlin95:
I'd recommend having the improvement for another card.

aerinsol:
Thank you @pylipp and @jlin95 , perfect resolution, I would have said exactly the same thing.

pylipp:
> Does this block release of unassigntags feature?

Let's discuss at SU today.

> Also since we are keeping the checked state on assign/unassign tags, do we want to do it for other actions too?

I think yes, but it can be part of the discussion for the new card (we'll talk during SU about the due date).

> no warning message if one of the selected boxes does not have any of the removed tags assigned (sorry about the back and forth; I gave you different directions in an earlier PR 🫤 )

This is the only part missing in the current PR now.

FYI @aerinsol

jlin95:
Please check the screen recording for the latest changes. 

cc @HaGuesto @aerinsol @pylipp 

Uploading Screen Recording 2025-04-17 at 01.37.33.mov…

pylipp (approved):
@jlin95 thank you very much for the implementation and the screen capture, makes it very clear. Pass from my side!

aerinsol:
My only comment is that if we are putting a toast for when tags are assigned, there should be a corollary one for when tags are unassigned @jlin95. Would be a nice to have to fix, but not a blocker for release. 

Otherwise LGTM!

aerinsol (changes requested):
Small change requested to have matching toast for tags that have been unassigned.

jlin95:
Hi hi, please check this screen recording for the toast message showing for unassign tags. 
@aerinsol @pylipp once again, thank you!

https://github.com/user-attachments/assets/d383fa2d-65b4-4ebe-a02e-63d60e3ed579

aerinsol (approved):
LGTM!

front/src/views/Boxes/components/BoxesActionsAndTable.tsx
```diff
@@ -248,21 +248,23 @@ function BoxesActionsAndTable({
...
         setAutoResetSelectedRows(false);
-
-        await unassignTags(
+        unassignTags(
           selectedBoxes.map((box) => box.values.labelIdentifier),
           tagIds.map((id) => parseInt(id, 10)),
-        );
+        ).then(() => {
+          setAutoResetSelectedRows(true);
+        });
```
jlin95, line 267: Nice

PR #2109, opened 2025-04-17
Update staging db seed

pylipp:
- Update cms_functions and related reference tables
- Add services and services_relations tables
- Update phinxlog
- Fix sizegroup for Male Bodywarmers standard product
- Add customfield columns to people table
- Add additional_field columns to camps table
- Use correct sizegroup for children standard products
- Update QR codes


pylipp:
[2.0 BUG: Size Range for Boys Shortsleeves is incorrect and does not match the Unified ASSORT list](https://trello.com/c/NZbfuGfA/1728-20-bug-size-range-for-boys-shortsleeves-is-incorrect-and-does-not-match-the-unified-assort-list)

PR #2111, opened 2025-04-19
Feat/tags filter manage boxes

jlin95:
Test scenarios described in the [ticket](https://trello.com/c/RcFYvV5S/1606-20-fe-manage-boxes-add-a-filter-for-tags-large) from 1 to 9 were tested. 

Note: This change introduces some more conditional checks into filtering and rendering option logic of `SelectColumnFilterUI` to cater to the tags use case. Let me know how you feel about this. 

Here's the screen recording which should cover all the 9 test scenarios. 

https://github.com/user-attachments/assets/6d31251f-a8e0-4bab-a70c-e221436f75af


front/src/components/Table/Filter.tsx
```diff
@@ -57,6 +57,9 @@ export function SelectColumnFilterUI({
             value={
               filterValue &&
               filterValue.map((value) => {
+                if (value.__typename === "Tag" && typeof value === "object" && value !== null) {
```
jlin95, line 60: only specific for tag because it's not just a string value

front/src/components/Table/Filter.tsx
```diff
@@ -97,16 +100,25 @@ export function SelectColumnFilter({
...
       const value = row.values[id];
-      // if the data passed to the table is more complex than a string we need to pass the data as value
-      if (typeof value === "object" && value !== null) {
+      // if the data passed to the table is more complex than a string we need to pass the data as value.
+      // This excludes the case for tags filtering where we introduce a second condition in the second if hook.
+      if (typeof value === "object" && value !== null && id.toLowerCase() !== "tags") {
         const objectToString = ObjectToString(value);
         groupedOptionLabels.add(objectToString);
         optionValues[objectToString] = value;
+      } else if (typeof value === "object" && value !== null && id.toLowerCase() === "tags") {
```
jlin95, line 109: Again very specific for tags use case

jlin95:
[2.0 FE Manage Boxes: add a filter for tags #large](https://trello.com/c/RcFYvV5S/1606-20-fe-manage-boxes-add-a-filter-for-tags-large)

aerinsol (approved):
Design review based on the vid alone LGTM! @pylipp @HaGuesto or @fhenrich33 can do a code review.

HaGuesto (approved):
## code review

it feels a bit hacky but I'm happy with it. Let's abstract more if we need to build more filter functions.

jlin95:
> ## code review
> it feels a bit hacky but I'm happy with it. Let's abstract more if we need to build more filter functions.

@HaGuesto my thoughts too. The generic SelectFilterUI in this use case did not work well with the nested structure of the tag data. I'd see if it's worth putting in a small refctor (since there's still like 4 hours of dev time left allocated to the budget for this task) wdyt?

PR #2119, opened 2025-04-22
Avoid confusing peewee with "registered" input

pylipp:
peewee then assumes there's a "registered" column if the first element
of the input array contains this key


pylipp:
[2.0: mutation to bulk-create beneficiaries (without error handling)](https://trello.com/c/ihefSwk0/1731-20-mutation-to-bulk-create-beneficiaries-without-error-handling)

PR #2104, opened 2025-04-16
Public Statviz with Stock Overview page

fhenrich33:
firefox_ofBnvyv6pI

PR #2120, opened 2025-04-22
Public Statviz infra configuration

fhenrich33:
Ignore everything non-yml/yaml until https://github.com/boxwise/boxtribute/pull/2104 is merged.


.circleci/config.yml
```diff
@@ -552,6 +609,16 @@ workflows:
             branches:
               only:
                 - master
+      - deploy-statviz:
```
pylipp, line 612: there's already `deploy-shared` below which should be used (update `requires` field to `build-statviz-staging`)

.circleci/config.yml
```diff
@@ -648,6 +725,17 @@ workflows:
               only: /^v.*/
             branches:
               ignore: /.*/
+      - deploy-statviz:
```
pylipp, line 728: needs to be called deploy-shared

.circleci/config.yml
```diff
@@ -648,6 +725,17 @@ workflows:
...
+          name: deploy-statviz-production
+          context: PRODUCTION
+          serviceName: v2-production
```
pylipp, line 731: should be shared-production

PR #2133, opened 2025-04-28
Update README and comments about beta-level

pylipp:
@aerinsol please LMK if the beta-levels are clearer now. I'm a bit reluctant to introduce a Google Sheet because I don't want to end in the same situation as with permissions (source of truth not being clear; code and sheet diverging)


aerinsol:
@pylipp yes it's much clearer, and I prefer this method as well. Since the default level is 3, what are the permissions / access you put on level 0, 1, and 2? 

Also curious about beta levels 5 and 6

pylipp:
> @pylipp yes I did see that. What I didn't understand was the reason for structuring the permissions by function in this way - is it meant to mimic our standard user permissions in one way or another?

@aerinsol the structure has historically developed this way. In the beginning it was about ManageBoxes that we wanted to enable for certain users. Later the transfer features that some should test before it was enabled for all.
Sure, the distinction of the beta levels smaller than the default is not relevant anymore but I've kept it.

aerinsol:
> @aerinsol the structure has historically developed this way. In the beginning it was about ManageBoxes that we wanted to enable for certain users. Later the transfer features that some should test before it was enabled for all. Sure, the distinction of the beta levels smaller than the default is not relevant anymore but I've kept it.

@pylipp   I always assumed that we migrated features "down" in beta levels as we progressed, but it sounds like rather than the default beta level increases with every feature we release first to a smaller group. I don't have any objections with that method as long as we update the documentation on the default beta level every time we do a wide release of a previously beta'd feature.

PR #2152, opened 2025-05-12
Update translate items in DB seed

pylipp:


pylipp:
[dropapp: Rename "Settings" in the User Dropdown with "Account" to make it more consistent with V2](https://trello.com/c/t7GeU3Is/1724-dropapp-rename-settings-in-the-user-dropdown-with-account-to-make-it-more-consistent-with-v2)

PR #2155, opened 2025-05-13
Create-box from ManageBoxes

pylipp:
firefox_mH8VKOtCnw

Mobile viewport:
2025-05-19-102753_371x664_scrot

Mobile viewport BEFORE this PR:
2025-05-19-101751_374x667_scrot

### DIFF
--- a/front/src/views/BoxCreate/BoxCreateView.tsx
+++ b/front/src/views/BoxCreate/BoxCreateView.tsx
@@ -92,10 +92,23 @@ function BoxCreateView() {
   // variables in URL
   const qrCode = useParams<{ qrCode: string }>().qrCode!;
 
-  // Query the QR-Code
-  const qrCodeExists = useQuery(CHECK_IF_QR_EXISTS_IN_DB, {
-    variables: { qrCode },
-  });
+  if (qrCode) {
+    // Query the QR-Code
+    const qrCodeExists = useQuery(CHECK_IF_QR_EXISTS_IN_DB, {
+      variables: { qrCode },
+    });
+    // Check the QR Code
+    useEffect(() => {
+      if (qrCodeExists.data?.qrExists === false) {
+        createToast({
+          title: "Error",
+          type: "error",
+          message: "The QR-Code is not from Boxtribute!",
+        });
+      }
+      // TODO: Add check if Qr-Code is associated to a Box
+    }, [createToast, qrCodeExists]);
+  }
 
   // Query Data for the Form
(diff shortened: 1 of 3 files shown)


pylipp:
[2.0 FE Manage Boxes: create Boxes from Manage Boxes #medium](https://trello.com/c/qV48bCAT/1649-20-fe-manage-boxes-create-boxes-from-manage-boxes-medium)

aerinsol:
@pylipp screencap please?

front/src/views/BoxCreate/BoxCreateView.tsx
```diff
@@ -92,10 +92,23 @@ function BoxCreateView() {
   // variables in URL
   const qrCode = useParams<{ qrCode: string }>().qrCode!;
 
-  // Query the QR-Code
-  const qrCodeExists = useQuery(CHECK_IF_QR_EXISTS_IN_DB, {
-    variables: { qrCode },
-  });
+  if (qrCode) {
+    // Query the QR-Code
+    const qrCodeExists = useQuery(CHECK_IF_QR_EXISTS_IN_DB, {
```
fhenrich33, line 97: @pylipp Conditional use of hooks violates the rules of hooks, we are likely to use useLazyQuery and remove the if statement while keeping this block of code.

I took over the task, but I figure it would be good to leave this FYI. 👍

pylipp: Good to know, thanks :+1:

pylipp:
Locally all front-end tests pass :heavy_check_mark:

HaGuesto (approved):

pylipp:
> What does the mobile viewport look like?

@aerinsol please check the first post above; I added screenshorts for mobile (also _before_ this PR). I reckon that the button should not contain the "Create Box" text on mobile, instead the "+"-icon only?

aerinsol:
> > What does the mobile viewport look like?
> 
> @aerinsol please check the first post above; I added screenshorts for mobile (also _before_ this PR). I reckon that the button should not contain the "Create Box" text on mobile, instead the "+"-icon only?

Not sure what you mean by before this PR, but yes, should be the plus icon only, and all action buttons should fit on the same line even in the smallest mobile viewport. Shrink the buttons if you have to.

pylipp:
> but yes, should be the plus icon only, and all action buttons should fit on the same line even in the smallest mobile viewport. Shrink the buttons if you have to.

@aerinsol made all button fit on the same line and updated the screenshot

aerinsol:
Ideally there should be a gap between the first four which are left aligned, and the last two, which are right aligned, so there's a visual breakup between the different control buttons. If you can't do it as a quick fix though it's ok.

PR #2173, opened 2025-05-21
Set default value for product comment to empty string

pylipp:
- replicate behavior of dropapp; and behavior of box-creation
- if a product is created with a null comment, changing the comment will
  result in history messages of form 'comments changed from "None" to "new comment"'
  which we want to avoid


pylipp:
[2.0: Set default value for product comment to empty string](https://trello.com/c/IaJolk2L/1767-20-set-default-value-for-product-comment-to-empty-string)

PR #2184, opened 2025-05-28
Delete or disable product table action

fhenrich33:
https://trello.com/c/wGjXJhl5/1747-20-add-table-delete-action-for-custom-products-and-disable-standard-product-in-all-products-table-small

firefox_dp1YndVONn

image


front/src/hooks/useDisableOrDeleteProducts.ts
```diff
@@ -0,0 +1,217 @@
...
+  const baseId = useAtomValue(selectedBaseIdAtom);
+  const { createToast } = useNotification();
+  const { triggerError } = useErrorHandling();
+
+  const [disableStandardProductMutation, { loading: disableStandardProductMutationLoading }] =
+    useMutation(DISABLE_STANDARD_PRODUCT_MUTATION);
+  const [deleteProductMutation, { loading: deleteProductMutationLoading }] =
+    useMutation(DELETE_PRODUCT_MUTATION);
+
+  const refetchQueries = useCallback(
```
HaGuesto, line 74: drop the refetchQueries

front/src/views/Products/components/DisableStandardProductAlert.tsx
```diff
@@ -0,0 +1,43 @@
...
+  instockItemsCount,
+  transferItemsCount,
+}: {
+  productName: string;
+  instockItemsCount: string;
+  transferItemsCount: string;
+}) {
+  return (
+    <>
+      You are attempting to disable the product {productName} with {instockItemsCount}{" "}
```
HaGuesto, line 14: make this flexible so it also works for deleting a product

front/src/hooks/useDisableOrDeleteProducts.ts
```diff
@@ -0,0 +1,217 @@
...
+        });
+      }
+    },
+    [createToast, disableStandardProductMutation, refetchQueries, triggerError],
+  );
+
+  const handleDeleteProduct = useCallback(
+    (productId: string) => {
+      if (productId) {
+        deleteProductMutation({
```
HaGuesto, line 157: do the same validation checks if boxes are inStock or inTransit as for the standardproduct

front/src/views/Products/components/DeleteProductOverlay.tsx
```diff
@@ -0,0 +1,60 @@
+import { VStack, chakra } from "@chakra-ui/react";
+import { AreYouSureDialog } from "components/AreYouSure";
+import { FaTrashAlt } from "react-icons/fa";
+
+interface IRemoveBoxesOverlayProps {
```
HaGuesto, line 5: pls rename

HaGuesto (commented):
@fhenrich33 a few pieces you can fix the few small inputs I gave, but generally this is on hold for a product decision.

aerinsol:
> @fhenrich33 pls post a video of the current state.
> 
> @aerinsol do you want an AreYouSure Dialog to open for each delete of a custom product. If yes, the behavior is then different to disabling a standardproduct.
> 
> In generally, I am not sure if we want an AreYouSure Dialog for each product to delete. It makes sense if you try to delete multiple items in bulk, but is a bit much if you are deleting single products.

No. It should reproduce standard product behavior with instock boxes. That's the only thing.

pylipp:
@fhenrich33 Feedback from Roanna about not having any AreYouSure dialog when deleting products is here. I'll functionally review after the second iteration.

fhenrich33:
@aerinsol @HaGuesto @pylipp Updated PR with a GIF.

pylipp (approved):
### Functional test passed
Just a small update to the items-in-use-alert requested.
Could also test error cases by tinkering with the BE resolvers.
Didn't go much into depth on the code review.

front/src/views/Products/components/DisableOrDeleteProductAlert.tsx
```diff
@@ -0,0 +1,45 @@
...
+  transferItemsCount: string;
+  disableOrDelete: "disable" | "delete";
+}) {
+  return (
+    <>
+      You are attempting to {disableOrDelete} the product {productName} with {instockItemsCount}{" "}
+      <Text fontWeight="600" color="#659A7E" display="inline">
+        InStock
+      </Text>
+      {!!transferItemsCount && (
```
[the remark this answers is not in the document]
HaGuesto: everything in curly brackets is a functional insert into the html. it is at the end a short-hand notation for a conditional

HaGuesto:
@aerinsol maybe we should do a small visual difference between the delete and disable buttons, e.g. different background color.

front/src/hooks/useDisableOrDeleteProducts.ts
```diff
@@ -0,0 +1,194 @@
...
+        }
+      };
+
+      if (
+        (instockItemsCount !== undefined && instockItemsCount > 0) ||
+        (transferItemsCount !== undefined && transferItemsCount > 0)
+      ) {
+        createToast({
+          title: `${disableOrDelete === "disable" ? "Disabling" : "Deleting"} Product with Active Stock`,
+          message,
```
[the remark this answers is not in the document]
fhenrich33: I wanted to keep the conditional message logic wrapped in the hook, at least it makes sense IMO, and since we can only declare and use Components in TSX files, I pass this as an argument directly.

front/src/views/Products/components/ProductsContainer.tsx
```diff
@@ -92,6 +97,37 @@ function ProductsContainer() {
...
+          <Button
+            onClick={(e) => {
+              e.stopPropagation();
+
+              handleDisableOrDeleteProduct(
+                row.original.isStandard ? "disable" : "delete",
+                <DisableOrDeleteProductAlert
+                  productName={row.original.name}
+                  instockItemsCount={row.original.instockItemsCount + ""}
+                  transferItemsCount={row.original.transferItemsCount + ""}
```
pylipp, line 115: @fhenrich33 I have to repeat my [previous comment](https://github.com/boxwise/boxtribute/pull/2184#discussion_r2123141509) about the alert message.
When `row.original.transferItemsCount` is 0, then the alert message should be 
```
      You are attempting to {disableOrDelete} the product {productName} with {instockItemsCount}{" "}
        InStock items in one or more locations. To continue, you must first reclassify all{" "} InStock
      boxes as a different product.
```
**No** info about MarkedForShipment/Transit states nor about shipments.
This was the old behavior on the StandardProductsTab but this is not the current behavior.


Screenshots from v2-staging:
- trying to disable an ASSORT product with only InStock boxes
2025-06-05-113140_1620x321_scrot
- trying to disable an ASSORT product with InStock and MarkedForShipment boxes
2025-06-05-113324_1611x332_scrot

I think the issue is that `row.original.transferItemsCount` is passed as string into the Alert component. A value of 0 becomes "0" with evaluates to true in the `{!!transferItemsCount && (` line. The following patch fixes the issue:
```diff
diff --git a/front/src/views/Products/components/DisableOrDeleteProductAlert.tsx b/front/src/views/Products/components/DisableOrDeleteProductAlert.tsx
index 832422a0..35e5d2af 100644
--- a/front/src/views/Products/components/DisableOrDeleteProductAlert.tsx
+++ b/front/src/views/Products/components/DisableOrDeleteProductAlert.tsx
@@ -7,8 +7,8 @@ function DisableOrDeleteProductAlert({
   disableOrDelete = "disable",
 }: {
   productName: string;
-  instockItemsCount: string;
-  transferItemsCount: string;
+  instockItemsCount: number;
+  transferItemsCount: number;
   disableOrDelete: "disable" | "delete";
 }) {
   return (
diff --git a/front/src/views/Products/components/ProductsContainer.tsx b/front/src/views/Products/components/ProductsContainer.tsx
index ef2179e0..4aa10eb7 100644
--- a/front/src/views/Products/components/ProductsContainer.tsx
+++ b/front/src/views/Products/components/ProductsContainer.tsx
@@ -111,8 +111,8 @@ function ProductsContainer() {
                 row.original.isStandard ? "disable" : "delete",
                 <DisableOrDeleteProductAlert
                   productName={row.original.name}
-                  instockItemsCount={row.original.instockItemsCount + ""}
-                  transferItemsCount={row.original.transferItemsCount + ""}
+                  instockItemsCount={row.original.instockItemsCount}
+                  transferItemsCount={row.original.transferItemsCount}
                   disableOrDelete={row.original.isStandard ? "disable" : "delete"}
                 />,
                 row.original.id,
diff --git a/front/src/views/Products/components/StandardProductsContainer.tsx b/front/src/views/Products/components/StandardProductsContainer.tsx
index c2bfe77c..a5528d1e 100644
--- a/front/src/views/Products/components/StandardProductsContainer.tsx
+++ b/front/src/views/Products/components/StandardProductsContainer.tsx
@@ -128,8 +128,8 @@ function StandardProductsContainer() {
                     "disable",
                     <DisableOrDeleteProductAlert
                       productName={row.original.name}
-                      instockItemsCount={row.original.instockItemsCount + ""}
-                      transferItemsCount={row.original.transferItemsCount + ""}
+                      instockItemsCount={row.original.instockItemsCount}
+                      transferItemsCount={row.original.transferItemsCount}
                       disableOrDelete="disable"
                     />,
                     row.original.instantiationId,
```

aerinsol:
> @aerinsol maybe we should do a small visual difference between the delete and disable buttons, e.g. different background color.

Sure, we should make it red. Ideally we should at an are you sure check for just the deleted as well.

aerinsol:
> > > @aerinsol maybe we should do a small visual difference between the delete and disable buttons, e.g. different background color.
> > 
> > Sure, we should make it red. Ideally we should at an are you sure check for just the deleted as well.
> 
> Does it mean the delete-button (for custom products) should have a slight-red background, and the disable-button (for SP instantiation) should have none?
> 
> @aerinsol you [said above](https://github.com/boxwise/boxtribute/pull/2184#issuecomment-2917427946) that you don't want any are-you-sure check.

You're right, my bad. And yes, that's what I mean.

fhenrich33:
@aerinsol @pylipp Which would be the "slightly red" colour we are referencing?

front/src/hooks/useDisableOrDeleteProducts.ts
```diff
@@ -0,0 +1,194 @@
...
+    }
+  `,
+  [],
+);
+
+export const DELETE_PRODUCT_MUTATION = graphql(
+  `
+    mutation DeleteProduct($productId: ID!) {
+      deleteProduct(id: $productId) {
+        __typename
```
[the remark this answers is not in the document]
pylipp: I found that using a fragment breaks the functionality of the table auto-updating after the delete/deactivate action, so I reverted it

pylipp:
> @aerinsol @pylipp Which would be the "slightly red" color we are referencing?

I think the `red.500` you chose is fine because it's used in other places as well.

PR #2189, opened 2025-06-05
Fix/add box to cache after create box

HaGuesto:


HaGuesto:
merged into #2188 to clean up history and to make fixing build issues easier

PR #2191, opened 2025-06-07
If already deleted, don't delete product again

pylipp:
- avoid update of deleted_on field
- avoid creating another history entry


pylipp:
[2.0 BE: If already deleted, don't delete product again](https://trello.com/c/2wyMiqvQ/1778-20-if-already-deleted-dont-delete-product-again)

PR #2196, opened 2025-06-13
Remove unused env variables

pylipp:


pylipp:
[CircleCI: check why FRONT_ env variables are used in config.yml but REACT_APP_ ones are stored in context](https://trello.com/c/VWuOJhRI/1753-circleci-check-why-front-env-variables-are-used-in-configyml-but-reactapp-ones-are-stored-in-context)

PR #2199, opened 2025-06-17
Slim down FE requests

pylipp:
- Query less shipment data for BoxView
- Slim down request for updating numberOfItems or state
- Slim down request for updating location
- Update query for BoxEditView
- Add state filter parameter to shipments query
- Only query Preparing shipments for BoxView


pylipp:
superceded by #2200

PR #2205, opened 2025-06-24
Move shared GraphQL definitions for statistics type into separate file

pylipp:


pylipp:
[2.0 public BE: implement query for standard products](https://trello.com/c/1c62yxaV/1787-20-public-be-implement-query-for-standard-products)

PR #2206, opened 2025-06-24
GraphQL API for public ASSORT page

pylipp:
- Add standardProducts query to public API
- Move definition for use_db_replica function
- Use DB replica for public standard products query


pylipp:
[2.0 public BE: implement query for standard products](https://trello.com/c/1c62yxaV/1787-20-public-be-implement-query-for-standard-products)

PR #2245, opened 2025-07-22
Update box history after assigning to/removing from shipment

pylipp:
- Remove unreachable code
- feat: Refetch box data after successful assignment
- Refetch box data after successfully removing from shipment

### DIFF
--- a/front/src/views/Box/BoxView.tsx
+++ b/front/src/views/Box/BoxView.tsx
@@ -459,10 +459,12 @@ function BTBox() {
             message: `Box has been successfully assigned to the shipment ${shipmentId}.`,
             status: "success",
           });
+          allData.refetch();
         }
       }
     },
     [
+      allData,
       assignBoxesToShipment,
       boxData,
       createToast,
--- a/front/src/views/Box/BoxView.tsx
+++ b/front/src/views/Box/BoxView.tsx
@@ -460,49 +460,10 @@ function BTBox() {
             status: "success",
           });
         }
-      } else {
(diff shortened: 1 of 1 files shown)


pylipp:
> why is the unreachable code being removed? I think there's a reassign box still in the Figma design?

Once a box is assigned to a shipment, the transfer tab is disabled and it effectivly can't be assigned to another shipment. One has to remove it from the first shipment, and then assign it to the second.
If you want the direct reassigning to be possible, the display of the transfer tab needs to be tweaked. @aerinsol

aerinsol:
I guess its not a major need right now. I'll bookmark this PR and associate it with the original design in case we need it again someday. Approved then on my side.

pylipp:
> I guess its not a major need right now. I'll bookmark this PR and associate it with the original design in case we need it again someday. Approved then on my side.
> 
> Edit @pylipp is it currently possible to remove the shipment from within BoxView, or does it have to be done from ShipmentView?

@aerinsol it's also possible to remove a box from a shipment within BoxView.

aerinsol:
> @aerinsol it's also possible to remove a box from a shipment within BoxView.

In that case the transfer tab is not disabled in the MarkedforShipment state, right?

pylipp:
> In that case the transfer tab is not disabled in the MarkedforShipment state, right?

that's correct, the transfer tab is enabled. It's the shipment-dropdown that's hidden, and instead the "Remove from Shipment" button is shown.

2025-07-29-201215_1346x502_scrot

PR #2319, opened 2025-08-19
More robust createQrCode back-end

pylipp:


pylipp:
too many conflicts by now

PR #2342, opened 2025-08-26
Use loaders for TransferAgreement.source_bases and .target_bases

pylipp:
Before: 16+7 DB calls to load source and target bases (for 4 agreements)
After: 4+5 DB calls to load source and target bases (along with orgs, agreements, shipments, users)


pylipp:
[2.0 BE: use loaders for TransferAgreement.source_bases and .target_bases](https://trello.com/c/bxOgtYt9/1828-20-be-use-loaders-for-transferagreementsourcebases-and-targetbases)

PR #2396, opened 2025-09-27
added a public newly registered beneficiary numbers endpoint

LREGS:
https://trello.com/c/cIItMVSu


back/boxtribute_server/business_logic/metrics/queries.py
```diff
@@ -16,3 +20,70 @@ def resolve_metrics(*_, organisation_id=None):
...
+    if last_quarter == 4:
+        next_quarter_start = date(year + 1, 1, 1)
+    else:
+        next_quarter_start = date(year, first_month_last_quarter + 3, 1)
+
+    quarter_end = next_quarter_start - timedelta(days=1)
+
+    # Last year
+    start_last_year = date(now.year - 1, 1, 1)
+    end_last_year = date(now.year - 1, 12, 31)
```
pylipp, line 55: Just a naming consistency remark: please use one of `first_day_of_last_X` or `start_last_X` for X in (month, quarter, year), same for `last_day`/`end_last`

back/boxtribute_server/business_logic/metrics/queries.py
```diff
@@ -16,3 +20,70 @@ def resolve_metrics(*_, organisation_id=None):
...
+
+    # Query counts
+    last_month_count = (
+        Beneficiary.select()
+        .where(
+            (Beneficiary.created_on >= first_day_of_last_month)
+            & (Beneficiary.created_on <= last_day_of_last_month)
+        )
+        .count()
```
pylipp, line 64: this block repeats two times, please add a small function `number_of_beneficiaries_registered_between(start, end)`.
Make the function return 0 if the result of count() would be None (to simplify lines 86-88)

back/test/data/beneficiary.py
```diff
@@ -109,6 +109,23 @@ def another_male_beneficiary_data():
     }
 
 
+# def recent_beneficiary_data():
```
pylipp, line 112: I'd actually like to see a recent beneficiary in the test data (maybe registered in the previous month?) to make the test more reliable.
Is there a particular reason why you commented this?

back/test/endpoint_tests/test_metrics.py
```diff
@@ -96,3 +96,29 @@ def test_metrics_query_for_god_user(
         "numberOfFamiliesServed": number_of_families_served,
         "numberOfSales": number_of_sales,
     }
+
+
+@pytest.mark.parametrize(
+    "lastMonth, lastQuarter, lastYear",
+    [
+        [0, 0, 0],
```
pylipp, line 104: Did you plan to have different parameters here?
If it's just one list you can remove the `parametrize` block and the arguments and hardcode the 0s in the assertion below.

pylipp (changes requested):
Congratz @LREGS for your first PR in the boxtribute repo 🎉 
I'm really impressed by the thorough implementation. You set up all development tools and made the full CI pipeline pass.
I put a couple of comments but this is not meant to come across as critizising! Only to adhere to our naming and testing standards.

back/test/endpoint_tests/test_metrics.py
LREGS: Yes, at one point and I've not removed it. 

I trying to add data but it just broke the tests. 

I gathered the dev site was populated differently than the tests db so I used the 
/public graphql interface instead.

back/boxtribute_server/business_logic/metrics/queries.py
```diff
@@ -16,3 +20,70 @@ def resolve_metrics(*_, organisation_id=None):
...
     # Pass organisation ID to child resolvers
     return {"organisation_id": organisation_id}
+
+
+@public_query.field("newlyRegisteredBeneficiaryNumbers")
+def resolve_newlyRegisteredBeneficiaryNumbers(*_):
+    now = datetime.now()
+
+    # Last month
+    first_day_this_month = now.replace(day=1)
```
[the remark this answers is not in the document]
LREGS: Ah thanks! I never noticed that but makes sense

pylipp (changes requested):
Hej @LREGS , thanks for adding the improvements.
The major request I have now is to implement a more robust test, please see my comments.
Let me know if you have any questions.

pylipp (changes requested):
nice, about to wrap it up 👍 
good job with the microseconds backwards-calculation 🤓

pylipp (approved):

PR #2400, opened 2025-10-05
added newly created boxes query

LREGS:
things I'd personally maybe want to change is not doing individual db calls 3 times in a row and maybe querying the db once for each call so it returned month, quarter year in a single query. 

Also not sure if get_time_ranges needs to return a dict and could just return a list of lists but didn't know if this makes it less readable so I stuck with current approach.

neither may even be an issue, just my thoughts :)


back/boxtribute_server/graph_ql/definitions/public/queries.graphql
```diff
@@ -3,4 +3,5 @@ type Query {
 
   standardProducts: [StandardProductInfo!]!
   newlyRegisteredBeneficiaryNumbers: BeneficiaryNumbers
+  newlyCreatedBoxes: BoxNumbers
```
pylipp, line 6: Can you name the query `newlyCreatedBoxNumbers`? 
When reading `newlyCreatedBoxes`, I'd assume the query returns box objects.

back/boxtribute_server/graph_ql/definitions/public/types.graphql
```diff
@@ -41,3 +41,12 @@ type BeneficiaryNumbers {
   " during the year before the current one "
   lastYear: Int!
 }
+
+"""
+Information on newly created boxes
+"""
+type BoxNumbers {
```
pylipp, line 48: pleasel copy the field comments from BeneficiaryNumbers over

back/test/data/box.py
```diff
@@ -41,6 +41,7 @@ def default_box_data():
 def box_without_qr_code_data():
     data = default_box_data()
     data["id"] = 3
+    data["created_on"] = datetime.today().replace(day=1) - timedelta(days=1)
```
pylipp, line 44: Unfortunately `box_without_qr_code_data()` is used to set up a lot of other test boxes in the same file.
Hence a couple of tests that rely on boxes, like `test_statistics.py` currently fail.
I think the effects would be less if you changed the creation date of a different box instead, e.g. `created_in_donated_location_box_data`.

back/test/endpoint_tests/test_metrics.py
```diff
@@ -115,3 +112,23 @@ def test_public_beneficiary_numbers(read_only_client):
...
+    current_month = datetime.today().month
+    assert response == {
+        "lastMonth": 14,
+        "lastQuarter": 14 if current_month in [1, 4, 7, 10] else 0,
+        "lastYear": 1 if current_month == 1 else 0,
+    }
+
+
+def build_newly_created_query(string):
+    return f"query {{ {string} {{ lastMonth lastQuarter lastYear }} }}"
```
pylipp, line 134: `query_name` instead of `string`? Then the purpose of the argument is more obvious.

pylipp (changes requested):
thanks for the fast implementation 🚀 
you found a nice solution in the crud.py file!
I'd ask you to change the creation date of a different box in `test/data/box.py`, and then check whether `test_statistics.py` succeeds, and if not, to fix that test, too.

back/test/endpoint_tests/test_metrics.py
```diff
@@ -115,3 +112,23 @@ def test_public_beneficiary_numbers(read_only_client):
         "lastQuarter": 1 if current_month in [1, 4, 7, 10] else 0,
         "lastYear": 1 if current_month == 1 else 0,
     }
+
+
+def test_public_box_number(
```
pylipp, line 117: I just remembered that we use the [freezegun](https://github.com/spulec/freezegun) package in a different place to manipulate the results of `datetime` functions. Maybe it would be interesting to use if for broader testing here. I was thinking of testing at least four different creation dates (of `created_in_donated_location_box_data`) for last year/quarter/month/none of those.
Please have a look at freezegun, experiment a bit in the test and let me know if you think it would be a good fit.

pylipp:
> things I'd personally maybe want to change is not doing individual db calls 3 times in a row and maybe querying the db once for each call so it returned month, quarter year in a single query.

I think there's not much you can do about combining 3 DB calls since it's with different date filters every time. It's not an expensive operation (though an index on created_on would probably benefit performance, but that's a different story), and it's well readable the current way.

> Also not sure if get_time_ranges needs to return a dict and could just return a list of lists but didn't know if this makes it less readable so I stuck with current approach.

Depends on the situation, but using dicts makes things clearer in my opinion because dict keys add context for dict values. Also it allows for constructing code with less repetition (see my comment in the queries.py file).

back/test/data/box.py
LREGS: Oops, I need better habits when checking tests I guess 'cause atm I'm only really running test_metrics but I should then be running the whole back-end tests and making sure I didn't accidently change something else?

pylipp: yes, that'd be best practice :smile: but the tests take about 7mins to complete.
I usually run one or two test files that I expect to be immediately affected by my changes, and push everything to Github. This triggers the CI pipelines (see below), and when one fails check it in more detail (the [tests tab](https://app.circleci.com/pipelines/github/boxwise/boxtribute/8216/workflows/27520db1-c932-47b1-8a28-0c0f06f3358e/jobs/50844/tests) is most helpful I find)

back/test/endpoint_tests/test_metrics.py
pylipp: means you could use the `@parametrize` decorator :)

LREGS: To clarify you're suggesting to create a new fixture that returns a number of new boxes with differing dates and then use freezegun to provide different times to test the new boxes against. 

I'm not sure that the introduction of freezegun would do anything that manually setting datetime() to a fixed date doesn't do anyway.

I would agree on reflection that using time.today() isn't a strong test because it isn't deterministic and I could just parametrize a few fixed dates instead.

Or, with differently dates from parametrize  I can keep the fixture returning a single date and make sure the logic is working by moving the test date.

pylipp: Honestly I was hoping to use only a single fixture with freezegun in the actual test function to manipulate different dates. In the end, this wouldn't work because the test data is set up before the test function (with freezegun) even runs. You're right that it wouldn't have any difference to manually settings datetime then.

If you have an idea on how to parametrize an underlying test box with different creation dates, go ahead :) One thing that just came to my mind is that you could leave the test data as-is (we avoid effects on unrelated tests), and instead create new boxes at different points in time in the test. You use `@parametrize` to set a list of interesting dates and in the test function, first execute the `createBox` mutation wrapped in `with freeze_time(creation_date):`.

Maybe that's all too complicated and you mean something different :sweat_smile: Because we could move around in time in the test, and hence influence the outcome of `get_time_ranges`. That's the easiest probably

aerinsol:
> I think there's not much you can do about combining 3 DB calls since it's with different date filters every time. It's not an expensive operation (though an index on created_on would probably benefit performance, but that's a different story), and it's well readable the current way.

If this is about month/quarter/year calls, not sure I agree @pylipp. I can't comment on readability, but SQL-side you can definitely structure it so a single query returns all three in the results

pylipp:
> If this is about month/quarter/year calls, not sure I agree @pylipp. I can't comment on readability, but SQL-side you can definitely structure it so a single query returns all three in the results.
> 
> I unfortunately can't read Python that well ^^;; so I'm not sure what you're doing in terms of the DB call, but I do have a lot of experience with data warehousing and pipelines, so maybe I can help if I understand.

thanks for tuning in @aerinsol . We have three queries like `SELECT COUNT(1) FROM people WHERE created > X and created < Y` with three different pairs of (X,Y), happy about any suggestions on how to reduce the number of queries.
Gemini suggests a conditional aggregation:
```sql
SELECT
    SUM(CASE WHEN created > X1 AND created < Y1 THEN 1 ELSE 0 END) AS count_range_1,
    SUM(CASE WHEN created > X2 AND created < Y2 THEN 1 ELSE 0 END) AS count_range_2,
    SUM(CASE WHEN created > X3 AND created < Y3 THEN 1 ELSE 0 END) AS count_range_3
FROM
    people;
```

pylipp:
> I'm not sure what the current proof of concept (output) of this is, but for insight purposes I would also recommend `GROUP BY` base, org, and maybe gender. With the `GROUP BY` you have the most common data exploration request done out of the box.

I understood that the new public API should provide high(est)-level statistics @aerinsol . For base and org statistics we already have queries in the authenticated API (the "statviz" queries).
Grouping by gender definitely makes sense for the public query.
If it's fine to publicly disclose base and org names with their statistics, we can do this in a follow-up task.

> Final question (not a blocker to merge) @LREGS - is this implementation currently parameterized?

The new GraphQL API is not parametrized but the underlying functions in the resolvers are (using start and end datetimes as parameters). It'd be straightforward to bring this into the queries, I was already considering it, too.

PR #2404, opened 2025-10-06
Log assigment of QR label to box in history table

pylipp:
https://trello.com/c/tMaZyO5s


pylipp:
The toast after successfully creating a box is the same for both scenarios
```
Box 54217120
Successfully created with Pillows (1x) in Stockroom.
```

**Scenario 1** (QR label created first, then scanned to create box): No history entry
**Scenario 2** (box created via ManageBoxes, then label created via BoxView): History entry "User created QR code label for box". After creating the label, a toast is displayed with 
```
Box 54217120
A label with QR code was successfully created. To show a printable PDF, please click the 
QR code icon next to the box number.
```

This replicates the dropapp behaviour. In scenario 1, on history entry was shown in dropapp's BoxView. In scenario 2, the history entry "New Qr-code assigned by pdf generation." was shown. @aerinsol

PR #2444, opened 2025-10-30
Pin vitejs/plugin-react-swc to v3

pylipp:


pylipp:
@copilot downgrade `@vitejs/plugin-react-swc` to version lower than v4 and update pnpm-lock.yaml

PR #2457, opened 2025-11-11
Addition of columns depending on filtering from url

MomoRazor:
This is the primary implementation of the task [here](https://trello.com/c/IlTpo3sX/1858-20-fe-show-manageboxes-columns-that-are-present-in-filters).

Right now it should handle any url param that is given to the the page, and should unhide the column by default. I was not able to quite understand how column specific filtering can be done from this table (if it is possible), but it should be handled as well with the approach implemented.

### DIFF
--- a/front/src/views/Boxes/BoxesView.tsx
+++ b/front/src/views/Boxes/BoxesView.tsx
@@ -172,7 +172,6 @@ function Boxes({
       "lastModifiedBy",
       "createdBy",
       "productCategory",
-      "id",
     ];
 
     const filterIds: string[] = [];


front/src/hooks/useTableConfig.ts
```diff
@@ -246,8 +285,20 @@ export const useTableConfig = ({
...
+      globalFilter: prevConfig.globalFilter,
+      columnFilters,
+      sortBy: prevConfig.sortBy,
+      hiddenColumns: prevConfig.hiddenColumns ? [...prevConfig.hiddenColumns] : undefined,
+    };
+
+    // Filter hiddenColumns based on filters
+    columnFilters.forEach((filter) => {
+      tableConfig.hiddenColumns = tableConfig.hiddenColumns?.filter((col) => col !== filter.id);
+    });
```
pylipp: The pattern of cloning a config, then setting its hiddenColumns based on the columnFilters repeats three times, can you put it into a function?
In the first case in line 199 you can probably drop the hasUrlParams condition?

MomoRazor: Agreed, in the interest of decreasing code repetition, I'll put the hiddenColumn modifiction in a function and remove the unnecessary check for hasUrlParams

front/src/hooks/useTableConfig.ts
```diff
@@ -179,23 +179,53 @@ export const useTableConfig = ({
...
+
+        // Clone default config to avoid mutation
         const tableConfig: ITableConfig = {
           globalFilter: defaultTableConfig.globalFilter,
           columnFilters: initialFiltersToPersist,
           sortBy: defaultTableConfig.sortBy,
-          hiddenColumns: defaultTableConfig.hiddenColumns,
+          hiddenColumns: defaultTableConfig.hiddenColumns
+            ? [...defaultTableConfig.hiddenColumns]
+            : undefined,
```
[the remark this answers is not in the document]
MomoRazor: I believe this might be an artifact of a possible solution to a rehydration issue I thought I was having. I'll revert this back if I see that it does not affect how it functions, and if it does, I'll investigate and explain whats happening in more detail here

MomoRazor:
Re excessive re-renders, this shouldn't have add any new re-renders, however, I do not think this is by any means the most optimized approach. Latest React best practices focus on decreasing reliance on useEffect as much as possible, but the current pattern of the table is heavily reliant on onMount hook calls. I did consider looking at alternative approaches, but it felt like a departure from the scope of the current task. I would suggest the creation of a backlog issue that focuses on reducing the use of the useEffect hook as much as possible, initially targeting this table component

MomoRazor:
I've been going through the failing frontend tests, and I've gathered that they might actually be due to new re-renders that my feature has added, because the tests are running out of mocks. Therefore I'm looking to improve my implementation to fix these tests without adding extra mocks, before I consider this done.

Also the backend tests started to fail when I merged master back to this branch. Apologies if I shouldn't have done that. If so I'll roll that particular commit back

pylipp:
> I've been going through the failing frontend tests, and I've gathered that they might actually be due to new re-renders that my feature has added, because the tests are running out of mocks. Therefore I'm looking to improve my implementation to fix these tests without adding extra mocks, before I consider this done.

Oh indeed then there's an extra request being issued by the FE. Appreciate the test fixing! 

> Also the backend tests started to fail when I merged master back to this branch. Apologies if I shouldn't have done that. If so I'll roll that particular commit back

I'm quite sure that's a flaky test that you can ignore. Haven't had it in quite a while. You can regularly merge latest master into the PRs, helps esp. when work happens on branches in parallel and master is frequently updated.

front/src/hooks/useTableConfig.ts
MomoRazor: Good spot on both of these notes. I've elected to implement both. Let me know if the code becomes less readable (once I have pushed the latest)

front/src/views/Boxes/BoxesView.tsx
```diff
@@ -153,29 +154,45 @@ function Boxes({
...
+      "gender",
+      "size",
+      "shipment",
+      "comment",
+      "age",
+      "lastModified",
+      "lastModifiedBy",
+      "createdBy",
+      "productCategory",
+      "id",
```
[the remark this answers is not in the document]
pylipp: true, left-over from #2363

PR #2459, opened 2025-11-16
QR Overlay Updates

MomoRazor:
Related to this [task](https://trello.com/c/ZnUZ1ij8/1467-20-qr-overlay-add-x-on-the-top-right-to-close-the-modal-make-it-uncloseable-otherwise).

Some general questions regarding the qr scanner

- I was unable to get open the qr scanner modal unless i was in mobile view (smaller width) and got the burger menu instead of the sidebar. Is this intentional, or am I just unable to find the QR Scanner button in PC mode?
- When testing the QR Reader on a previous task, I began by manually navigating to /bases/1/qrreader. Is there a way to get to this url normally, or should this url not be accessible technically?

### DIFF
--- a/front/src/components/QrReader/components/QrReader.tsx
+++ b/front/src/components/QrReader/components/QrReader.tsx
@@ -106,7 +106,7 @@ function QrReader({
                   onChange={(e) => onBoxLabelInputChange(e.currentTarget.value)}
                   isDisabled={findBoxByLabelIsLoading}
                   value={boxLabelInputValue}
-                  placeholder={"12345678"}
+                  placeholder="12345678"
                   borderRadius={0}
                 />
                 <InputRightElement>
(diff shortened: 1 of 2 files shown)


pylipp:
> Some general questions regarding the qr scanner

@MomoRazor Appreciate these questions! Anything that seems really off to you?

> * I was unable to get open the qr scanner modal unless i was in mobile view (smaller width) and got the burger menu instead of the sidebar. Is this intentional, or am I just unable to find the QR Scanner button in PC mode?

This is intentional because we consider scanning QR codes a mobile-only feature (use case: volunteers in warehouses scanning/creating boxes through their mobile device which is not really practical with a PC/laptop). I can't find relevant code though rn :sweat_smile: 

> * When testing the QR Reader on a previous task, I began by manually navigating to /bases/1/qrreader. Is there a way to get to this url normally, or should this url not be accessible technically?

This is the "landing page" when users log in to boxtribute on mobile, and it's also the fallback route on mobile when a URL is not found (NotFoundView.tsx).
FYI you can also go to /qrreader and it'll forward to /bases/1/qrreader (see App.tsx).

PR #2460, opened 2025-11-17
Minor FE fixes

pylipp:
https://trello.com/c/jjdKtc0c


MomoRazor:
I see no issue with any of this code, however I have not yet managed prompt the Alerts effected, or have data to see the Demographic Graph affected. 

On a related but separate note, during trying to test I also noticed that the god user (some.admin@boxtribute.org) does not seem to have access to Multi Box on the QR Reader, which I'm not sure if is intended.

Either way, I belive this code is fine given it's slight nature.

PR #2462, opened 2025-11-18
Quick ADR with the most obvious options for zxing replacement

MomoRazor:


docs/adr/adr_zxing_replacement.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Considered Options
+
+There are a couple of possible replacements that I have found up till now:
+
+- [zxing-wasm](https://www.npmjs.com/package/zxing-wasm) - This is another zxing based package that is however, kept up to date (last upload was 9 days ago as of writing). The package will probably be the easier replacements as the API will be the closest to the current implementation (due to it also working with zxing under the hood). The package is fairly popular with a 100k weekly downloads and an upwards trend.
+
+- [qrcode](https://www.npmjs.com/package/qrcode) - This is the most popular QR Code reader in the NodeJS ecosystem. Having almost 4M weekly downloads, this package is the basic standard of QR Code readers. Unfortunately, it has not been updated for the past year, suggesting that the package might be running out of maintainers, with little activity reported within the github to suggest maintainer handover under way.
+
+Personally, while `qrcode` seems to be the more popular option, it feels like it is on the way out in terms of maintenance, and therefore I would suggest using `zxing-wasm` which has showed very recent signs of activity up to this point.
```
pylipp: Also found https://github.com/nimiq/qr-scanner which is not maintained for 3y now, 
and https://github.com/mebjas/html5-qrcode with the last update 2y ago, looking for maintainers.
Seems these libraries don't live too long 😅 
There's https://github.com/yudielcurbelo/react-qr-scanner which has 400 stars on GH opposed to the 165 of zxing-wasm, did you see it?

docs/adr/adr_zxing_replacement.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Decision
+
+The proposed change is to replace the current library with either of the options above. This will require significant changes in the QRCode reader components of the boxtribute app specifically.
+
+## Consequences
```
pylipp: We use the zxing functionality only in https://github.com/boxwise/boxtribute/blob/master/front/src/components/QrReader/components/QrReaderScanner.tsx, can you assess how difficult porting would be?

docs/adr/adr_zxing_replacement.md
```diff
@@ -0,0 +1,46 @@
...
+
+Pros:
+
+- Intermittant issues with the QRCode reader should decrease or be removed completely.
+- If issues arrise due to lack of support, easier paths to support will be open (due to an active community on the packages)
+- A MUCH smaller bundle size, as the current zxing library embed more parts of the zxing engine then required by more modern browsers
```
pylipp: Can you tell a bit of an estimate here, like 500kB vs 50kB?

docs/adr/adr_zxing_replacement.md
MomoRazor: Yes alot of these packages kind of fall in disuse, but I believe I might have a theory for why hah.

Basically since the QR Code reading feature does not really require maintenance, there is not much to expand on with these packages, so they get forgotten. Then after some time, the ecosystem moves on, and suddenly they start to break when major changes happen in NodeJS, or Javascript (or Typescript), or maybe major changes in the devices that might affect barcode scanning. So essentially they get forgotten because they don't need alot of maintenance to work, but when they'll suddenly break, there is no one left to make sure they keep working.

The really truly future proof solution might be to fork one of these engines and keep a version of them up to date with consistently, but that feels like super overkill, at least at this point. That said, that is exactly how zxing-wasm was created, in fact its maintainer has an issue open on the @zxing/library repo explaining their motivation for zxing-wasm.

As for [react-qr-scanner](https://github.com/yudielcurbelo/react-qr-scanner), I did see it, however I didn't mention it simply because it references react in the package title, and that usually means that the package includes a specifically pre-build component for react, which might limit our UI customizability. On further inspection it does seem like they provide a 'Scanner' component, however it is not immediately obvious if we might have visual limitations, so I can add this as an option if we feel safer with its popularity, and we would like to check if it limits us in any way.

docs/adr/adr_zxing_replacement.md
```diff
@@ -0,0 +1,46 @@
...
+
+Proposed
+
+## Context or Problem Statement
+
+The current library used by the frontend to parse QR Codes has been in maintenance mode for the past 2 years. It would be ideal to move the QR Code reading to library that is still actively worked on, so as not to run into issues as devices continue to change and older software gets more and more out of date.
+
+## Decision Drivers
+
+Maintaining QR Code reader functionality, increasing longevity and fixing some rare issues that already occure on certain devices.
```
aerinsol: I would group these points under the headings:
1. Reliability: (What you said about functionality and longevity)
2. Maintainabilty + Bug Fixing: (what you said about issues increasing as devices change and older software gets more out of date)

docs/adr/adr_zxing_replacement.md
aerinsol: @MomoRazor if you have time to examine / compare the points between `react-qr-scanner` vs `zxing-wasm`, and discuss them in this section that would be ideal.

I would move your sentence about your current recommendation (and its reasoning) into the Decision section of the document.

docs/adr/adr_zxing_replacement.md
aerinsol: See my earlier comment about moving your comments about your recommended path into this Decision section.

docs/adr/adr_zxing_replacement.md
aerinsol: Pros/Cons need to be less vague and have either quantiites (e.g. package size) or examples (issues you believe will resolve)

docs/adr/adr_zxing_replacement.md
```diff
@@ -0,0 +1,46 @@
...
+
+Author: Maurovic Cachia
+
+## Status
+
+Proposed
+
+## Context or Problem Statement
+
+The current library used by the frontend to parse QR Codes has been in maintenance mode for the past 2 years. It would be ideal to move the QR Code reading to library that is still actively worked on, so as not to run into issues as devices continue to change and older software gets more and more out of date.
```
aerinsol: General feedback @MomoRazor , it's important to at least mention the user story / user impact or platform importance, or at least the scope of the issue. 

Specifically for this ADR, I would add the following context:

> QR code generation and scanning is a core part of the user workflows for warehousing and inventory management in Boxtribute. In particular, the in-house QR code scanner has X amount of functionality (or however many lines of custom code) associated with it. However, the current library used by the frontend to parse QR Codes (@MomoRazor link and name the library here) has been in maintenance mode for the past 2 years. This creates a medium-high risk  for both active field workers and the development team that increases with time as devices continue to change and older software becomes less and less compatible. For active field workers, the risk with increasing incompatiblity is if the QR scanner functionality breaks, their ability to effectively receive and distribute aid is compromised. For the development team, since QR scanning is a core application workflow an emergency fix will most likely be needed, which will impact the backlog and timelines of other active projects.

MomoRazor:
> [This component](https://github.com/boxwise/boxtribute/blob/master/front/src/components/QrReader/components/QrReaderScanner.tsx) was just a nightmare to write.

Heh I understand, React can get really irritable with re-renders sometimes, but challenge accepted!

docs/adr/adr_zxing_replacement.md
```diff
@@ -12,43 +12,82 @@ Proposed
...
+This creates a medium-high risk for both active field workers and the development team that increases with time as devices continue to change and older software becomes less and less compatible.
+
+For active field workers, the risk with increasing incompatiblity is if the QR scanner functionality breaks, their ability to effectively receive and distribute aid is compromised. For the development team, since QR scanning is a core application workflow an emergency fix will most likely be needed, which will impact the backlog and timelines of other active projects.
+
+Apart from this, more modern packages could bring secondary benefits, such as smaller bundle sizing (making load times and network usage decrease), quicker detection, and less battery utilization through lower level optimizations.
 
 ## Decision Drivers
 
-Maintaining QR Code reader functionality, increasing longevity and fixing some rare issues that already occure on certain devices.
+1. Reliability
```
aerinsol: @MomoRazor since we're talking about size now please change the title to "Reliability and Performance"

PR #2483, opened 2025-12-02
Fetch tag options after creating box with new tag

pylipp:
Avoid re-creating tag of same name when running "Save & create another"

https://trello.com/c/MqyXECSQ

### DIFF
--- a/front/src/views/BoxCreate/BoxCreateView.tsx
+++ b/front/src/views/BoxCreate/BoxCreateView.tsx
@@ -214,6 +214,16 @@ function BoxCreateView() {
         newTagNames,
         qrCode,
       },
+      refetchQueries:
+        // update tag options if new tag(s) created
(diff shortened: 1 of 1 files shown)


pylipp:
> How heavy are products and location queries here? Because reloading them here is technically not required (since the change is only to tags), if I understand correctly, but I'm aware that right now, tags, products and locations are all pulled with this same query.
> 
> It might merit looking into possible splitting up this query for optimization purposes later.

@MomoRazor That's a good point; in general it'd be more efficient to query only for tags, or to use the Apollo cache and request the tags as part of the mutation response (then again, that's redundant if we don't add any new tags).

On v2-staging, base 1, the options-query returns a response of 3.4kB in size. In total 100 elements, most of them products. On production it depends on the base ofc. I'd say it's a pragmatic solution to request the entire bunch only in the case of new tags being added.

PR #2505, opened 2025-12-16
Enable filtering of complex ManageBoxes content

pylipp:
https://trello.com/c/SVHhEK1Y
**Screenshot of fix**
2025-12-16-123711_1241x910_scrot

**Before**
2025-12-16-124500_1452x538_scrot


aerinsol:
UI currently looks fine to me, needs code review from @HaGuesto . When it's on staging I would like to test:
- if it works across columns
- if it works with spaces / quotes
- this is a FE only search right? So if the column is not shown it won't work? @pylipp

pylipp:
@aerinsol on staging now, please validate.

I can see the value of having an advanced search with multiple search words but I'd consider it a separate feature. Also it'd be a different UX than before the bug, and then there should be some explanation to the user.

aerinsol:
Validated on staging

PR #2526, opened 2026-01-08
Tag Module

MomoRazor:
Three bugs in the Tags overview table: header not sticky, date columns not sorting, and a crash when sorting tags with null dates.

## Sticky header (`TagsTable.tsx`)
`<Flex overflowX="auto">` created a new scroll container, scoping `<Thead position="sticky">` to a container that never scrolled vertically. Changed outer Flex to `height="100%"` (matches `BoxesTable`) and moved `overflowX="auto"` to a `<Box>` wrapping only `<Table>`.

## Date column sorting (`TagsContainer.tsx`)
All three date columns used `sort: "datetime"` — not a react-table v7 column property. Changed to `sortType: "datetime"`.

## Null date sort crash (`TagsContainer.tsx`)
react-table's built-in `"datetime"` sort calls `.getTime()` unconditionally, throwing `TypeError` when `lastUsedOn` (or other date fields) is `null` on a newly created tag. Replaced with a custom sort function that pushes nulls to the end:

```ts
function nullableDateSortType(rowA, rowB, columnId) {
  const a = rowA.values[columnId];
  const b = rowB.values[columnId];
  if (!a && !b) return 0;
  if (!a) return 1;
  if (!b) return -1;
  return a.getTime() - b.getTime();
}
```

## Pre-existing TS error (`transformers.ts`)
`lastModifiedOn` fallback was `new Date(createdOn)` — `createdOn` can be `null`, causing a type error. Changed to `undefined` when both are null, consistent with how `lastUsedOn` is handled.

### DIFF
--- a/front/src/components/Table/FilterPanel.tsx
+++ b/front/src/components/Table/FilterPanel.tsx
@@ -29,7 +29,7 @@ export function FilterPanel({
   children,
 }: FilterPanelProps) {
   return (
-    <Popover placement={placement} isOpen={isOpen} onOpen={onOpen} onClose={onClose}>
+    <Popover placement={placement} isOpen={isOpen} onOpen={onOpen} onClose={onClose} isLazy>
       <PopoverTrigger>
         <IconButton
           icon={<MdFilterList color={"black"} size={25} />}


MomoRazor:
Hi guys! Quick update on this:

- Firstly, I tackled all the feedback I marked as resolved here
- I fixed the column filtering working as a union rather then an intersect. A side fix for this was using `Portal` to make the Filter UI be able to show even if the Table is too small for the UI to fit.
- Fixed the create and update tag when selecting a different application
- Removed the dropdown and rendering the buttons next to each other
- Switched the column name of 'Application' to the dropapp name 'Apply To'
- Fixed the Update form title
- Added a helper text when Updating a Tag on the Application field
- Have a random colour selected when creating a new Tag

With those done, what remains are the following:

- Adding created and modified dates on the table
- Stopping the checkout column from being hideable
- The rest of the unresolved stuff

Let me know if I may have missed anything :)

front/src/views/Tags/UpdateTagView.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+  const onSubmit = useCallback(
+    (updateTagOutput: ICreateTagFormOutput) => {
+      updateTag({
+        variables: {
+          id: props.tagId,
+          name: updateTagOutput.name,
+          type: updateTagOutput.application,
+          color: updateTagOutput.color,
+          description: updateTagOutput.description,
```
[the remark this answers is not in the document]
MomoRazor: Is this to mean that it the description **shouldn't** be optional, or that it is currently not working as optional? If it is the latter, I'm afriad I cannot replicate it, so maybe I might have fixed it inadvertantly, let me know if this is still an issue.

pylipp:
> Instead, I'm wondering if it's possible for you to put the hex selection code in the drop-down for more usability, an option with some default colors and then a proper color picker, like we did in Dropapp? I think it's likely a JS package, I'm sure there must be similar in React. 

@aerinsol re possible packages, @MomoRazor and I had a discussion previously. 
- We currently use Chakra v2.
- v3 brings a [ColorPicker component](https://www.chakra-ui.com/docs/components/color-picker) to the library. However the upgrade to v3 needs lots of adjustments in the codebase first, hence not an option for now
- there is a custom [ColorPicker package](https://github.com/Buupu/chakra-color-picker/blob/main/src/colorPicker.tsx) for Chakra v1. @MomoRazor I understand we can't install the package due to the Chakra version conflict but I think we didn't consider copying over and adjusting the main component? It's only 83 LOC and I reckon it could be compatible. Not sure how its UI looks like though
- maybe the changes you requested are already possible with Mauro's custom implementation, so we don't need to consider the v1 component at all

MomoRazor:
The following is an example of how the new custom color picker works. Let me know what you think :)

[Screencast from 2026-01-19 17-44-55.webm](https://github.com/user-attachments/assets/b5e5ed42-0f5d-4da6-8c91-ae9363e8d461)

aerinsol:
LGTM!

front/src/views/Tags/CreateTagView.test.tsx
```diff
@@ -0,0 +1,475 @@
...
+
+    // Click nevermind button
+    const nevermindButton = screen.getByRole("button", { name: /nevermind/i });
+    await user.click(nevermindButton);
+
+    // Verify navigation back
+    expect(mockNavigate).toHaveBeenCalledWith("..");
+  });
+
+  it("displays loading state during submission", async () => {
```
pylipp: this can be removed

front/src/views/Tags/UpdateTagView.test.tsx
```diff
@@ -0,0 +1,562 @@
...
+    const nevermindButton = screen.getByRole("button", { name: /nevermind/i });
+    await user.click(nevermindButton);
+
+    // Verify navigation back
+    expect(mockNavigate).toHaveBeenCalledWith("..");
+  }, 20000);
+
+  // Note: Component doesn't render when tagId is missing - this is expected behavior
+
+  it("displays loading state during submission", async () => {
```
pylipp: no value

PR #2542, opened 2026-01-15
Take tag assignments into account for reached beneficiaries statistic

pylipp:
https://trello.com/c/G0zpb662

### DIFF
--- a/back/boxtribute_server/business_logic/metrics/crud.py
+++ b/back/boxtribute_server/business_logic/metrics/crud.py
@@ -4,6 +4,7 @@ from datetime import datetime, timedelta, timezone
 
 from peewee import fn
 
+from ...enums import TaggableObjectType
 from ...models.definitions.base import Base
 from ...models.definitions.beneficiary import Beneficiary
 from ...models.definitions.box import Box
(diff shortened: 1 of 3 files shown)


back/boxtribute_server/business_logic/metrics/crud.py
```diff
@@ -209,6 +211,13 @@ def number_of_beneficiaries_reached_between(start, end):
                 ServicesRelation.created_on <= end,
             )
         )
+        | (
+            TagsRelation.select(TagsRelation.object_id.alias("id")).where(
+                TagsRelation.object_type == TaggableObjectType.Beneficiary,
+                TagsRelation.created_on >= start,
+                TagsRelation.created_on <= end,
+            )
+        )
```
pylipp: @aerinsol extending the UNION query for reached beneficiaries with 
```sql
SELECT object_id AS id
FROM tags_relations
where object_type = "People"
and created_on >= X and created_on <= Y
```

pylipp: @aerinsol a tag can be of the three types you mention. This is directly stored in the type field of the tags table.

When a tag is assigned to a box or a beneficiary, this relation is stored in the tags_relations table. We store a generic object ID of the tagged box/beneficiary, so we also need an object type to trace back if the tag was assigned to a box or a beneficiary

aerinsol: Approved SQL

PR #2641, opened 2026-03-24
Return only boxes from non-archived locations in boxes query

pylipp:


pylipp:
Not changing the default behavior of including archived locations for now.
Instead, with the new filter UI users will be able to filter for e.g. Donated state + location in a single go.

Issue #2308, opened 2025-08-12
Update eslint to v9


pylipp:
### Background
We use eslint as tool for our front-end code, currently v8.56 which has reached EOL in 2024.

### Task
- update eslint to v9
- take into consideration the migration guide from https://eslint.org/docs/latest/use/migrate-to-9.0.0
- migrate all config files

### More instructions

- write modern, correct TypeScript with valid Type annotations that don't result in any tsc errors
- make sure all FE builds pass with `pnpm -C front build` and `pnpm -C statviz build`
- make sure all FE tests pass with `pnpm test`
- try to do as little code changes as possible to achieve the requested functionality
- use Claude Opus 4 model to solve this

Horlawale-ui:
@pylipp Assign me to this issue so that I can fix and make a PR

pylipp:
> superceded by https://trello.com/c/kquoFxow

tracked via trello

PR #2673, opened 2026-04-15
Eslint update 2 (split part 1)

MomoRazor:
This is a PR cherry picking the simplest changes from [this PR](https://github.com/boxwise/boxtribute/issues/2308) in an effort to help decrease the review load needed for Eslint 9 here. This PR include changes which I believe need minimum review effort, as mostly remove syntactic sugar which is no longer recommended by Eslint, with actual conditionals.

If any of the new logic here has any effect, then it is an affect that the system was already suffering from in all likelyhood.

@pylipp let me know if this helps or if I should attempt to split it up further! This is planned to be part 1 of 3 :)


pylipp:
https://trello.com/c/kquoFxow

pylipp (approved):

PR #2704, opened 2026-05-05
Avoid pulling pre-2023 data into movedBoxes stat

pylipp:
https://trello.com/c/EFSdLs3F

This didn't affect data display because in the FE we filter by a date range (earliest 2023-01-01).
It reduces the querying time and request size (and hence latency) for bases with a considerable amount of pre-2023 boxes.

### DIFF
--- a/back/boxtribute_server/business_logic/statistics/sql.py
+++ b/back/boxtribute_server/business_logic/statistics/sql.py
@@ -35,11 +35,13 @@ BoxHistory AS (
         h.id AS id
     FROM history h
     JOIN ValidBoxes s ON h.record_id = s.id
-    AND ((h.to_int IS NOT null AND h.id >= %s)
-         OR h.changes = "Record created"
-         OR h.changes = "Record deleted"
-         OR h.changes = "Box was undeleted."
+    AND (
+        h.id >= %s
+        AND (
+            h.to_int IS NOT NULL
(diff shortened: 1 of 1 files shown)


pylipp:
https://github.com/copilot/c/3f260545-dd88-47cb-89f0-7c910bf30847

aerinsol:
@pylipp I assume that %s is a variable - is there a reason the computation date isn't hardcoded in? Or are you setting this in a central module where it's then called by different functions?

pylipp:
> @pylipp I assume that %s is a variable - is there a reason the computation date isn't hardcoded in? Or are you setting this in a central module where it's then called by different functions?

@aerinsol yes, the query is invoked from one place which passes the cut-off history ID. Since that's not compatible for testing, this variable is only set in production. 
We could probably switch it for a date-date-threshold and hardcode it but I have to check if the history.changes column is indexed

pylipp:
> We could probably switch it for a date-date-threshold and hardcode it but I have to check if the history.changedate column is indexed

@aerinsol the changedate column is not indexed (only composite index with tablename or record_id), so I stick with the approach to filter for a variable `h.id` (for testing set to 0; for production, the first ID in 2023), [code for reference](https://github.com/boxwise/boxtribute/blob/11fa78da61f1e404ae3c3467aaf989fcab0aec91/back/boxtribute_server/business_logic/statistics/crud.py#L399-L402)

aerinsol (approved):

PR #2815, opened 2026-07-06
FE for box monetary value/weight

pylipp:
https://trello.com/c/vJPyKiR8

## Additional Changes

Added inline editing of box weight and monetary value in the ShipmentContent table:

- **New "WEIGHT" and "MONETARY VALUE" columns** added after the "PRODUCT" column in the shipment box table
- **Inline cell editing**: clicking a cell turns it into an input field (accepting non-negative floats) with a unit label and a Save button (disabled until a valid number is entered)
- **Empty value indicator**: cells with no weight/monetary value show a bold red hyphen on a light red background
- **`updateMarkedForShipmentBox` mutation** is called on Save to persist the updated weight or monetary value


pylipp:
@copilot+claude-sonnet-4.6 I need an update of the ShipmentContent/ShipmentTable w.r.t box weight and monetary value.

- add two column definitions after product with header "WEIGHT" and "MONETARY VALUE"
- in both columns there are custom cells that allow for inline editing of the values
- the cell UI depends on its state:
    - when inactive (i.e. not being clicked on), it shows the box weight/monetary value as label, with the resp. unit
    - if no weight/monetary value present, render the cell background slight red, and show a bold red hyphen without any unit
    - if the user clicks the cell, it turns into an input with white background (that accepts non-negative floats, optional), the unit label next to it (Box.weightDisplayUnit.symbol or €), with a Save action right next to it (disabled unless a valid number is entered)
    - when Save is clicked, run the updateMarkedForShipmentBox mutation

PR #2859, opened 2026-08-11
ADR for website tracking tool

pylipp:
https://trello.com/c/VjUWWEy3

[The ADR with markdown rendered](https://github.com/boxwise/boxtribute/blob/adr-website-tracking/docs/adr/website-tracking.md).

### DIFF
--- a/docs/adr/website-tracking.md
+++ b/docs/adr/website-tracking.md
@@ -18,11 +18,13 @@ For assessing popularity and usage it is important for the team to have informat
 ## Decision Drivers
 
 - no costs
-- low set-up and maintainence effort
+- low set-up and maintenance effort
 - integration into existing NextJS website
 - no cookie-banner required (no data storage acc. to GDPR)
 - tracking of call-to-actions (CTA), downloads, page visits (with source via UTM tags)
 - distinguish human and robot traffic
+
+Note that in the main app, we already use Heap for tracking user interaction.
 
 ## Considered Options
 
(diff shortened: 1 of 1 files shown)


HaGuesto (approved):
@pylipp looks good. can you please add what data they are storing of visitors and in which region? 

Could you also send me a Data Processing Agreement (DPA) with them once you sign up? We probably need to adapt our privacy policy on the webpage.

aerinsol (changes requested):
I took a look at both Plausible and Umami last week, and I noticed a few things that were missed by this version of the ADR:

1. Plausible is EU cloud based, Umami is US based
2. In terms of data sub-processors, Umami explicitly lists Claude and other AI agents. It didn't go into much detail. Per our new AI policy, I think we need to flag this as an explicit risk. 

For the purposes of a simple marketing campaign where we touch no sensitive data, I think the analysis still holds. But if at some point later we revisit this with thought of expanding usage to more sensitive data or self-hosting, we need to consider this and make an explicit decision on the hosting. Thoughts / feedback @HaGuesto @pylipp ?

HaGuesto:
@aerinsol I agree even though I have a few remarks:
- I would not expect that we would process more sensitive data in the future. For our scale I do not know why we would need more personalised website tracking. We also would need to add a cookie banner and I have a strong feeling about them.
- @pylipp In your analysis it says that plausible bot filtering is "not available in cookieless mode". They do not have a cookie mode. That was also a driver for me to go to plausible, but actually their filtering capabilities seem to be the strongest.
- Umami would still be ok for me if they host in the EU and traffic does not go through the US. Actually an interesting case to specify more clearly in the AI policy.

--> @pylipp let's go for plausible.

pylipp:
> can you please add what data they are storing of visitors and in which region?

@HaGuesto I added two rows to the table for comparing stored data and server location.

While [umami states](https://docs.umami.is/docs/cloud/faq#where-are-your-servers-located) that they servers in both US and EU, I couldn't find a way to control this as a user.

> In your analysis it says that plausible bot filtering is "not available in cookieless mode". They do not have a cookie mode. That was also a driver for me to go to plausible, but actually their filtering capabilities seem to be the strongest.

I got the columns mixed up. You're right, plausible doesn't use any cookies.

> let's go for plausible.

Mind the capabilities of the least expensive Plausible tier (9€/mo, vs the next one at 14€/mo):
- team size of one (three)
- shared links for the analytics dashboard not available

> But if at some point later we revisit this with thought of expanding usage to more sensitive data or self-hosting

@aerinsol If we self-host Umami, it depends on the infrastructure we use. Should be in the EU, and I think Umami then has no rights/means to process or store any of the data.

pylipp:
> While [umami states](https://docs.umami.is/docs/cloud/faq#where-are-your-servers-located) that they servers in both US and EU, I couldn't find a way to control this as a user.

@HaGuesto during sign-up, one can select the [data region](https://docs.umami.is/docs/cloud/sign-up#account-setup).

The concerns about Umami's AI subprocessors remain.

aerinsol:
@HaGuesto, I don't think we need to expand it for the webpage, but I can see a situation for example, where we'd want to unify both internal product tracking and website clicks. 

In such a situation, for our product tracking, we already list Heap Analytics in our subprocessors - they actually have [quite refined controls](https://help.heap.io/hc/en-us/articles/37271887185041-Heap-Privacy-Features) to allow us to make sure that Heap doesn't capture / we don't send any sensitive information into their servers.  - but it's quite a sophisticated product tracking tool that Plausible / Umami would likley not be able to emulate.

HaGuesto:
@pylipp @aerinsol I'd still go for plausible even if it costs 9€ per month. Their product offering is solid and it definitely agrees with our AI policy.
I'm still fine with Umami hosted in the EU since the type of data we would be sharing is not sensitive, but the long-term commitment to reduce dependency on US tech would speak against it.

Please make a decision and go for it

aerinsol:
I just checked, and Plausible appears to have 3 years data retention vs the 6 months for Umami hobbyist - and Plausible allows  us to request special pricing. Let's go for Plausible and email them asking for a discount during/after the free trial since we're a small-scale ANBI that will never get 10k pageviews.

aerinsol:
@pylipp would be great if you could update the ADR though.

PR #2868, opened 2026-08-18
Require stock:write permission when creating QR code for box

pylipp:
https://trello.com/c/xImBGdcT


pylipp:
resolved via 1ed80be35

PR #2890, opened 2026-09-01
Hotfix for localstorage mocking

MomoRazor:
This PR focusing on fixing an issue that was added with the latest usage of Local Storage, which broke all the tests since testing frameworks usually do not have access to the Local Storage API. With this quick mock, tests should be able to pass!


pylipp:
https://trello.com/c/aUVf2Z1A

pylipp (approved):

